PR #36, opened 2018-04-09
added lerna config, updated scripts property accordingly

rkzel:
`npm install` and `test` are working via lerna. Also changed name of the app from 'app-name' to 'range-voting'.


Quazia:
I think the issue is that this was branched off of master and all the Travis integration lives in PR 1-contributing-documentation.

Quazia:
LGTM, we'll see if we run into issues with this further down the line haha.

Issue #69, opened 2018-04-24
Add totalParticipation variable to the vote structure in RangeVoting


Quazia:
In order to measure the participation without iterating through votes it's necessary to hold this value in the vote struct. This variable will be updated through the vote function in the RangeVoting contract.

PR #86, opened 2018-05-11
23 implement can execute

### LINKED ISSUE #23: Implement canExecute
`canExecute` is used to check that the participation has been met and the vote has reached its end before the execute function is called.
    * param `_voteId`: The ID of the Vote which would be executed.
    * return `True` if the vote is eligible for execution.

`_minParticipationPct` is percentage of voters that must participate in a vote for it to succeed (expressed as a 10^18 percentage, (eg 10^16 = 1%, 10^18 = 100%)

This issue should be accompanied by tests.

capsulecorplab:
closes #23 and #69 

rebased from https://github.com/Giveth/planning-app/commit/f236150175ebb9219b4ce8938c3e80b47437d1b1

### DIFF
--- a/apps/range-voting/contracts/RangeVoting.sol
+++ b/apps/range-voting/contracts/RangeVoting.sol
@@ -189,7 +189,6 @@ contract RangeVoting is IForwarderFixed, AragonApp {
     }
 
     /**
-    * @param _voteId id for vote structure this 'ballot action' is connected to
     * @notice `addCandidate` allows the `ADD_CANDIDATES_ROLE` to add candidates
     *         (or options) to the current vote.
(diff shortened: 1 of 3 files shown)


capsulecorplab:
Reopening and pulling changing into branch 146-rangevoting-contract-tests, since it should've been working.

Quazia:
Once the merges are in I'll review. Looks like there's some issues with the way the check is being handled as noted previously. I would also look at the way totalParticipation is being tallied.

capsulecorplab:
Branch should be ready to merge as of https://github.com/Giveth/planning-app/pull/86/commits/e97f88aec9569304aff6e94354fb0c279aaa816b.

PR #106, opened 2018-05-24
Initial backend & frontend integration

deamme:
This is based on the work of branch `28-range-voting-integration` that @Quazia did. I fixed some things on the frontend, added some dev commands and added documentation on current project structure workflow. Work still needs to be done for full backend and frontend integration.


rkzel:
-  handleCreateProjectOpen = () => {
+  handleCreateProjectOpen() {

I'm curious, why?

In manifest.json, it should be "Planning application".

deamme:
Performance wise arrow functions are not that great in classes but I do like the auto-bind property of it. I didn't change this myself though, so maybe @Quazia can add something to this.

deamme:
Closing this in favour of branch `28-range-voting-integration`

PR #125, opened 2018-07-06
54 addressbook

### LINKED ISSUE #54: Address Book - add entity
Here is an initial starting point for a basic registry. This contract needs to be adapted and turned into a full Aragon app  (with a barebones UI) that supports the input fields in the invision.
https://github.com/aragonlabs/registry/blob/master/contracts/RegistryApp.sol
https://projects.invisionapp.com/d/main#/console/14049654/292530913/preview

For this use case we need to store:
Name, Address, and type.
For type I would use an enum, the key for the registry entry should still be a bytes32(this will be a hash of the name) and then the address is an address.

For the general app strucutre I would follow along with the general structure of the tutorial. https://hack.aragon.org/docs/tutorial.html 
Just a simple text entry field that emulates the side-panel should be fine for a starting point.

capsulecorplab:
Closes #54


Quazia:
I'll review and merge as soon as the split-apps branch has been pulled in.

Issue #114, opened 2018-06-25
Separate contracts into designated folders


capsulecorplab:
file structure should look like:
```
planning-app/
└─ apps/
   ├─ address-book/
   ├─ payout-engine/
   └─ range-voting
```

stellarmagnet:
not sure if there should be a parent "planning app" folder since one of the apps is the planning app

file structure should look like:
```
└─ apps/
   ├─ address-book
   ├─ allocations
   └─ range-voting
   └─ planning
```

capsulecorplab:
@stellarmagnet the root directory is called `planning-app` by default because that's the name of the repo.

Quazia:
I think we're on the same page with this. I'm all for calling the folder that holds the "PayoutEngine" contract allocations, I think for clarity sake we should probably rename the contract to Allocations?

capsulecorplab:
I concur. 'Allocations', semantically speaking, makes more sense.

PR #127, opened 2018-07-08
114 separate contracts into designated folders

capsulecorplab:
UI elements (i.e., `app/` folder) currently subsides in `apps/range-voting/`. Will probably need to be relocated later. Also, need to clean up package.json in multiple directories.


capsulecorplab:
Will reopen once #126 and #114 are merged into dev

PR #132, opened 2018-08-01
M2 base

Quazia:


ottodevs:
Fixed conflicts

Quazia:
I added a *redundant* dep in the root package which we should figure out a way to remove/get around but for the time being this makes everything pass.

PR #142, opened 2018-08-04
Update PlanningKit.sol

stellarmagnet:
some of the permissions seemed off... did i do this right?

### DIFF
--- a/contracts/PlanningKit.sol
+++ b/contracts/PlanningKit.sol
@@ -108,25 +108,25 @@ contract PlanningKit is KitBase {
         // Allocations permissions:
         acl.createPermission(ANY_ENTITY, allocations, allocations.START_PAYOUT_ROLE(), root);
         acl.createPermission(rangeVoting, allocations, allocations.SET_DISTRIBUTION_ROLE(), root);
-        acl.createPermission(voting, allocations, allocations.EXECUTE_PAYOUT_ROLE(), root);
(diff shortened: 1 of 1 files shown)


contracts/PlanningKit.sol
```diff
@@ -108,25 +108,25 @@ contract PlanningKit is KitBase {
...
         // Range-voting permissions
-        acl.createPermission(voting, rangeVoting, rangeVoting.CREATE_VOTES_ROLE(), voting);
+        acl.createPermission(allocations, rangeVoting, rangeVoting.CREATE_VOTES_ROLE(), voting);
         acl.createPermission(ANY_ENTITY, rangeVoting, rangeVoting.ADD_CANDIDATES_ROLE(), root);
         acl.createPermission(voting, rangeVoting, rangeVoting.MODIFY_PARTICIPATION_ROLE(), root);
-        InstalledApp(voting, apps[3]);
+        InstalledApp(rangeVoting, apps[3]);
 
         // TokenManager permissions
         acl.createPermission(voting, tokenManager, tokenManager.ASSIGN_ROLE(), voting);
```
stellarmagnet: should this also close with "root" (since i guess it seems that's the current dev way...?

ottodevs: Good catch! I think I introduced the errors when refactoring the kit, too much copy-paste, my apologies.
The fix seems ok for me.
Thank you 😄

PR #183, opened 2018-10-14
148 vote updates ui

Quazia:
Isn't cleaned up completely but I'm moving on to getting the vote working then will circle back

### DIFF
--- a/apps/range-voting/app/script.js
+++ b/apps/range-voting/app/script.js
@@ -28,9 +28,15 @@ async function handleEvents(response){
       console.info('[RangeVoting > script]: received StartVote')
       nextState = await startVote(nextState, response.returnValues)
       break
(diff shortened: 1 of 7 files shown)


apps/range-voting/app/script.js
```diff
@@ -112,27 +130,41 @@ function loadVoteData(voteId) {
...
-    const nextVotes = Array.from(votes)
     nextVotes[voteIndex] = await transform(nextVotes[voteIndex])
-    return nextVotes
   }
+  return nextVotes
 }
 
-async function updateState(state, voteId, transform) {
-  const { votes = [] } = state ? state : []
+async function getCandidate(voteId, candidateIndex) {
```
topocount: I like this solution

PR #219, opened 2018-10-21
Fix #197 wrapper icons not shown when using npm run start:dev (httpProvider option)

### LINKED ISSUE #197: Application icons in wrapper
When npm starts in http mode (so all the apps are served by separate parcel instances), icons do not appear in the wrapper. Names do, icons don't. Build process for individual apps needs to put icon.svg into /dist/images/icon.svg, so request for http://localhost:3333/images/icon.svg can be successful.

ottodevs:
Icons are now shown properly with httpProvider.
Also fixes the problem where script.js was not copied to dist folder when running npm start:dev


apps/allocations/package.json
```diff
@@ -13,7 +13,8 @@
     "ganache-cli:test": "../shared/test-helpers/ganache-cli.sh",
     "lint": "solium --dir ./contracts",
     "prepublishOnly": "truffle compile --all",
-    "publish:http": "aragon apm publish --http localhost:2222 --http-served-from ./dist",
+    "publish:http": "npm run publish:http-copy-files && aragon apm publish --http localhost:2222 --http-served-from ./dist",
```
Quazia: What's the reason for using a different port for each app?

ottodevs: Because with httpProvider we are simultaneously running 4 independent parcel server instances (actually 3 right now, because AddressBook is not), parcel is smart enough to figure if a instance is running and use a new random port, but since we need to previously know the port to tell aragon apm publish we define a different one in each app.

Issue #205, opened 2018-10-19
Improve issues with number of zeros displayed for token balances


stellarmagnet:
For the side panel in Range Vote (# voting tokens) and the Account cards in Allocations page (balance), we need an elegant solution for the number of zeros/"token decimals" etc that are displayed, so its consistent and also aligns with norms.

(there is currently an implementation for this, but it's not the elegant solution)

Quazia:
Should be addressed in PR #474

PR #221, opened 2018-10-22
Tweak styles in Range Voting, Slider UI and others.

### LINKED ISSUE #203: Range Voting: Slider UI enhancements
Range Voting Side Panel
- [ ] When the slider is at the 0 position, the circle should be left-aligned with the text of the options label and the list of options
- [ ] The vote input box should be center-aligned with the slider

image

ottodevs:
Each change is detailed in the individual commit.
Closes #203 
Workaround for #205 

I couldn't manage to customize AragonUI default Slider, so I needed to override the component to define custom width and align the handle and the bar to their parent container

### DIFF
--- a/apps/allocations/arapp.json
+++ b/apps/allocations/arapp.json
@@ -1,6 +1,6 @@
 {
   "appName": "allocations.aragonpm.eth",
-  "version": "7.0.0",
+  "version": "0.0.1",
   "roles": [
     {
       "name": "Initiate an allocations payout",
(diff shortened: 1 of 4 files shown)


ottodevs:
We are using custom width for the Slider Area, and also moved a little the Bar and the Handle to fit the required specs, I opened an issue in AragonUI, but at this point I am not sure if there is a way to style by targeting specific CSS selectors, I tried to do that but I couldn't manage to do it, my styles never get applied.

Quazia:
Interesting, worth looking into more at some point for the sake of the Aragon ecosystem but for now this seems to do the trick.

PR #230, opened 2018-10-25
allocations tests passing with guards

topocount:
I've reenabled some guards to get the tests passing and the contract performing safely in addition to updating the tests to work with the contract updates


topocount:
Allocation tests passing now after merge. should be good to merge into dev now, @Quazia

PR #237, opened 2018-10-27
Updates to UI module and missing packages in Projects

ottodevs:
Updates UI components in order to solve several problems and warnings happening in the browser DevTools.

Also adds missing modules to Projects App.

- rxjs-compat and babel-core seemed requirements to have the script running
- graphql-request is our minimal requirement to interact with Github API

Dependencies are tested and there were no conflicts found, script.js is working now.

### DIFF
--- a/apps/allocations/package.json
+++ b/apps/allocations/package.json
@@ -22,7 +22,7 @@
   },
   "dependencies": {
     "@aragon/client": "1.0.0-beta.9",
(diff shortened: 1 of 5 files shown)


ottodevs:
We are already using the same codebase, so there is no need to relay on an aragon-ui fork at this moment, we can switch to the main repo

PR #238, opened 2018-10-27
Connect Projects script with the frontend

ottodevs:
This PR addresses the connection and communication between script.js and App.js in projects app.

I also added the initial graphql requests to combine the results in the script.js state, separating concerns and following the single source of truth best practices.

It probably needs other Project related pull requests merged before working correctly


ottodevs:
Fixed merged conflict, should be green now

PR #243, opened 2018-10-29
Fix coverage after identify a well known truffle bug

ottodevs:
Fixed by using 2nd form of the overloaded function Kernel.newAppInstance in the 4 occurrences in test files.

This addresses this bug we are getting in the CI coverage build:

https://github.com/trufflesuite/truffle/issues/737

The issue does not happen locally but using this solution does not break.

This bug seems to be fixed in the next version truffle@5 https://github.com/trufflesuite/truffle/issues/1171#issuecomment-410536873


Quazia:
Can you resolve the conflicts here locally quick if you get a chance?

PR #247, opened 2018-10-30
185 update range voting tests

topocount:


apps/range-voting/test/range-voting.js
```diff
@@ -531,6 +501,28 @@ contract('RangeVoting App', accounts => {
         const canExecute = await app.canExecute(voteId)
         assert.equal(canExecute, false, 'canExecute should be false')
       })
+      it('holder can add candidates', async () => {
+        //console.log(accounts[5], accounts[6],'all accounts:', accounts)
```
Quazia: We should probably start pulling all the extraneous event logging out before the review

PR #254, opened 2018-11-24
133 address book

rkzel:


apps/allocations/app/components/Form/Input/OptionsInput.js
```diff
@@ -18,14 +18,17 @@ class OptionsInput extends React.Component {
...
   }
 
   addOption = () => {
     // TODO: Implement some rules about what an 'Option can be' duplicates, etc
     const { input, name, value } = this.props
-    if (input && !value.includes(input)) {
+    if (input && !value.includes(input) && this.props.validator(input)) {
       this.props.onChange({ target: { name, value: [...value, input] } })
       this.props.onChange({ target: { name: 'optionsInput', value: '' } })
+      this.props.error = true
```
sohkai: You generally don't want to do this; you probably want to have a component set its state instead (changing the prop like this doesn't cause a rerender, and the next rerender might overwrite these existing props leading to this state being lost).

apps/allocations/app/components/Panel/NewAllocation.js
```diff
@@ -1,16 +1,9 @@
 import PropTypes from 'prop-types'
 import React from 'react'
 import styled from 'styled-components'
+import { BigNumber } from 'bignumber.js'
```
sohkai: Not a big deal ATM, but highly recommend using [BN.js](https://github.com/indutny/bn.js/) instead.

apps/allocations/app/components/Panel/NewAllocation.js
```diff
@@ -66,23 +59,48 @@ class NewAllocation extends React.Component {
...
     // clear input here.
+
     let informational = (this.state.allocationTypeIndex === 0)
     let recurring = !informational && this.state.payoutTypeIndex != 0
     // TODO: period should be smarter: now the only option is monthly
     let period = recurring ? 86400 * 31 : 0
-
+    let optionsInput = this.state.optionsInput
+    this.setState({
+      addressError: this.state.addressError,
```
sohkai: This probably doesn't need to be reset here

apps/allocations/app/components/Panel/NewAllocation.js
```diff
@@ -66,23 +59,48 @@ class NewAllocation extends React.Component {
...
+      allocationError: false,
+    })
+    if (!(this.isAddress(optionsInput) || optionsInput === '')) {
+      this.setState({ addressError: true })
+      this.setState()
```
sohkai: Could probably remove this extra `setState()`

apps/allocations/app/components/Panel/NewAllocation.js
```diff
@@ -66,23 +59,48 @@ class NewAllocation extends React.Component {
...
+      return
+    }
+    if (this.isAddress(optionsInput)) {
+      this.state.options.push(this.state.optionsInput)
```
sohkai: What you probably want here is to use the [function argument form of `setState()`](https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous) instead (and as a rule, **never** directly modify state).

apps/allocations/app/components/Panel/NewAllocation.js
```diff
@@ -66,23 +59,48 @@ class NewAllocation extends React.Component {
...
-      balance: this.state.amount,
+      balance: this.state.amount * 10e17,
+    }
+    if (allocation.balance > this.props.limit && !informational) {
+      this.setState({ allocationError: true })
+      return
     }
-
     this.props.onSubmitAllocation(allocation)
+    this.state.allocationError = false
```
sohkai: See above

apps/allocations/app/components/Panel/NewAllocation.js
Quazia: Why exactly do you recommand BN.js, in my experience BigNumber has a richer feature set.

apps/allocations/app/components/Form/Input/OptionsInput.js
ottodevs: OptionsInput was refactored to SFC and props were reviewed

apps/allocations/app/components/Panel/NewAllocation.js
ottodevs: Error system at NewAllocation panel was completely refactored

apps/allocations/app/components/Panel/NewAllocation.js
ottodevs: Done

apps/allocations/app/components/Panel/NewAllocation.js
ottodevs: Fixed

apps/allocations/app/components/Panel/NewAllocation.js
ottodevs: 👍 fixed

ottodevs:
All review comments were addresses, it remains pending the decision to switch to BN.js

apps/allocations/app/components/Panel/NewAllocation.js
sohkai: That's true, there are specific cases where `BigNumber.js` is much easier to work with, but `BN.js` is faster and most implementations are moving to it.

PR #257, opened 2018-11-30
resolve #226: bounties integration

### LINKED ISSUE #226: Projects app [Bulk Bounties] - Web3

capsulecorplab:
Implements integration for issuing and accepting bulk bounties.


topocount:
@Quazia mind giving the okay to merge this? It seems gauche to merge without a third party signing off.

apps/projects/contracts/Projects.sol
```diff
@@ -1,14 +1,66 @@
...
+    function acceptFulfillment(
+        uint _bountyId, 
+        uint _fulfillmentId
+    ) external;
+}
 
 
 contract Projects is AragonApp {
-    function initialize( // solium-disable-line blank-lines
-        //Vault _vault
```
topocount: @capsulecorplab Can we actually put all vault references back in? We will eventually incorporate the vault here. If you feel strongly against it, we don't have to include it, but I think it's good way to demonstrtate near-term intent to add it.

capsulecorplab: SGTM

capsulecorplab: @topocount We still have it uncommented in [L11](https://github.com/spacedecentral/planning-suite/pull/257/files/9ee12d3dcb75ebb8548d970535b1abfd33f2b8ac#diff-8bb0f56b9a5eaab6509b2ba5a2518e0cL11)

apps/projects/contracts/Projects.sol
```diff
@@ -23,95 +75,224 @@ contract Projects is AragonApp {
...
 
     // Fired when a repository is added to the registry.
-    event RepoAdded(bytes32 id);
+    event RepoAdded(bytes32 repoId);
     // Fired when a repository is removed from the registry.
-    event RepoRemoved(bytes32 id);
+    event RepoRemoved(bytes32 repoId);
     // Fired when a bounty is added to a repo
-    event BountyAdded(bytes32 owner, bytes32 repo, uint256 issueNumber, uint256 bountySize);
+    event BountyAdded(bytes32 owner, bytes32 repoId, uint256 issueNumber, uint256 bountySize);
```
Quazia: Do we need all of this information in this event?

apps/projects/contracts/Projects.sol
```diff
@@ -23,95 +75,224 @@ contract Projects is AragonApp {
...
+    // Fired when fulfillment is accepted
+    event FulfillmentAccepted(bytes32 repoId, uint256 issueNumber, uint fulfillmentId);
```
Quazia: If the "primary key' is all three of these values would it make sense to create a unique PK that maps into a struct of these three values?

apps/projects/contracts/Projects.sol
```diff
@@ -23,95 +75,224 @@ contract Projects is AragonApp {
...
+     */
+    function addBounties(
+        bytes32 _repoId,
+        uint256[] _issueNumbers, 
+        uint256[] _bountySizes, 
+        uint256[] _deadlines,
+        bool[] _tokenBounties,
+        address[] _tokenContracts,
+        string _ipfsAddresses
+    ) public payable auth(ADD_BOUNTY_ROLE)
```
Quazia: This will be complicated to handle on the front end; we might have to use the hacky proxy approach being used in the allocations since it's currently hard to call payables from the front end.

apps/projects/contracts/Projects.sol
```diff
@@ -23,95 +75,224 @@ contract Projects is AragonApp {
...
+    {
+        // ensure the transvalue passed equals transaction value
+        checkTransValueEqualsMessageValue(msg.value, _bountySizes,_tokenBounties);
+        
+        string memory ipfsHash;
+        uint standardBountyId;
+        // submit the bounty to the StandardBounties contract
+        for (uint i = 0; i < _bountySizes.length; i++) {
```
Quazia: Gas concerns here but you've raised this in the review PR.

apps/projects/contracts/Projects.sol
Quazia: We should probably re-factor this to go through the vault but that's not really a short-term strategy.

apps/projects/test/projects.test.js
```diff
@@ -89,72 +100,316 @@ contract('Projects App', accounts => {
...
-  context('creating and retrieving repos and bounties', function() {
-    it('creates a repo id entry', async function() {
-      repoID = (await app.addRepo(
-        'MDEyOk9yZ2FuaXphdGlvbjM0MDE4MzU5',
-        'MDEwOlJlcG9zaXRvcnkxMTYyNzE4MDk='
-      )).logs.filter(x => x.event == 'RepoAdded')[0].args.id
+  context('creating and retrieving repos and bounties', () => {
+    let repoId
+
+    beforeEach(async () => {
```
Quazia: Might be better to move this to a before instead of a beforeEach.

apps/projects/test/projects.test.js
```diff
@@ -89,72 +100,316 @@ contract('Projects App', accounts => {
...
+      repoId = addedRepo(
+        await app.addRepo(
+          'MDEyOk9yZ2FuaXphdGlvbjM0MDE4MzU5',
```
Quazia: If these are hard-coded probably best to move them to a constant.

apps/projects/test/projects.test.js
```diff
@@ -89,72 +100,316 @@ contract('Projects App', accounts => {
...
+          'MDEwOlJlcG9zaXRvcnkxMTYyNzE4MDk='
+        )
+      )
+    })
+
+    it('creates a repo id entry', async () => {
       assert.equal(
-        repoID,
+        repoId,
         '0xd1f2b806d3ffc90a501a7c22dbbcbb3f1f14e136cd9da208dc0b6e6b0f64b777',
```
Quazia: Move hardcoded value to constant.

apps/projects/test/projects.test.js
```diff
@@ -89,72 +100,316 @@ contract('Projects App', accounts => {
...
+    })
+
+    it('retrieve repo information successfully', async () => {
+      const repoInfo = await app.getRepo(repoId, { from: owner1 })
+      const result = web3.toAscii(repoInfo[0])
 
-      // result = repoInfo[0]
       assert.equal(
         result,
         'MDEyOk9yZ2FuaXphdGlvbjM0MDE4MzU5',
```
Quazia: move hardcoded value to constant.

apps/projects/test/projects.test.js
```diff
@@ -89,72 +100,316 @@ contract('Projects App', accounts => {
...
-        'invalid repo info returned'
+        'valid repo info returned'
```
Quazia: Is this inversion correct?

apps/projects/test/projects.test.js
```diff
@@ -89,72 +100,316 @@ contract('Projects App', accounts => {
...
-      const bountyAdder = accounts[2]
-      repoID = (await app.addRepo('abc', String(123))).logs.filter(
-        x => x.event == 'RepoAdded'
-      )[0].args.id
-      issue3Receipt = (await app.addBounties(repoID, [1, 2, 3], [10, 20, 30], {
-        from: bountyAdder,
-      })).logs.filter(x => x.event == 'BountyAdded')[2]
-      issue3Bounty = issue3Receipt.args.bountySize.toNumber()
-      assert.equal(issue3Bounty, 30, 'bounty not added')
+    context('standard bounty verification tests', () => {
```
Quazia: If we're going to do standard bounties tests they should be in another test file. Only exception would be if it's tests that are routed through the Projects contract.

apps/projects/test/projects.test.js
```diff
@@ -89,72 +100,316 @@ contract('Projects App', accounts => {
...
+      )
+      assertRevert(async () => {
+        await app.addBounties(
+          repoId,
+          [1, 2, 3],
+          [10, 20, 30],                   // 60 total Wei should be sent
+          [Date.now() + 86400, Date.now() + 86400, Date.now() + 86400],
+          [false, false, false],
+          [0, 0, 0],
+          'QmbUSy8HCn8J4TMDRRdxCbK2uCCtkQyZtY6XYv3y7kLgDCQmbUSy8HCn8J4TMDRRdxCbK2uCCtkQyZtY6XYv3y7kLgDCQmbUSy8HCn8J4TMDRRdxCbK2uCCtkQyZtY6XYv3y7kLgDC',
```
Quazia: I would use constants and concatenation to make it more clear what's happening here.

apps/projects/test/projects.test.js
```diff
@@ -89,72 +100,316 @@ contract('Projects App', accounts => {
...
+        )
+      )
+      await app.addBounties(
+        repoId,
+        [1, 2, 3],
+        [10, 20, 30],
+        [Date.now() + 86400, Date.now() + 86400, Date.now() + 86400],
+        [false, false, false],
+        [0, 0, 0],
+        'QmbUSy8HCn8J4TMDRRdxCbK2uCCtkQyZtY6XYv3y7kLgDCQmVtYjNij3KeyGmcgg7yVXWskLaBtov3UYL9pgcGK3MCWuQmR45FmbVVrixReBwJkhEKde2qwHYaQzGxu4ZoDeswuF9w',
```
Quazia: Same as above

apps/projects/truffle.js
```diff
@@ -23,8 +23,6 @@ module.exports = {
       host: 'localhost',
       port: 8545,
       network_id: '*',
-      gas: 7984452, // Block Gas Limit same as latest on Mainnet https://ethstats.net/
```
Quazia: Do things not work with mainnet gas usage?

apps/shared/test-helpers/contracts/lib/bounties/inherited/HumanStandardToken.sol
```diff
@@ -0,0 +1,57 @@
+/*
+This Token Contract implements the standard token functionality (https://github.com/ethereum/EIPs/issues/20) as well as the following OPTIONAL extras intended for use by humans.
```
Quazia: Might be useful to also test with MiniMe token but this is low priority.

apps/projects/contracts/Projects.sol
```diff
@@ -23,95 +75,224 @@ contract Projects is AragonApp {
...
+            ipfsHash = substring(_ipfsAddresses, i.mul(46), i.add(1).mul(46));
```
Quazia: Main line to comment

apps/projects/contracts/Projects.sol
```diff
@@ -23,95 +75,224 @@ contract Projects is AragonApp {
...
+        uint256[] _issueNumbers, 
+        uint256[] _bountySizes, 
+        uint256[] _deadlines,
+        bool[] _tokenBounties,
+        address[] _tokenContracts,
+        string _ipfsAddresses
+    ) public payable auth(ADD_BOUNTY_ROLE)
+    {
+        // ensure the transvalue passed equals transaction value
+        checkTransValueEqualsMessageValue(msg.value, _bountySizes,_tokenBounties);
```
Quazia: I would add some checks in here to make sure the length of the _ipfsAddresses string and all other arrays passed in are consistent in length.

apps/projects/contracts/Projects.sol
```diff
@@ -23,95 +75,224 @@ contract Projects is AragonApp {
...
+    function checkTransValueEqualsMessageValue(
+        uint256 _msgValue,
+        uint256[] _bountySizes,
+        bool[] _tokenBounties
+    ) internal pure 
+    {
+        uint256 transValueTotal = 0;
+        for (uint i = 0; i < _bountySizes.length; i++) {
+            if (!(_tokenBounties[i])) {
+                transValueTotal = transValueTotal.add(_bountySizes[i]);
```
Quazia: Since you're already commited to the computational complexity of safe math you might be better off setting the initial total to msg.value and then subtracting from that. That way you'll dump out of the function quicker if it's not enough to cover all the bounties. This might be one of those situations where the decreased readability outweighs the performance improvement though.

apps/projects/contracts/Projects.sol
```diff
@@ -23,95 +75,224 @@ contract Projects is AragonApp {
...
+    }
+
+    function substring(
+        string str,
+        uint startIndex,
+        uint endIndex
+    ) internal pure returns (string)
+    {
+        // first char is at location 0
+        //IPFS addresses span from 0 (startindex) to 46 (endIndex)
```
Quazia: Since this is a generalized function I'd re-locate these comments and maybe explain this more generally here.

PR #262, opened 2018-12-13
added tests for 'invalid operations'

capsulecorplab:
- moved/renamed 'should revert when adding an entry for repeated name'
 - added 'should revert when adding an entry by unauthorized address'
 - added 'should revert when removing an entry by unauthorized address'

### DIFF
--- a/apps/address-book/test/address-book.js
+++ b/apps/address-book/test/address-book.js
@@ -50,19 +50,18 @@ contract('AddressBook App', accounts => {
     app = AddressBook.at(
       receipt.logs.filter(l => l.event == 'NewAppProxy')[0].args.proxy
     )
-    
-    await app.initialize();
 
-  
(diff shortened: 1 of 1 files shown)


ottodevs:
.solcover.js file should be updated as well to skip misc/Migrations.sol , that may help to reflect the correct coverage

capsulecorplab:
> .solcover.js file should be updated as well to skip misc/Migrations.sol , that may help to reflect the correct coverage

Thanks! Added misc/Migrations.sol to skipFiles in .solcover.js

> The first overview seems ok in terms of code, but it is also weird, since this is a test focused PR it should increase the covered lines report, but it remained the same, Any thoughts on that?

Probably because the tests are just running the same 'it' tests, but with different accounts to make sure they revert for unauthorized addresses as expected. So code coverage should be the same

PR #394, opened 2019-02-08
323 repo card

rkzel:


Quazia:
Looks good to me! The filtering is so smooth too. Works like a charm.

PR #407, opened 2019-02-09
Issues 269 and  270: projects contract features

topocount:


topocount:
I still need to run through the demo script

topocount:
@Quazia this is ready for review

apps/projects/contracts/Projects.sol
```diff
@@ -121,12 +135,21 @@ contract Projects is AragonApp {
     event FulfillmentAccepted(bytes32 repoId, uint256 issueNumber, uint fulfillmentId);
     // Fired when settings are changed
     event BountySettingsChanged();
+    // Fired when user requests issue assignment
+    event AssignmentRequested(address applicant, bytes32 indexed repoId, uint256 issueNumber);
```
Quazia: I don't know if we need to fire the address for these events, we'll have to re-sync all of the info for the issues regardless. Thoughts?

apps/projects/contracts/Projects.sol
```diff
@@ -91,14 +91,28 @@ contract Projects is AragonApp {
...
         uint256 number; // May be redundant tracking this
         bool hasBounty;
+        bool fulfilled;
         uint256 bountySize;
         uint256 priority;
         address bountyWallet; // Not sure if we'll have a way to "retrieve" this value from status open bounties
         uint standardBountyId;
+        address assignee;
+        address[] applicants;
+        address[] workSubmittors;
```
Quazia: Will we have multiple people submit work or only one?

apps/projects/contracts/Projects.sol
```diff
@@ -287,6 +312,101 @@ contract Projects is AragonApp {
...
+    function requestAssignment(
+        bytes32 _repoId, 
+        uint256 _issueNumber, 
+        string _application
+    ) external isInitialized 
+    {
+        require(bytes(repos[_repoId].issues[_issueNumber].assignmentRequests[msg.sender]).length == 0, "User already applied for this issue");
+
+        repos[_repoId].issues[_issueNumber].applicants.push(msg.sender);
+        repos[_repoId].issues[_issueNumber].assignmentRequests[msg.sender] = _application;
```
Quazia: Do we want to do any validation to ensure the _application is an IPFS hash?

apps/projects/contracts/Projects.sol
```diff
@@ -287,6 +312,101 @@ contract Projects is AragonApp {
...
+     * @param _issueNumber the github issue up for resolution
+     * @param _contributor address of the asignee who submitted work for review
+     * @param _approved decision to accept the contribution 
+     */
+    function reviewSubmission(
+        bytes32 _repoId, 
+        uint256 _issueNumber, 
+        address _contributor,
+        bool _approved
+    ) external isInitialized auth(TASK_MGR_ROLE)
```
Quazia: Do we want to split this into two roles?

apps/projects/contracts/Projects.sol
topocount: Yeah I agree. I'll remove it

apps/projects/contracts/Projects.sol
topocount: I thought it might be good to keep a log of everyone who's submitted work, otherwise we won't be able to have any state where someone's work is rejected. Maybe it's not necessary to store a rejected submission after the task has been reassigned. 

Users can already reset the state of the submissions by resubmitting new work, so that should suffice. I'll update this as well

apps/projects/contracts/Projects.sol
topocount: yeah that makes sense. Someone who reviews work might be have different permissions than someone who accepts

apps/projects/contracts/Projects.sol
topocount: Would this be to prevent people from creating garbage submissions outside of the frontend? Not sure it's necessary if this isn't a concern.

topocount: I think it makes more sense to just check in in the FE, since string validation on-chain can be expensive

PR #417, opened 2019-02-13
400,416 fixes

rkzel:


apps/projects/app/components/Panel/NewIssue/NewIssue.js
```diff
@@ -5,18 +5,26 @@ import { Mutation } from 'react-apollo'
...
+  <div
+    style={{
+      display: 'flex',
+      justifyContent: 'center',
+      alignItems: 'center',
+      height: '100%',
+      flexDirection: 'column',
+    }}
+  >
+    <LoadingAnimation src={ethereumLoadingAnimation} />
```
Quazia: This asset isn't rendering for me, it's just showing as a broken link.

rkzel: Fixed, assets cleaned up.

PR #427, opened 2019-02-15
293 review work panel

topocount:
Please let me know if my stateless functional components need to be moved elsewhere, or if I need to modify this to conform to any other FE conventions


apps/projects/app/components/Panel/ReviewWork/ReviewWork.js
```diff
@@ -1,20 +1,224 @@
...
+          }
+        />
+
+        <FormField
+          label="Feedback"
+          input={
+            <TextInput.Multiline
+              name='feedback'
+              rows={5}
+              style={{ resize: 'none', height: 'auto' }}
```
ottodevs: The current multiline from aragon ui has no top-padding should we add a little padding or reuse the [DescriptionInput](https://github.com/AutarkLabs/planning-suite/blob/7daf53218c0af317d7ea3cd645633a525bbf4ea9/apps/allocations/app/components/Form/Input/DescriptionInput.js#L10) component that has the padding already defined? (As seen on Allocations->NewProject)
Or better let it be that way? (no padding Top-bottom) cc @stellarmagnet for this

apps/projects/app/components/Panel/ReviewWork/ReviewWork.js
```diff
@@ -1,20 +1,224 @@
...
+              onChange={this.changeField}
+              placeholder="Do you have any feedback to provide the contributor?"
+              wide
+            />
+          }
+        />
+
+        <ReviewRow>
+          <ReviewButton emphasis="negative" onClick={this.onReviewApplicationReject}>Reject</ReviewButton>
+          <ReviewButton emphasis="positive" onClick={this.onReviewApplicationAccept}>Accept</ReviewButton>
```
ottodevs: These buttons need a little horizontal separation IMHO

apps/projects/app/components/Panel/ReviewWork/ReviewWork.js
```diff
@@ -1,20 +1,224 @@
...
+  line-height: 38px;
+`
+const IssueLinkRow = styled.div`
+  height: 31px;
+  display: flex;
+  align-items: center;
+  margin-bottom: 10px;
+`
+const SubmissionDetails = styled.div`
+  border: 1px solid #DAEAEF;
```
ottodevs: Should we use theme values here? or maybe these colors are not present on the aragon ui theme

apps/projects/app/components/Panel/ReviewWork/ReviewWork.js
```diff
@@ -1,20 +1,224 @@
...
+
+function CommentsCheck(props) {
+  if (!props.comments) return null
+  else{
+    return <FieldTitle>Additional Comments</FieldTitle>
+  }
 }
 
+function RatingAlert(props) {
+  return props.alert ? <Info.Alert>Please Select A Quality Rating</Info.Alert> : null
```
ottodevs: Is it needed to have Title Case Here?

apps/projects/app/components/Panel/ReviewWork/ReviewWork.js
```diff
@@ -1,20 +1,224 @@
...
+}
+
+const IssueTitle = styled(Text)`
+  color: #717171;
```
ottodevs: Should we use theme values here? or maybe these colors are not present on the aragon ui theme

apps/projects/app/components/Panel/ReviewWork/ReviewWork.js
```diff
@@ -1,20 +1,224 @@
...
+`
+const IssueLinkRow = styled.div`
+  height: 31px;
+  display: flex;
+  align-items: center;
+  margin-bottom: 10px;
+`
+const SubmissionDetails = styled.div`
+  border: 1px solid #DAEAEF;
+  background-color: #F3F9FB;
```
ottodevs: Should we use theme values here? or maybe these colors are not present on the aragon ui theme

apps/projects/app/components/Panel/ReviewWork/ReviewWork.js
```diff
@@ -1,20 +1,224 @@
...
+  margin-bottom: 14px;
+`
+const UserLink = styled.div`
+  display: flex;
+  align-items: center;
+`
+const Separator = styled.hr`
+  height: 1px;
+  width: 100%;
+  color: #D1D1D1;
```
ottodevs: Should we use theme values here? or maybe these colors are not present on the aragon ui theme

apps/projects/app/components/Panel/ReviewWork/ReviewWork.js
```diff
@@ -1,20 +1,224 @@
+
```
ottodevs: This empty line shouldn't be here

apps/projects/app/components/Panel/ReviewWork/ReviewWork.js
```diff
@@ -1,20 +1,224 @@
...
+  }
+
+  generateRatingChange = () => index => {
+    this.setState({rating: index})
+    console.log('index: ', index)
+  }
+
+  render() {
+    const { issue } = this.props
+    const submittor = work.user
```
ottodevs: Is submittor an acceptable word? I am leaning more towards submitter but this is just an opinion

apps/projects/app/components/Panel/ReviewWork/ReviewWork.js
rkzel: Yes, I noticed that yesterday actually, updated the other panels like this:

https://github.com/AutarkLabs/planning-suite/tree/414-fe-tweaks

apps/projects/app/components/Panel/ReviewWork/ReviewWork.js
topocount: Changed the styled button width from a raw pixel value to a percentage, so they should be separated in any view context

PR #430, opened 2019-02-15
[WIP]296 bounty allocation

Quazia:
There's still some changes that need to be made here but all of the _integration_ is done. There's some front end elements that can be handled in the panel integration such as how to handle the various bounty states. There's also an issue with the way that tokens are being handled in the settings; the default token doesn't have an address so you need to switch the token. Could probably be resolved by a re-work of the initial settings but this is already a massive PR so as much as possible we can do further tweaking separately.


topocount:
@Quazia starting the review. Do you have some time to get the contract tests passing again? Looks like the address running the tests is just missing a role assignment.

topocount:
@Quazia I'm a good chunk of the way reading through the code, but it'd also be useful to document your test workflow in the test script.

apps/projects/app/components/App/App.js
```diff
@@ -219,6 +223,88 @@ class App extends React.PureComponent {
     }))
   }
 
+  onSubmitBountyAllocation = async issues => {
+    console.log('repos:', this.props.repos)
+    console.log('bounty allocation submitted', issues)
+    let bountySymbol = this.props.bountySettings.bountyCurrency
+    let bountyToken
+    this.props.tokens.forEach(
```
topocount: why not use map here?

apps/projects/contracts/Projects.sol
```diff
@@ -428,24 +469,28 @@ contract Projects is AragonApp {
...
         // ensure the transvalue passed equals transaction value
-        checkTransValueEqualsMessageValue(msg.value, _bountySizes,_tokenBounties);
-        
+        //checkTransValueEqualsMessageValue(msg.value, _bountySizes,_tokenBounties);
         string memory ipfsHash;
         uint standardBountyId;
         // submit the bounty to the StandardBounties contract
         for (uint i = 0; i < _bountySizes.length; i++) {
-            ipfsHash = substring(_ipfsAddresses, i.mul(46), i.add(1).mul(46));
+            ipfsHash = "";//substring(_ipfsAddresses, i.mul(46), i.add(1).mul(46));
```
topocount: the hash should be passed in now that the integration is complete, right?

apps/projects/contracts/Projects.sol
```diff
@@ -26,6 +29,18 @@ import "@aragon/os/contracts/lib/math/SafeMath.sol";
 * applying bounties in bulk and accepting fulfillment via this contract
 *******************************************************************************/
 interface Bounties {
+
+  function issueAndActivateBounty(
```
topocount: This can probably be removed from the interface

ottodevs:
@Quazia , just if we have time to keep travis green, CI fails because 2 tests seem to be failing: 

```
          "before each" hook: issue bulk bounties for "verifies bounty data contains correct IPFS hashes":
     Error: VM Exception while processing transaction: revert APP_AUTH_FAILED

  2) Contract: Projects App
       invalid operations
         cannot accept unfulfilled bounties:
     Error: VM Exception while processing transaction: revert APP_AUTH_FAILED
```

Also package-lock.json size has changed too much, maybe we should double check by regenerating it with `npm run clean:commit` (just deletes the old package-lock.json and regenerates a new one by bootstrapping, is not really long ~3m)

apps/projects/app/components/App/App.js
Quazia: Bad form to use map if you arn't returning an array.

apps/projects/contracts/Projects.sol
Quazia: Yeah, the substring function crashes it since it's not in constant time. I'll fix the substring function and then re-enable this.

apps/projects/contracts/Projects.sol
Quazia: Agreed

apps/projects/app/components/Content/Settings.js
```diff
@@ -5,16 +5,19 @@ import NumberFormat from 'react-number-format'
 
 const bountyDeadlines = ['Weeks', 'Days', 'Hours']
 const bountyDeadlinesMul = [168, 24, 1] // it is one variable in contract, so number * multiplier = hours
-const bountyCurrencies = ['BTC', 'ETH', 'TL', 'ANT', '🦄']
 
 class Settings extends React.Component {
-  state = {}
+  state = {
+    bountyCurrencies: this.props.tokens.map (token => token.symbol)
```
rkzel: This fails with TypeError: r.props.tokens.map is not a function message on my system.

ottodevs: We should not set the state that way, why not use the props directly? Is it really needed to extra cache the immutable props into the state? It is a react anti-pattern and should be refactored

PR #460, opened 2019-02-22
421 review applicant integration

Quazia:
Address #421

### DIFF
--- a/apps/projects/app/components/Card/Issue.js
+++ b/apps/projects/app/components/Card/Issue.js
@@ -105,15 +105,21 @@ const Issue = ({workStatus, title, repo, number, labels, isSelected, onClick, on
       }
       {workStatus !== undefined &&
         <ContextMenu>
-          <ContextMenuItem onClick={onSubmitWork}>
-            <ActionLabel>Submit Work</ActionLabel>
-          </ContextMenuItem>
-          <ContextMenuItem onClick={onRequestAssignment}>
-            <ActionLabel>Request Assignment</ActionLabel>
-          </ContextMenuItem>
-          <ContextMenuItem onClick={onReviewApplication}>
-            <ActionLabel>Review Application</ActionLabel>
-          </ContextMenuItem>
+          {workStatus === 'submit-work' || workStatus === 'review-work' &&
+            <ContextMenuItem onClick={onSubmitWork}>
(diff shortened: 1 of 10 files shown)


apps/projects/app/script.js
```diff
@@ -128,13 +128,25 @@ async function handleEvents(response) {
   case 'RepoUpdated':
     console.log('[Projects] RepoUpdated', response.returnValues)
     nextState = await syncRepos(appState, response.returnValues)
+  case 'AssignmentApproved':
```
topocount: I was able to crash the app with the following error message in the console after approving an assignment during the test script: `getRepoData failed:  TypeError: Cannot read property 'request' of undefined` 

Not sure why the client variable declared above would fail to be initialized

apps/projects/app/components/Content/IssueDetail/components/Detail.js
```diff
@@ -210,8 +217,17 @@ const Detail = ({
               </Text.Block>
             </div>
             <div style={{ ...column, flex: 0, alignItems: 'flex-end' }}>
-              <DropDownButton enabled />
-              {/* <Text>$500 USD</Text> */}
+              <DropDownButton enabled>
```
topocount: The dropdowns both here and on the issue detail card stopped rendering on click after the app crashed. I think it's not a coincidence, but it's worth noting

apps/projects/app/components/Card/Issue.js
```diff
@@ -105,15 +105,21 @@ const Issue = ({workStatus, title, repo, number, labels, isSelected, onClick, on
...
-          <ContextMenuItem onClick={onSubmitWork}>
-            <ActionLabel>Submit Work</ActionLabel>
-          </ContextMenuItem>
-          <ContextMenuItem onClick={onRequestAssignment}>
-            <ActionLabel>Request Assignment</ActionLabel>
-          </ContextMenuItem>
-          <ContextMenuItem onClick={onReviewApplication}>
-            <ActionLabel>Review Application</ActionLabel>
-          </ContextMenuItem>
+          {workStatus === 'submit-work' || workStatus === 'review-work' &&
```
topocount: I think the workstatus here is causing no options. I'm guessing it's because the logic isn't grouped here

PR #461, opened 2019-02-22
418 submit work integration

topocount:
Had to make some modifications to the projects and standardBounties.sol contract to get this working. We definitely need to look into the other integration option for M5

### DIFF
--- a/apps/projects/app/script.js
+++ b/apps/projects/app/script.js
@@ -140,6 +140,17 @@ async function handleEvents(response) {
     nextState = syncIssues(appState, response.returnValues, data)
     appState = nextState
     break
+  case 'WorkSubmitted':
+    console.log('[Projects] WorkSubmitted', appState, response.returnValues)
+    if(response.returnValues === null || response.returnValues === undefined) {
+      break
+    }
+    data = await loadIssueData(response.returnValues)
+    const submissionData = await loadSubmissionData(response.returnValues, data.assignee)
+    data.workStatus = status[2]
+    data.work = submissionData
+    nextState = syncIssues(appstate, response.returnValues, data)
+
   case 'BountyAdded':
     console.log('[Projects] BountyAdded', appState, response.returnValues)
     if(response.returnValues === null || response.returnValues === undefined) {
@@ -270,13 +281,13 @@ function loadRepoData(id) {
 
 function loadIssueData({repoId, issueNumber}) {
   return new Promise(resolve => {
-    app.call('getIssue', repoId, issueNumber).subscribe(({ hasBounty, standardBountyId, balance, token, dataHash}) => {
+    app.call('getIssue', repoId, issueNumber).subscribe(({ hasBounty, standardBountyId, balance, token, dataHash, assignee}) => {
       let contentJSON
       ipfs.get(dataHash, (err, files) => {
         for(const file of files) {
           contentJSON = JSON.parse(file.content.toString('utf8'))
         }
-        resolve({ balance, hasBounty, token, standardBountyId, ...contentJSON})
+        resolve({ balance, hasBounty, token, standardBountyId, assignee, ...contentJSON})
       })      
     })
   })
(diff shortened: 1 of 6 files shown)


apps/projects/app/script.js
```diff
@@ -296,15 +308,27 @@ function loadRequestsData({repoId, issueNumber}) {
...
-    app.call('getApplicant', repoId, issueNumber, applicantId).subscribe( (address) => {
-      app.call('getAssignmentRequest', repoId, issueNumber, address).subscribe( (hash) => {
-        let contentJSON
-        ipfs.get(hash, (err, files) => {
-          for(const file of files) {
-            contentJSON = JSON.parse(file.content.toString('utf8'))
-          }
-          resolve({contributorAddr: address, ...contentJSON})
-        })
+    app.call('getApplicant', repoId, issueNumber, applicantId).subscribe( ({address, hash}) => {
```
topocount: I believe the subscription should be passed an object from a call the returns more than one parameter

apps/projects/contracts/Projects.sol
```diff
@@ -393,24 +400,27 @@ contract Projects is IsContract, AragonApp {
      * @param _repoId the github repo id of the issue
      * @param _issueNumber the github issue up for assignment
      * @param _submissionAddress IPFS hash of the Pull Request
-     * @param _fulfillmentId retrieved from event after the work is submitted to the bounties contract externally 
+     * //param _fulfillmentId retrieved from event after the work is submitted to the bounties contract externally
```
topocount: This needs to be uncommented once all branches are merged

apps/projects/app/script.js
```diff
@@ -296,15 +308,27 @@ function loadRequestsData({repoId, issueNumber}) {
...
+        }
+        resolve({contributorAddr: address, ...contentJSON})
+      })
+    })
+  })
+}
+
+function loadSubmissionData({issueNumber, repoId}, assignee) {
+  return new Promise(resolve => {
+    app.call('getSubmission', repoId, issueNumber, assignee).subscribe(async ({submissionHash, fulfillmentId, status}) => {
```
topocount: the function passed as the subscription to getSubmission probably doesn't need to be async

apps/projects/contracts/Projects.sol
```diff
@@ -393,24 +400,27 @@ contract Projects is IsContract, AragonApp {
...
-        string _submissionAddress,
-        uint256 _fulfillmentId
+        string _submissionAddress
+        //uint256 _fulfillmentId
     ) external isInitialized
     {
-        require(msg.sender == repos[_repoId].issues[_issueNumber].assignee, "User not assigned to this issue");
-
-        repos[_repoId].issues[_issueNumber].workSubmissions[msg.sender] = WorkSubmission(
+        //require(msg.sender == repos[_repoId].issues[_issueNumber].assignee, "User not assigned to this issue");
```
topocount: Should be uncommented before merge

topocount:
TODO: I've realized that githubuser data needs to be passed into the submission before it's created in IPFS.

topocount:
Added the relevant test steps to the test script.

apps/projects/test/projects.test.js
```diff
@@ -1,4 +1,5 @@
 /* global artifact, ... */
+console.log('Hello Kevin')
```
Quazia: Probably don't need this anymore

apps/projects/contracts/Projects.sol
```diff
@@ -393,24 +400,27 @@ contract Projects is IsContract, AragonApp {
...
+        GithubIssue storage issue = repos[_repoId].issues[_issueNumber];
```
Quazia: Might be better to have this before the require and then check against it instead of pulling the data twice but there's an argument to be made for both ways.

apps/projects/app/script.js
Quazia: Yeah... try it without it and if it works then delete it?

apps/projects/app/script.js
Quazia: Yeah this is correct, can we change the variable names to the return names from getApplicant for clarity or do you think this is more clear? i.e. applicant application

PR #474, opened 2019-02-25
473 hotfix execute vote

Quazia:
Addresses #473


apps/range-voting/app/components/Panels/VotePanelContent.js
```diff
@@ -78,11 +78,9 @@ class VotePanelContent extends React.Component {
       combineLatest(tokenContract.balanceOf(user), tokenContract.decimals())
         .first()
         .subscribe(([balance, decimals]) => {
-          const adjustedBalance = Math.floor(
-            parseInt(balance, 10) / Math.pow(10, decimals)
-          )
           this.setState({
-            userBalance: adjustedBalance,
+            userBalance: balance,
+            decimals: decimals
```
ottodevs: Do we really want to have something like decimal, that is constant on each token contract into our hot state? I think this should come as a const from the contract better, what do you think?

apps/range-voting/app/components/Panels/VotePanelContent.js
```diff
@@ -261,7 +259,10 @@ class VotePanelContent extends React.Component {
             <h2>
               <Label>Your voting tokens</Label>
             </h2>
-            {BigNumber(this.state.userBalance).toString()}
+            {BigNumber(this.state.userBalance)
+              .div(BigNumber(this.state.decimals))
+              .dp(3)
+              .toString()}
```
ottodevs: There is no clear advantage on having these values into the component state, instead I suggest to calculate the balance as it should be shown on the view and pass it as a prop from the background worker, should that make sense? We can let this way for a while but this seems like technical debt

apps/range-voting/app/components/Slider.js
```diff
@@ -52,7 +52,7 @@ class Slider extends React.Component {
   updateValueFromClientX(clientX) {
     const rect = this.getRect()
     const x = parseFloat(
-      Math.min(rect.width, Math.max(0, clientX - rect.x)).toFixed(2)
+      Math.min(rect.width, Math.max(0, clientX - rect.x))
```
ottodevs: If you are not using toFixed then the value will not be converted to number, so it does not need to use parseFloat anymore

apps/range-voting/app/components/Slider.js
```diff
@@ -112,7 +112,7 @@ class Slider extends React.Component {
   render() {
     const { pressed } = this.state
     const value = parseFloat(
-      Math.max(0, Math.min(1, this.props.value)).toFixed(2)
+      Math.max(0, Math.min(1, this.props.value))
```
ottodevs: Same here, Math.max outputs a number, no need to use parseFloat anymore

apps/range-voting/app/components/Slider.js
```diff
@@ -52,7 +52,7 @@ class Slider extends React.Component {
   updateValueFromClientX(clientX) {
     const rect = this.getRect()
     const x = parseFloat(
-      Math.min(rect.width, Math.max(0, clientX - rect.x))
+      Math.floor(Math.min(rect.width, Math.max(0, clientX - rect.x)) * 1000 ) / 1000
```
ottodevs: This will help with values with more than 4 decimal figures like 5.0005 that will be 5000.5 then floor to 5000 and become 5 after dividing again by 1000 but seem not really accurate to cover all cases

apps/range-voting/app/components/Slider.js
```diff
@@ -112,7 +112,7 @@ class Slider extends React.Component {
   render() {
     const { pressed } = this.state
     const value = parseFloat(
-      Math.max(0, Math.min(1, this.props.value))
+      Math.floor(Math.max(0, Math.min(1, this.props.value)) * 1000) / 1000
```
ottodevs: Same is going here

apps/allocations/contracts/Allocations.sol
```diff
@@ -174,13 +174,12 @@ contract Allocations is AragonApp, Fundable {
...
     function runPayout(uint256 _payoutId) external payable isInitialized auth(EXECUTE_PAYOUT_ROLE) returns(bool success) {
         Payout storage payout = payouts[_payoutId];
-        uint256 pointsPer;
         uint256 totalSupport;
         uint i;
         for (i = 0; i < payout.supports.length; i++) {
             totalSupport += payout.supports[i];
         }
-
+        totalSupport = totalSupport.div(10000);
```
ottodevs: We should not use random operations with literal values, better use constants and define the behavior on this operation

apps/allocations/app/components/Card/Account.js
```diff
@@ -14,6 +14,8 @@ import {
   theme,
 } from '@aragon/ui'
 
+const ETH_DECIMALS = BigNumber(10e17)
```
ottodevs: Better reuse common constants into maybe utils/constant file instead of having the same value declared over an over on several files

apps/range-voting/app/components/Panels/VotePanelContent.js
```diff
@@ -152,8 +152,8 @@ class VotePanelContent extends React.Component {
     } = vote.data
 
     // TODO: Show decimals for vote participation only when needed
-    const voterParticipation = (participationPct * 10e15).toFixed(2)
-
+    const disParticipationPct = (participationPct).toFixed(2)
```
ottodevs: disParticipation and disMinParticipation variable names seem not really helpful to understand their meaning, could we use more representative names for these variables?

apps/allocations/contracts/Allocations.sol
Quazia: Outdated

apps/range-voting/app/components/Panels/VotePanelContent.js
Quazia: I'm just doing it how it's done in the A1 vote panel.

Quazia: I'm just doing it the way it's done in the A1 panels.

apps/range-voting/app/components/Panels/VotePanelContent.js
Quazia: It's supposed to be short for 'display' but I can write the full thing out?

Issue #464, opened 2019-02-22
Reset checkmarks after allocating bounty


stellarmagnet:
Steps to reproduce:

Visit Projects app; Issues tab
Select multiple issues
Allocate bounty
Submit

You will see that the checkmarks are still selected

PR #485, opened 2019-02-27
#464, deselect all issues after they are used for populating a panel

rkzel:
Trivial UX change


stellarmagnet:
@rkzel it looks like you are resetting the checkmarks after the user selects the action from the action menu, instead of after the user hits "submit" on side panels.

The reset should happen after submitting in the side panel, I think.

Opinion: @ottodevs, @Quazia, @topocount?

Issue #351, opened 2019-01-29
If I already have a repo connected to my projects app, this repo should not be displayed in the "New Project" panel 


stellarmagnet:
If I already have a repo connected to my projects app, this repo should not be displayed in the "New Project" panel whenever I try to add a subsequent project.

Steps to reproduce:
1) Navigate to New Project, select a repo to connect it to Projects app
2) Navigate to New Project again. You will see that your currently connected repo is still being displayed.

Note: in my previous designs, this was a checkbox selection, and previously connected repos would show up as a checkbox that has been checked off, and you cant click on it. But with the radio button route, we probably should just not display them at all.

Quazia:
Can you explain how to duplicate this issue? Is this just a matter of making sure repos that have already been added arn't displayed?

stellarmagnet:
@Quazia yes that's what I mean. I just updated the ticket for clarity

PR #486, opened 2019-02-27
#351, removed projects already added from list of repos in NewProject…

rkzel:


apps/projects/app/components/Panel/NewProject/RepoSelector.js
```diff
@@ -128,7 +132,11 @@ class Repo extends React.Component {
         </div>
       )
 
-    const repositories = repos.length > 0 ? repoList : <Loading />
+    // if there are repos to add - list them. otherwise either all repos
+    // are already added or we're still waiting for the list from external
+    const repositories = repos.length > 0 ? repoList :
```
Quazia: Can we maybe just make this into a function? The chained ternary is sort of hard to read. I'm fine keeping it as is if you think it's better this way but to me it seems like a little too much logic to chain.

rkzel: You are right, nested ternaries are nasty. Improved (and saved myself from a lot of future confusion).

apps/projects/app/components/Panel/NewProject/RepoSelector.js
```diff
@@ -85,50 +91,54 @@ class Repo extends React.Component {
...
+        There are no repositories matching{' '}
+        <Text weight="bold">{this.state.filter}</Text>
+      </Text.Block>
+      <ClearSearch onClick={this.handleClearSearch}>
+        Clear Search
+      </ClearSearch>
+    </div>
+  )
+
+  stillLoading = () => !(this.props.data && this.props.data.viewer && this.props.data.viewer.repositories.edges.length > 0)
```
Quazia: Can you try and reduce the line length here? This is still hard to read...

apps/projects/app/components/Panel/NewProject/RepoSelector.js
```diff
@@ -85,50 +91,54 @@ class Repo extends React.Component {
...
-          <Text.Block>
-            There are no repositories matching{' '}
-            <Text weight="bold">{this.state.filter}</Text>
-          </Text.Block>
-          <ClearSearch onClick={this.handleClearSearch}>
-            Clear Search
-          </ClearSearch>
-        </div>
-      )
+    // if there are repos to add - list them. otherwise either all repos
```
Quazia: Much better for this part IMO

apps/projects/app/components/Panel/NewProject/NewProject.js
```diff
@@ -27,7 +29,7 @@ class NewProject extends React.Component {
       : this.props.onGithubSignIn
 
     if (this.state.started)
-      return <RepoSelector onCreateProject={this.props.onCreateProject} />
+      return <RepoSelector onCreateProject={this.props.onCreateProject} reposAlreadyAdded={this.props.reposAlreadyAdded}/>
```
ottodevs: It seems you forgot an space character before closing tag

apps/projects/app/components/Panel/NewProject/RepoSelector.js
```diff
@@ -22,20 +22,26 @@ class Repo extends React.Component {
     repoSelected: -1,
   }
 
+  filterAlreadyAdded = repos => {
+    if (this.props.reposAlreadyAdded === undefined) return repos
+    const reposAlreadyAddedIds = this.props.reposAlreadyAdded.map(repo => repo.data._repo)
```
ottodevs: If we just need `repo.data._repo` fields for that, maybe we should just pass an array of this field as props instead of the whole repos data.
This way we should just need to use `Array.filter()` function to filter out repos with the same id as `reposAlreadyAddedIds`

apps/projects/app/components/Panel/NewProject/RepoSelector.js
```diff
@@ -22,20 +22,26 @@ class Repo extends React.Component {
...
+    return repos.filter(repo => reposAlreadyAddedIds.indexOf(repo.node.id) === -1)
```
ottodevs: IMHO `Array.some()` or `Array.includes()` seem more readable for these cases instead of `Array.indexOf()` and then compare index

apps/projects/app/components/Panel/NewProject/RepoSelector.js
```diff
@@ -83,50 +89,54 @@ class Repo extends React.Component {
...
+        There are no repositories matching{' '}
+        <Text weight="bold">{this.state.filter}</Text>
+      </Text.Block>
+      <ClearSearch onClick={this.handleClearSearch}>
+        Clear Search
+      </ClearSearch>
+    </div>
+  )
+
+  stillLoading = () => !(this.props.data && this.props.data.viewer && this.props.data.viewer.repositories.edges.length > 0)
```
ottodevs: [Apollo handles loading state on its own](https://www.apollographql.com/docs/react/essentials/queries.html#loading), so we should not use custom loading conditions at all

apps/projects/app/components/Panel/NewProject/RepoSelector.js
```diff
@@ -83,50 +89,54 @@ class Repo extends React.Component {
...
-        </div>
-      )
+    // if there are repos to add - list them. otherwise either all repos
+    // are already added or we're still waiting for the list from external
+    let repositoriesDisplay
 
-    const repositories = repos.length > 0 ? repoList : <Loading />
+    if (repos.length > 0)
+      repositoriesDisplay = this.repoList(visibleRepos, repoArray)
+    else if (this.stillLoading())
```
ottodevs: It maybe should be better to use[ Apollo loading state flow](https://www.apollographql.com/docs/react/essentials/queries.html#loading) here

apps/projects/app/components/Panel/NewProject/RepoSelector.js
```diff
@@ -265,6 +275,19 @@ const Loading = () => (
...
+  <div
+    style={{
+      display: 'flex',
+      justifyContent: 'center',
+      alignItems: 'center',
+      height: '20%',
+      flexDirection: 'column'
+    }}
+  >
+    <Text>No more repos to add...</Text>
```
ottodevs: Maybe just an opinion, but for the sake of consistency we should use same denomination like "repositories" at L108, or change that if we decide to use "repos"

apps/projects/app/components/Panel/NewProject/RepoSelector.js
```diff
@@ -285,8 +308,8 @@ query {
    repositories(
      first: 100,
      orderBy: {field: UPDATED_AT, direction: DESC}
-     ownerAffiliations: [OWNER, COLLABORATOR, ORGANIZATION_MEMBER],
-     affiliations: [OWNER, COLLABORATOR, ORGANIZATION_MEMBER]) {
+     ownerAffiliations: [OWNER, COLLABORATOR],
```
ottodevs: Could you please elaborate on why is this change needed?

apps/projects/app/components/Panel/NewProject/RepoSelector.js
rkzel: Improved, although it still looks not very smart. There must be a better way of chaining conditions like that: if X and if X.Y and if X.Y.Z...

apps/projects/app/components/Panel/NewProject/NewProject.js
rkzel: Duly noted :)

apps/projects/app/components/Panel/NewProject/RepoSelector.js
rkzel: Repositories. Changed this line, consistency is important (and looks better).

apps/projects/app/components/Panel/NewProject/RepoSelector.js
rkzel: Oh sorry, forgot to change it back after testing.

apps/projects/app/components/Panel/NewProject/RepoSelector.js
rkzel: Done.

apps/projects/app/components/Panel/NewProject/RepoSelector.js
rkzel: Improved.

apps/projects/app/components/Panel/NewProject/RepoSelector.js
rkzel: Done.

apps/projects/app/components/Panel/NewProject/RepoSelector.js
rkzel: Done.

PR #492, opened 2019-02-28
fixed removing repo functionality

rkzel:
One thing I'm not sure of is why exactly is it emitting RepoUpdated event (in Projects.sol) for repo that was moved in repoIndex array?

emit RepoUpdated(repoToMove, repos[repoToMove].owner, rowToDelete);

It doesn't seem to be doing much...


apps/projects/contracts/Projects.sol
```diff
@@ -302,6 +302,7 @@ contract Projects is IsContract, AragonApp {
      * @param _owner Github id of the entity that owns the repo to add
      * @param _repoId Github id of the repo to add
      * @return index for the added repo at the registry
+     * // RepoUpdated event doesn't seem useful at the moment
```
topocount: We should scrub the RepoUpdated event and its definition completely from the contract and save the gas

ottodevs:
> One thing I'm not sure of is why exactly is it emitting RepoUpdated event (in Projects.sol) for repo that was moved in repoIndex array?
> 
> emit RepoUpdated(repoToMove, repos[repoToMove].owner, rowToDelete);
> 
> It doesn't seem to be doing much...

I added that part!
I will try to explain briefly, we use a unordered struct to store the packages and an additional mapping of repoId (since those they are unique bytes) to Repo objects, as a registry index to be able to iterate our registry.
When a Repo is removed, we delete from the registry and then move the last repo of the index to fill the mapping gap of the old repo, that is why removing a repo perform 2 operations, remove the old one and change the last index to fill the gap, I am not sure if it is the most efficient way to that by most Solidity literature does it this way, also has some gas optimizations during the process

PR #497, opened 2019-02-28
M2 review allocations store

ottodevs:
Allocations store, solves m2 review and sets the path for store structure consisting in event handlers for each entity managed and common initializer patterns for every app, providing an easy way to hook events from external apps or abstracting the initialization and store setup process.

Code and configs are very inspired from payroll app and other aragon apps.


apps/allocations/app/store/entry.js
```diff
@@ -0,0 +1,54 @@
...
+  const addressBookApp = app.external(addressBook, addressBookAbi)
+
+  return addressBookApp
+    .getEntry(addr)
+    .first()
+    .map(
+      entry =>
+        // cover removed entries
+        entry === undefined
+          ? 'removed'
```
Quazia: Do we actually need to track that it's removed with this field? Would it be possible to just maintain a data structure that strictly contains the current entities or would that make things more complicated?

ottodevs: Solved 👍

apps/allocations/app/store/init.js
```diff
@@ -0,0 +1,35 @@
...
+      if (!state) state = initialState
+
+      try {
+        const next = await handleEvent(state, event)
+        const nextState = { ...initialState, ...next }
+        // Debug point
+        console.log('[Allocations store]', nextState)
+        return nextState
+      } catch (err) {
+        console.error('[Allocations script] initStore', event, err)
```
topocount: I'm getting an error thrown by this line. when I execute the `runpayout` transaction and the account balance doesn't update. However, when I clean the cache in the aragon wrapper, The event passes through successfully and the allocations account balance updates to the correct value. The error is as follows:

`[Allocations script] initStore {logIndex: 0, transactionIndex: 0, transactionHash: "0x7e26226963c93531592a4baf1656c16abb379eabf2382f5df11220787c90efbf", blockHash: "0xec23d32ddd50989c86759d2e08675ece620ced40fb3b452ab072933c0eac70d5", blockNumber: 73, …} TypeError: argument is not a function. Are you looking for mapTo()?`

PR #509, opened 2019-03-04
NewBounty panel, changed Hours field from DropDown to text.numeric

rkzel:


apps/projects/app/components/Panel/NewBountyAllocation/NewBountyAllocation.js
```diff
@@ -246,6 +246,12 @@ class NewBountyAllocation extends React.Component {
   }
 }
 
+const HoursInput = styled(TextInput.Number)`
```
ottodevs: We could use [attrs](https://www.styled-components.com/docs/api#attrs) here to pass some initial attributes that should help to validate the field, like min, max or step:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#Additional_attributes
Then maybe we could save some conditional code at lines 91-94

rkzel: Great tip - added :)

stellarmagnet:
I'm not sure if this is the culprit, but I am unable to allocate a bounty. Did you allocate a bounty after this fix and it worked? I hit "allocate bounties" and there is no transaction to sign, panel just closes.

PR #510, opened 2019-03-05
479 review work action

rkzel:


rkzel:
Ok, so: "work" prop is what is inserted into Issue when a person who applied for assignment, approved and submitted "work" - this is what the variable is called in the Issue object, so I left it like that. it gets passed in {...issue}. Its presence serves as true/false defined/undefined check by itself, so no additional bool variable seems required? And it does work as intended (the intention is that BountyContextMenu shows only available options - when no work has been submitted it is "Submit Work" only, when work has been submitted it is both "Submit" and "Review".)
PropsTypes - added.

ottodevs:
> Ok, so: "work" prop is what is inserted into Issue when a person who applied for assignment, approved and submitted "work" - this is what the variable is called in the Issue object, so I left it like that. it gets passed in {...issue}. Its presence serves as true/false defined/undefined check by itself, so no additional bool variable seems required? And it does work as intended (the intention is that BountyContextMenu shows only available options - when no work has been submitted it is "Submit Work" only, when work has been submitted it is both "Submit" and "Review".)
> PropsTypes - added.

Thanks for the explanation, what I was trying to clarify is why `work` prop was not passed to those component and what type of entity it was because without propTypes and no value assigned it was not clear if it was an object, array, or boolean prop.

ottodevs:
> Ok, so: "work" prop is what is inserted into Issue when a person who applied for assignment, approved and submitted "work" - this is what the variable is called in the Issue object, so I left it like that. it gets passed in {...issue}. Its presence serves as true/false defined/undefined check by itself, so no additional bool variable seems required? And it does work as intended (the intention is that BountyContextMenu shows only available options - when no work has been submitted it is "Submit Work" only, when work has been submitted it is both "Submit" and "Review".)
> PropsTypes - added.

A prop was introduced into 3 components we should add proptypes to any component that is receiving that prop.
Also I don't understand when is work prop inserted into Issue, it is not apparently not passed into parent AppComponent, maybe I am missing something?

apps/projects/app/components/Shared/BountyContextMenu.js
```diff
@@ -42,6 +42,10 @@ BountyContextMenu.propTypes = {
   onReviewApplication: PropTypes.func.isRequired,
   onReviewWork: PropTypes.func.isRequired,
   workStatus: PropTypes.oneOf([undefined, 'new', 'review-applicants', 'submit-work', 'review-work', 'finished']),
+  work: PropTypes.oneOf([
```
ottodevs: In this case it should be required, then you probably will get a prop error since Issues and Details are not receiving the work prop (they should have proptypes as well to help detect this bug), also I don't like the solution of passing work with "undefined" value

rkzel: That's because the "work" is embedded in an Issue and passed implicitly around - in Content/Issues.js it is passed down to IssueDetail and Card via unfolding of {...issue}. It is undefined when there is no bounty or no work has been submitted yet - otherwise it is {}.
Now, I admit that having PropTypes check like "it's there or not" is not good - maybe 'work' property can be set to an empty {} right when the Issue is first mentioned? But then again it's cleaner to check if it is defined or not instead of checking if {} is empty or not? In any case there are no errors showing up on my console.

ottodevs: Ok now I understand it is embedded inside the issue object, ok, yes, I agree even this solution does not seen conventional, it works so I will approve the PR.
great job 👍

PR #512, opened 2019-03-06
styling of baseRate inputs

rkzel:


apps/projects/app/components/Content/Settings.js
```diff
@@ -340,17 +328,23 @@ const BaseRate = ({ baseRate, onChange }) => (
       Define your organization’s hourly rate. This is multiplied by the bounty
       size and converted into the bounty currency under the hood.
     </Text.Block>
-    <Field label="RATE PER HOUR">
+    <FieldTitle style={{marginBottom: '0px'}}>Rate per hour</FieldTitle>
```
ottodevs: For consistence and convention zero values are passed without units on styles

apps/projects/app/components/Content/Settings.js
```diff
@@ -340,17 +328,23 @@ const BaseRate = ({ baseRate, onChange }) => (
...
+    <StyledInputDropDown style={{marginBottom: '0px'}}>
```
ottodevs: same about zero values on styles

apps/projects/app/components/Content/Settings.js
```diff
@@ -340,17 +328,23 @@ const BaseRate = ({ baseRate, onChange }) => (
...
       <NumberFormat
         customInput={StyledNumberInput}
         fixedDecimalScale
         decimalScale={2}
         value={baseRate}
         allowNegative={false}
-        onChange={onChange}
+        onChange={onChangeRate}
+        style={{marginRight: '0px'}}
```
ottodevs: same about zero values

apps/projects/app/components/Content/Settings.js
```diff
@@ -331,7 +299,27 @@ const BountyContractAddress = ({ bountyAllocator }) => (
   </div>
 )
 
-const BaseRate = ({ baseRate, onChange }) => (
+const StyledInputDropDown = styled.div`
```
ottodevs: We should avoid duplicating code and use the existing InputDropDown component instead

apps/projects/app/components/Content/Settings.js
rkzel: A-ha. Sure.

apps/projects/app/components/Content/Settings.js
rkzel: I have tried that, but InputDropDown has predefined Text Input field, and in this case the field is NumericInput instead. Perhaps InputDropDown needs to be generalised to accept arbitrary fields, but it felt out of scope...

ottodevs:
LGTM I basically agree Quazia's comment

PR #521, opened 2019-03-09
429 and 442 - Rewards App Tabs and Empty State Card

### LINKED ISSUE #442: My Rewards tab: No data use case
If the user visits My Rewards tab and they their account has never been awarded a reward, then display a no data screen.

Icon: https://www.dropbox.com/s/esi2v1k7mm565k4/empty-rewards.svg?dl=0
Text primary: You have not been awarded any rewards.
Text secondary: Get started now by creating a new reward.
Button: New Reward

### [See Design](https://projects.invisionapp.com/d/main#/console/14716432/347910834/preview)

topocount:
I just used the aragon-ui TabBar component to get these created. also created the empty state card in this branch to resolve #442


apps/rewards/app/components/App/App.js
```diff
@@ -1,27 +1,51 @@
...
 class App extends React.Component {
   static propTypes = {
     app: PropTypes.object.isRequired,
-    accounts: PropTypes.arrayOf(PropTypes.object),
+    rewards: PropTypes.arrayOf(PropTypes.object),
+  }
+
+  state = {
+    selected: 0,
+    rewardsEmpty: this.props.rewards === undefined || this.props.rewards === []
```
rkzel: This way you may need extra functionality to track changes in props - if new rewards are added rewardsEmpty will need to be explicitly updated. Alternatively this.props.rewards === undefined can be checked directly in render() of App or, ultimately, in renders of specific tabs.

apps/rewards/app/components/App/App.js
```diff
@@ -1,27 +1,51 @@
...
+  }
+
   render() {
 
     return (
       <StyledAragonApp>
         <Title text="Rewards" />
+        <TabBar
+          items={[ 'Overview', 'My Rewards' ]}
+          selected={this.state.selected}
```
rkzel: Please create those tabs? In separate files, like components/Content in Projects.

apps/rewards/app/components/App/App.js
topocount: The new `Overview` and MyRewards` components now contain all empty-case logic, so this is much simpler

apps/rewards/app/components/App/App.js
topocount: I've created them and they're hooked up to the tabs as well.

apps/address-book/test/address-book.js
```diff
@@ -95,14 +95,14 @@ contract('AddressBook App', accounts => {
...
   context('invalid operations', () => {
-    let [borg, jeanluc] = accounts.splice(1, 2)
+    let [ borg, jeanluc ] = accounts.splice(1, 2)
     before(async () => {
       app.addEntry(borg, 'Borg', 'Individual')
     })
 
     it('should revert when adding duplicate address', async () => {
       assertRevert(async () => {
-        await app.addEnty(borg, 'Burg', 'N/A')
```
Quazia: This makes me think at some point it might be good to do more specific revert handling now that there are revert messages?

topocount: I now _always_ check that the revert message is as expected before wrapping assertrevert around a transaction attempt to ensure that the revert is happening for the expected reason

Issue #496, opened 2019-02-28
Add 2x more padding for Projects cards between description and commit info


stellarmagnet:
![image](https://user-images.githubusercontent.com/2584493/53567740-6e9a0e80-3b60-11e9-807a-8cb86779e6b6.png)

Issue #405, opened 2019-02-08
Decrease line-height on range vote row for Options text


stellarmagnet:
![image.png](https://images.zenhubusercontent.com/5ab2c6f24b5806bc2bcbd667/b0779868-2943-402e-9350-9d896b68794d)

PR #534, opened 2019-03-11
fixed misc styling issues, incl.  #405, #496

stellarmagnet:
fixed misc styling issues, incl.  #405, #496 
tweaked css in a few other places as well


apps/range-voting/app/components/ProgressBar.js
```diff
@@ -26,13 +26,24 @@ const Main = styled.div`
   width: 100%;
   align-items: center;
 `
+
+const Label = styled(Text)`
+  overflow: hidden;
+  display: -webkit-box;
+  -webkit-line-clamp: 2;
+  -webkit-box-orient: vertical;
```
ottodevs: These rules seem to be non-standard:
https://caniuse.com/#search=line-clamp
https://developer.mozilla.org/en-US/docs/Web/CSS/box-orient****
Apart from this, we should not use browser specific rules since babel / styled-components handle that, could we have an alternative style for lines 32-34?

ottodevs:
It seems these browser specific css rules have no effect on the CSS layout

Issue #317, opened 2019-01-22
Use Address Badge component throughout our Aragon Apps


stellarmagnet:
Anywhere where we are displaying an Ethereum address in our apps, we should use the Address Badge component instead.

Please double check all of our existing apps, and ensure this standard is implemented going forward (for Rewards app etc).

- [ ] Allocations
- [ ] Projects
- [ ] Range Voting
- [ ] Address Book

ottodevs:
Solved at #536

Issue #182, opened 2018-10-13
Use blockExplorerUrl() from @aragon/ui for etherscan URL


stellarmagnet:
Our apps do not have the proper logic for the etherscan URL. We should follow standard conventions of other Aragon apps.

Please double check all of our apps, see if it is using etherscan, and update as appropriate
- [ ] Projects app
- [ ] Allocations app
- [ ] Range Voting app
- [ ] Address Book app

see here - https://app.zenhub.com/workspaces/all-aragon-repos-5996d0a88c7c6963f4a4dfdc/issues/aragon/aragon/457

ottodevs:
The etherscanBaseURL is getting pulled from an environment variable in Aragon's Survey App, we should maybe take the same approach: https://github.com/aragon/aragon-apps/blob/ced3a611c394de9f51a38e7f1df580b576940037/apps/survey/app/src/App.js#L19

ottodevs:
Solved at #536

PR #536, opened 2019-03-11
Implement dynamic url for etherscan links and use IdentityBadge from @aragon/ui

ottodevs:
Solves  #182 and #317


Quazia:
Can you pull from dev to resolve the coveralls issue?

PR #579, opened 2019-03-15
452-your-reward-panel

rkzel:
Solves:

- Your Reward side panel [front-end] #452

One small detail: IconTime can't have its color changed, we'll need to add it to our repo or live with it being like this.


apps/rewards/app/components/App/App.js
```diff
@@ -11,17 +11,27 @@ import BigNumber from 'bignumber.js'
 
 const ASSETS_URL = 'aragon-ui-assets/'
 
+const reward = {
+  creator: '0xb4124cEB3451635DAcedd11767f004d8a28c6eE7',
+  isMerit: true,
+  referenceToken: 'SDN',
+  rewardToken: 0x0,
+  amount: BigNumber(17e18),
+  amount: 10,
```
ottodevs: Amount seems duplicated here

Issue #333, opened 2019-01-23
Range vote - display what the logged in user previously voted


stellarmagnet:
We need to display what the logged in user voted in "Voting Results" section of the Range Vote panel. Accomplish this by appending a`YOUR VOTE: xx%` tag directly to the right of the option name.

![image](https://user-images.githubusercontent.com/2584493/52300081-46ffbe00-2987-11e9-86d9-9b2c1af515ce.png)

PR #589, opened 2019-03-16
333 display previous votes

ottodevs:
Solve #333


apps/range-voting/app/components/Panels/VotePanelContent.js
```diff
@@ -129,6 +131,29 @@ class VotePanelContent extends React.Component {
...
+      .toPromise()
+
+    // TODO: Bignumber.js vs >8.2 supports .sum function to initialize from a sum of bignumbers, replace when updating
+    const totalVotesCount = result.reduce(
+      (acc, vote) => acc.plus(vote),
+      new BigNumber(0)
+    )
+    //
+
+    const voteWeights = result.map(e =>
```
Quazia: This is going to represent their 'relative' votes not the actual token amounts they voted with is this the desired behavior?

ottodevs:
Final result:

image

PR #594, opened 2019-03-19
451 New Reward Integration

topocount:
Tasks:
- [x] Set up script.js with the app store
- [x] Integrate vault with arbitrary tokens at the store level
- [x] Hook up New Rewards Panel with the app contract call 
- [x] Set up script.js to catch the new reward
- [ ] Write Cypress tests for the flow of adding a reward


apps/rewards/app/components/App/App.js
```diff
@@ -15,10 +15,20 @@ class App extends React.Component {
...
     app: PropTypes.object.isRequired,
     rewards: PropTypes.arrayOf(PropTypes.object),
+    balances: PropTypes.arrayOf(PropTypes.object),
   }
 
   onNewReward = reward => {
-    console.log('Create New Reward from', reward)
+
+    this.props.app.newReward(
+      reward.rewardType ? false : true, //bool _isMerit,
```
Quazia: Can we adjust the rewardType so these are kept in sync, this seems like it would get confusing quick to have them inverted.

apps/rewards/app/components/Panel/NewReward/NewReward.js
```diff
@@ -36,21 +37,25 @@ class NewReward extends React.Component {
 
   changeField = ({ target: { name, value } }) =>
     this.setState({ [name]: value })
-  
+
   onSubmit = () => {
-    const dataToSend = this.state
-    dataToSend.currency = currencies[this.state.amountCurrency]
+    const dataToSend = { ...this.state }
+    const BLOCK_DURATION = 15
```
Quazia: We should move the time handling to a helper function in a shared folder so we don't have to duplicate it.

apps/rewards/app/store/token.js
```diff
@@ -0,0 +1,198 @@
...
+export async function vaultLoadBalance(state, { returnValues }, settings) {
+  const { token } = returnValues
+  const r = await updateBalances(
+    state,
+    token || settings.ethToken.address,
+    settings
+  )
+  return {
+    ...state,
+    balances: r.newBalances,
```
Quazia: I don't like the naming here "balance" over "token" but I don't think to change it everywhere is worth it/this keeps in sync with Aragons naming style but curious on your opinion.

topocount: I share your views pretty much exactly, Now that it's working I'm leaning towards changing it though

apps/rewards/contracts/RewardsCore.sol
```diff
@@ -119,11 +149,16 @@ contract RewardsCore is IsContract, AragonApp {
...
+        if (supply == 0) {
+            return 0;
+        }
+
         uint initialBalance = reward.referenceToken.balanceOfAt(msg.sender, reward.blockStart);
         uint endingBalance = reward.referenceToken.balanceOfAt(msg.sender, reward.blockStart + reward.duration);
-        require(initialSupply < endingSupply, "The supply must have increased over the period");
-        require(initialBalance < endingBalance, "The user must have earned tokens over the period");
-        supply = endingSupply - initialSupply;
+        //require(initialSupply < endingSupply, "The supply must have increased over the period");
```
Quazia: Both of these requires should be valid, where were they causing issues?

apps/rewards/app/utils/web3-utils.js
```diff
@@ -0,0 +1,41 @@
...
+ * Examples:
+ *   shortenAddress('0x19731977931271')    // 0x1973…1271
+ *   shortenAddress('0x19731977931271', 2) // 0x19…71
+ *   shortenAddress('0x197319')            // 0x197319 (already short enough)
+ *
+ * @param {string} address The address to shorten
+ * @param {number} [charsLength=4] The number of characters to change on both sides of the ellipsis
+ * @returns {string} The shortened address
+ */
+export function shortenAddress(address, charsLength = 4) {
```
Quazia: Are we using this in other apps?

apps/rewards/app/store/init.js
```diff
@@ -0,0 +1,46 @@
...
+      } catch (err) {
+        console.error('[Rewards script] initStore', event, err)
+      }
+      // always return the state even unmodified
+      return state
+    },
+    [
+      // Always initialize the store with our own home-made event
+      of({ event: INITIALIZATION_TRIGGER }),
+      //of({ event: 'test1' }),
```
Quazia: Clean these up

apps/rewards/contracts/RewardsCore.sol
topocount: The reason these `requires` have no place in this function is that this is used as a view function by the frontend for the current user. We'll come back to these later

apps/rewards/app/utils/web3-utils.js
topocount: I think we have our own address-shortening function. This one came with the other aragon functions I was utilizing for the Vault integration

apps/rewards/app/store/init.js
topocount: missed these before merging the PR. Just merged in this change to dev manually

Issue #522, opened 2019-03-09
My Rewards tab: Tables


stellarmagnet:
There will be two tables displayed on the My Rewards tab:

* Unclaimed Rewards
* Claimed Rewards

[Specs](https://docs.google.com/document/d/1Ub332Ad8aRXmJJ7klWoQ2ogHLdYy8fzE5RrKkUnKd6A/edit#heading=h.32kv09uzc713)
[Designs](https://projects.invisionapp.com/d/main#/console/14716432/332905225/preview)
[Responsive Row Design](https://projects.invisionapp.com/d/main#/console/14716432/353428560/preview)

**Acceptance testing**
- [ ] The design should be responsive (see [Finance example](https://github.com/aragon/aragon-apps/pull/619/files#diff-c6b357d815fc23c7aea312476aabe3be))

Issue #437, opened 2019-02-20
Overview tab: Tables


stellarmagnet:
There will be three tables displayed on the Overview tab: 
* Current Rewards
* Pending Rewards
* Past Rewards

[Specs](https://docs.google.com/document/d/1Ub332Ad8aRXmJJ7klWoQ2ogHLdYy8fzE5RrKkUnKd6A/edit#heading=h.5v0q08q0nfd)

[Design](https://projects.invisionapp.com/d/main#/console/14716432/332905229/preview)

[Responsive Row Design](https://projects.invisionapp.com/d/main#/console/14716432/353428560/preview)

**Acceptance testing**
- [ ] The design should be responsive (see [Finance example](https://github.com/aragon/aragon-apps/pull/619/files#diff-c6b357d815fc23c7aea312476aabe3be))

PR #595, opened 2019-03-20
437 rewards overview tables

rkzel:
Solves:

- Overview tab: Tables #437
- My Rewards tab: Tables #522

...and generally Rewards FE. One possibly missing thing is that graph in MyRewards, but it can be easily added later, when FE-BE integration happens and there is real data to work on.


apps/rewards/app/components/Content/MyRewards.js
```diff
@@ -1,27 +1,203 @@
...
+]
+
+const generateOpenDetails = (reward, openDetails) => () => {
+  console.log('calling openDetails on', reward)
+  openDetails(reward)
+}
+
+const translateToken = (token) => {
+  if (token == 0x0) {
+    return 'ETH'
```
topocount: This block shouldn't be needed because the ETH (and all other tokens') symbol is included in the App's state via the vault integration.

apps/rewards/app/components/Panel/ViewReward/ViewReward.js
```diff
@@ -126,16 +129,16 @@ class ViewReward extends React.Component {
             <div>
               <FieldTitle>Period</FieldTitle>
               <div>
-                {Intl.DateTimeFormat().format(startDate)}
+                {format(startDate, 'd-m-yyyy')}
```
topocount: I don't think it's wise to hard-code the format here or on line 134, since it varies by locale and this format will confuse any Americans without more context. 

I utilized the `Intl` object specifically because it is locale-sensitve and will show the date in a way users are expecting to see it.

apps/rewards/app/components/Content/MyRewards.js
```diff
@@ -1,27 +1,203 @@
...
+      <ClickableTableRow key={i} onClick={generateOpenDetails(reward, openDetails)}>
+        <TableCell>
+          <RewardDescription>
+            {reward.description}
+          </RewardDescription>
+        </TableCell>
+        <TableCell>
+          <IdentityBadge
+            networkType={network.type}
+            entity={reward.creator}
```
topocount: There's no creator shown in the spec or the invision wireframes: https://projects.invisionapp.com/d/main#/console/14716432/332905229/preview

I think this column should be removed

apps/rewards/app/components/Content/RewardsTables.js
```diff
@@ -0,0 +1,121 @@
...
+  justify-content: space-between;
+  align-items: center;
+  padding: 20px 10px;
+`
+const AmountBadge = styled(Badge).attrs({
+  background: '#D1D1D1',
+  foreground: theme.textPrimary,
+})`
+  padding: 10px;
+  margin: 20px
```
topocount: missing a semicolon here

rkzel: Added.

apps/rewards/app/components/Content/MyRewards.js
rkzel: That's 'transaction' column. Here it shows an eth address, which happened to be creator's address, but it should be changed in integration phase.

apps/rewards/app/components/Panel/ViewReward/ViewReward.js
rkzel: Good point. I wondered about it before, but went with design. But I guess we can as well start thinking about i18n now. Changed.

apps/rewards/app/components/Content/MyRewards.js
topocount: A tx hash is 32 bytes and is essentially the id for a given transaction. An address is 20 bytes, so I'm pretty sure the Identity badge isn't applicable here.

rkzel: Aw. Of course. TX ID, not address. Fixed!

apps/rewards/app/components/Content/MyRewards.js
rkzel: ok, I added some mock vault data and removed the translation - is it sufficient for now?

PR #610, opened 2019-03-20
update aragon/ui from 0.30.1 to 0.32.0

rkzel:
Because of Viewport added in 0.31.


ottodevs:
Just a suggestion, does it make sense to use the target version on the branch name? this way we will be able to create multiple branches without conflicting names when updating single dependencies.

rkzel:
Added the lock file.

rkzel:
This update sneaked in in Rewards. This PR is now obsolete. Potential problems (that prevented this PR to be merged) we are aware of...

PR #644, opened 2019-03-25
Allocations rework

### LINKED ISSUE #216: [Allocations] Increase efficiency of runPayout require guard
### https://github.com/Giveth/planning-app/blob/b517470cf19f63e15d7447b886c36abd87446f56/apps/allocations/contracts/Allocations.sol#L214

Since we may want to exit soon if the requires don't meet, I think it is a good idea to move lines 214-223 to line 207 (just after payout declaration)

We are always running a `for` loop that lot of times will be unneeded

### LINKED ISSUE #305: Allocations - Execute payout doesn't work when the payouts are being sent to other Allocations Accounts
UPDATE:
I have tested that this works now when sending to regular Ethereum addresses, but it does not work when I am trying to send to Allocations Accounts.

Is this because of gas issues? 4 months ago it worked when sending to allocations accounts, I think we have to figure out what is causing this.

OLD TEXT:
script.js breaks after a completed, voted-on allocation payout is executed

clearing the caches shows the new allocations screen, so that executePayout event is causing a logjam in script.js

### LINKED ISSUE #580: Allocations contract testing
Cover line:53, create a test to trigger the payable function.

### LINKED ISSUE #616: Allow Range Votes to be executed / distributed out of sequence

Quazia:
resolves #580
resolves #616
resolves #305
resolves #216


apps/allocations/app/components/App/Payouts.js
```diff
@@ -0,0 +1,150 @@
...
+            {payout.description}
+          </PayoutDescription>
+          <Text.Block size="small" color={theme.textTertiary} style={{ marginTop: '5px' }}>
+            {showStatus(payout)}
+          </Text.Block>
+        </div>
+        <div style={{ display: 'flex', flexWrap: 'nowrap', alignItems: 'center' }}>
+          <div style={{ marginRight: '10px' }}>
+            <AmountBadge>
+              {displayCurrency(BigNumber(payout.amount))}{' '}{translateToken(payout.rewardToken,tokens)}
```
stellarmagnet: This says "rewardToken" -- i assume since it was reusing code from rewards app. double check this @topocount

apps/allocations/app/components/Panel/NewAllocation.js
```diff
@@ -13,7 +13,6 @@ import {
 } from '../Form'
 
 // TODO: Extract to shared
-const AVAILABLE_TOKENS = [ 'ETH', 'ANT', 'GIV', 'FTL', '🦄' ]
 const ALLOCATION_TYPES = [ 'Informational', 'Token Transfer' ]
```
stellarmagnet: Weren't we supposed to remove informational?

topocount: It's removed from the backend. Can we just remove the dropdown that selects between token transfer and Informational?

apps/allocations/app/components/App/Payouts.js
topocount: This has been changed from `rewardToken` to `token`

Issue #204, opened 2018-10-19
Range Voting: Enhancements to vote Status logic


stellarmagnet:
- [ ] Token transfer vote: If the vote is closed, minimum participation has been met, and it HAS NOT been executed, the status should be "Approved". Display "Execute Vote" button in RV landing page, below status.
- [ ] Token transfer vote: If the vote is closed, minimum participation has been met, and it HAS been executed, the status should be "Approved and executed". Do not display the "Execute Vote" button in RV landing page since it already has been executed.
- [ ] Informational Vote: If the vote is closed, minimum participation has been met. The status should be "Approved (informational)". There should never be the ability to Execute the vote since it's informational
- [ ] Token transfer vote or Informational: If the vote is closed, minimum participation HAS NOT been met, the status should be "Rejected" (tbd - what do you think?)
- [ ] Double check that status is correctly updated in the main panel as well as the sidepanel (see: #379)
- [ ] If there are two range votes that still need to execute their payout, need to inform the user/not allow the 2nd one to be executed until the first one has run (see #339 )

Related to issue #300

PR #649, opened 2019-03-27
[204] Add accurate vote status information

Quazia:
Addresses issue #204


apps/range-voting/app/components/VoteRow.js
```diff
@@ -62,7 +67,18 @@ class VoteRow extends React.Component {
...
+              {open ? <Countdown end={endDate} /> : <VoteStatus vote={vote} />}
+            </div>
+            {!open && getVoteStatus(vote) === VOTE_STATUS_SUCCESSFUL &&
+              <div>
+                <Button mode="strong" wide onClick={this.handleExecuteVote}>
+                  Execute Vote
+                </Button>
+              </div>
+            }
+          </div>
```
ottodevs: It seems the status is not being updated after the vote ends, until I open the sidepanel or refresh the app

apps/range-voting/app/components/VoteStatus.js
```diff
@@ -19,15 +20,20 @@ const ATTRIBUTES = {
...
   [VOTE_STATUS_FAILED]: {
     label: 'Rejected',
     Icon: IconCross,
     color: theme.negative,
   },
 }
 
-const VoteStatus = ({ vote: { data, support, quorum } }) => {
-  const status = getVoteStatus(data, support, quorum)
+const VoteStatus = ({ vote: vote }) => {
```
ottodevs: Shouldn't it work with `const VoteStatus = ({ vote }) => {` ?

Quazia: I think that was giving me issues but feel free to try it.

Issue #654, opened 2019-03-27
Clear Stale GitHub Token from cache


Schwartz10:
1. sign in with github
2. revoke access to github (from github)
3. refresh the projects app 

The app is using a stale github token, so we get a bunch of errors when trying to fetch repos and issues

The app needs to handle a 403 error (or whatever it comes back as) from github and clear the aragon cache's github token

jayalaves:
@Schwartz10 Is this issue related to what just happened to me here? I revoked GitHub, refreshed the app, and it basically became unusable until I cleared the cache.

<img width="1440" alt="Screenshot 2019-11-18 at 11 31 48" src="https://user-images.githubusercontent.com/28843778/69060774-3c934a00-0a18-11ea-90bf-bdd4c91d3957.png">
<img width="1440" alt="Screenshot 2019-11-18 at 11 32 21" src="https://user-images.githubusercontent.com/28843778/69060819-52087400-0a18-11ea-8b57-20390d1032fc.png">

Schwartz10:
@jayalaves it's been a while since I was in this part of the code, but it _smells_ like the same issue. The app should remove the token from cache automatically if it attempts to log in to your github and comes back with a `401` or `403` presumably. What's your network tab say?

e18r:
I was able to reproduce this by following @Schwartz10's instructions.
![Screenshot from 2019-12-10 12-39-00](https://user-images.githubusercontent.com/16065447/70553961-2c1f4c80-1b4a-11ea-9e30-d2a07d5a752a.png)

The Projects app becomes unusable until you clear cache, after which the user is logged out.
![Screenshot from 2019-12-10 12-43-47](https://user-images.githubusercontent.com/16065447/70554278-d7c89c80-1b4a-11ea-8cca-67bc4d0dc3cc.png)

@javieralaves What should happen in this scenario? I suppose simply log the user out immediately? Maybe an error message?

rkzel:
Ok. There are actually several places that need checking - each time a GH query errors out with networkError 401 we need to show a 'log in again' button with appropriate message, like "you have been kicked out of GH, please log ing again". Places:
1. Overview tab, adding new Project,
2. Issues, any interaction within,
3. IssueDetails.

I tried clearing `token`,  but by itself it does cause positive UI interaction - what happens instead is that the error is still displayed. So, I can make the error look like information and button to re-login:

[...]

PR #655, opened 2019-03-27
368 380 GitHub issues

Schwartz10:


Schwartz10:
Hey reviewers! 

A couple things to note:

- occasionally the app triggers a hot reload, which can then cause the app to infinite loop. This shouldn't be an issue in production (lol after all this asking for HMR...)
- This specifically fixes the issues described above, I've verified the other console errors are occurring in `dev` and were not introduced in this PR
- lastly, the `appState` frequently changes and gets mutated, this is concerning because there may be other race conditions that occur on connections or browsers that are different than mine. It is difficult to fix this without some major changes to how application state is handled in `script`.

apps/projects/app/script.js
```diff
@@ -282,6 +306,17 @@ async function syncTokens(state, { token }) {
  *                     *
  ***********************/
 
+const resetAppState = (nextState) => Object.assign({}, nextState)
```
topocount: Not a spread syntax fan?

Schwartz10: Ah! Yes I am a big fan, and will change :)

Schwartz10:
Thanks for the review @topocount ! 

> On another attempt the server didn't respond and I ultimately received a github object with an undefined token, which we haven't set up handling for yet

Ah yes, this is related to: https://github.com/AutarkLabs/planning-suite/issues/654 I can wrap that up.

> we could alternatively use the pre-built loading animation while we wait for a response. I understand that this may or may not fit in under the scope of this PR

imo this is a "nice to have" and requires a significant amount of work for a small amount of gain (over other issues https://github.com/AutarkLabs/planning-suite/issues/657, https://github.com/AutarkLabs/planning-suite/issues/658, https://github.com/AutarkLabs/planning-suite/issues/659 )

Several parts of the app can "trigger" a github auth, and we would have to write state variables that inform the front-end of a loading state or not. Definitely doable but has lift. 

My vote is to punt on the loading, handle errors

Schwartz10:
@Quazia nothing changed, i just needed to re-request your review after dismissing @topocount 's - can one of you two approve so i can merge this in please?

apps/projects/app/components/App/App.js
```diff
@@ -102,16 +102,9 @@ const getURLParam = param => {
...
-    let json = await response.json()
-    if (json.token) return json.token
-    else throw Error(`${json.error}`)
-  } catch (e) {
-    console.error('Error from Authentication server:', e)
-  }
+  const response = await fetch(`${AUTH_URI}/${code}`)
+  const json = await response.json()
+  return json.token
 }
```
ottodevs: Not a big deal since we probably will refactor this part to make use of hooks and reusability, just wondering if we should still keep try-catching eventual fetch errors, since it is a common scenario on poor connectivity conditions

PR #662, opened 2019-03-28
653 ipfs propagation

Schwartz10:


apps/projects/app/components/App/App.js
```diff
@@ -28,7 +29,7 @@ let CLIENT_ID = ''
 let REDIRECT_URI = ''
 let AUTH_URI = ''
 
-let ipfs = ipfsClient({ host: 'localhost', port: '5001', protocol: 'http' })
+let infuraIpfs = ipfsClient({ host: 'ipfs.infura.io', port: '5001', protocol: 'https' })
```
Quazia: Long term we should migrate to using the new aragon.js ipfs utilities but this approach is fine for now.

apps/projects/package.json
```diff
@@ -29,6 +29,7 @@
     "@aragon/ui": "0.30.1",
     "@tps/ui": "^0.0.1",
     "apollo-boost": "^0.1.22",
+    "axios": "^0.18.0",
```
Quazia: I don't love importing this _just_ for the one line but if I'm the only person that feels that way it's not a big deal.

apps/projects/app/components/App/App.js
stellarmagnet: aragon.js has utilities that pins to ipfs?

apps/projects/package.json
Schwartz10: The problem i was having with using native browser `fetch` is that it can send a preflight request to a backend to establish a connection (or for various other reasons i'm not aware of). The prefligth request is an `OPTIONS` request, which gives back a CORS error from infura 

Maybe our recent upgrades will fix this tho and then we can take it out and potentially not even write our own getter method

apps/projects/app/components/App/App.js
Schwartz10: Yeah long term there are a lot of cool things we can do, even use the dweb loader: https://github.com/DistributedWeb/dweb (probably after filecoin gets established). For now i think this is fine

PR #664, opened 2019-03-28
initial upgrade for @aragon/api

Schwartz10:


apps/allocations/app/store/entry.js
```diff
@@ -40,15 +42,15 @@ const loadEntryData = async (addr, addressBook) => {
 
   return addressBookApp
     .getEntry(addr)
-    .first()
-    .map(
+    .pipe(first())
+    .pipe(map(
```
Quazia: Could move this into the previous pipe, if the map seems too large for readability could be a custom operator. I think this mapping functionality is re-used. Neither is critical though.

apps/range-voting/app/script.js
```diff
@@ -117,98 +118,93 @@ async function loadVoteDescription(vote) {
...
+        const returnObject = {
+          ...marshallVote(voteDescription),
+          metadata,
+          canExecute,
+          options,
+        }
+
+        allocations.getPayout(voteDescription.externalId)
+          .pipe(first())
+          .subscribe(payout => (resolve({
```
Quazia: This subscription will potentially have some grim merge conflicts, no changes needed here but something to watch for.

apps/range-voting/app/script.js
```diff
@@ -117,98 +118,93 @@ async function loadVoteDescription(vote) {
...
-      .subscribe(voteData => {
+      .pipe(first())
+      .subscribe(async voteData => {
         let funcSig = voteData.executionScript.slice(58, 66)
         if (funcSig == 'f2122136') {
           console.log('Loading Projects Data')
           resolve(loadVoteDataProjects(voteData, voteId))
         } else {
           console.log('Loading Allocations Data')
+          console.log('vote data: ', await loadVoteDataAllocation(voteData, voteId))
```
Quazia: I don't think that pushing this anonymous function to become async is worth being able to log it here. If anything if you want this info I'd log it out in like handleEvents or updateState after the updateVotes call?

apps/range-voting/app/script.js
```diff
@@ -117,98 +118,93 @@ async function loadVoteDescription(vote) {
...
-          }
-          resolve(returnObject)
-          // Project specific code
+        const voteDescription = await loadVoteDescription(vote)
+        let options = []
+        for (let i = 0; i < totalCandidates; i++) {
+          let candidateData = await getProjectCandidate(voteId, i)
+          options.push(candidateData)
+        }
+        resolve({
```
Quazia: Same as above, look here when merging 310.

apps/range-voting/app/script.js
```diff
@@ -273,17 +269,19 @@ function loadVoteSettings() {
...
             .call(name)
-            .first()
-            .map(val => {
-              if (type === 'number') {
-                return parseInt(val, 10)
-              }
-              if (type === 'time') {
+            .pipe(
+              first(),
+              map(val => {
```
Quazia: Seems re-usable could be a custom operator. Again doesn't need to be addressed now.

apps/allocations/app/store/account.js
```diff
@@ -54,8 +59,7 @@ const loadAccountData = async accountId => {
     // TODO: Should we standarize the naming and switch to getAccount instead of getPayout?
     app
       .call('getPayout', accountId)
-      .first()
-      .map()
+      .pipe(first(), map())
```
Quazia: I went to check if this empty map could be removed and as far as I can tell this entire function is actually unused now.

apps/range-voting/app/script.js
Schwartz10: agreed, would love to take out all of these logs. I kept it in there and added the `await` keyword because there was a log in there that called `loadVoteDataAllocation` but it didn't await, so the log was useless. I'll take it out all together

apps/allocations/app/store/entry.js
Schwartz10: Will change

apps/range-voting/app/script.js
Schwartz10: Good idea

apps/allocations/app/store/account.js
Schwartz10: Cool will take out

Issue #639, opened 2019-03-25
Applied Filter Tiles and filter bar reset functionality


stellarmagnet:
In the Projects app, Issues tab, we need to implement **Applied Filter Tiles**. 

Directly to the right of the Keyword Search box, there will be "applied filter tiles" for each individual checkbox selected from the Filter menus, or from typing into the keyword search. A "Clear Filters" link will also be displayed as soon as there is one tile, and this resets to the table to the default query state (except it doesn't reset the Sort option).

**Design**
https://invis.io/JGR85SZ5BS4

PR #683, opened 2019-03-30
639 filter tiles

Schwartz10:
Add support for https://github.com/AutarkLabs/planning-suite/issues/639


Schwartz10:
@Quazia this could use a second pair of eyes - I made some significant changes to the way the filter state is managed. Previously it was managed in two different places, with the beef of the filter information kept in a local state that was unreachable from a sibling component.

I moved all state management up to a single parent. The `sortBy` functionality could use this too to avoid future bugs. But wanted to check this in to get your opinion on if this was good enough and keep moving to fix more stuff before the deadline

apps/projects/app/components/Content/FilterTile.js
```diff
@@ -0,0 +1,41 @@
...
+          }}>
+            {this.props.text}
+            <div
+              style={{
+                marginLeft: '10px',
+                paddingRight: '4px',
+                cursor: 'pointer'
+              }}
+              role="button" aria-pressed="false"
+              onClick={this.props.disableFilter}
```
Schwartz10: I usually get some linter settings here that give more instructions. I'm not sure if i missed some cross-site script blocking tags

apps/projects/app/components/Content/Filters.js
```diff
@@ -0,0 +1,96 @@
...
+    {
+      filter1Name: [path, to, filter1, on, parent],
+      filter2Name: [path, to, filter2, on, parent],
+    }
+
+    to make it easier to deselect filters from this view without multiple state objects
+  */
+  calculateFilters = () => {
+    const { issues, bountyIssues } = this.props
+    const filterInformation = prepareFilters(issues, bountyIssues)
```
Schwartz10: This `prepareFilters` functionality I stripped out of the `FilterBar` component and reused it here - it provides the information about filters that the UI needs to display stuff. This is working really well, but I would still love to abstract all of this logic out and store it all in one state variable on the parent `Issues.js` or use the render props pattern. Feeling like that might be overkill for the moment. Happy to do more abstraction and cleanup around this tho. Just lmk

apps/projects/app/components/Content/Issues.js
```diff
@@ -99,7 +101,7 @@ class Issues extends React.PureComponent {
 
   handleFiltering = filters => {
     // TODO: why is reload necessary?
```
Schwartz10: I didn't spend too long trying to get this to work without the `reload` state variable, but this is generally caused by the fact that state was stored in two different locations, and state updates on the parent not being propogated to the child that actually needs to get rerendered on a state change (due to a changed filter)

apps/projects/app/components/Content/Issues.js
```diff
@@ -210,9 +208,26 @@ class Issues extends React.PureComponent {
     this.setState({ showIssueDetail: false, currentIssue: null })
   }
 
-  actionsMenu = () => (
-    <div>
+  disableFilter = (pathToFilter) => {
+    let newFilters = { ...this.state.filters }
+    recursiveDeletePathFromObject(pathToFilter, newFilters)
```
Schwartz10: This function may seem a little confusing, but really all it's doing is taking a path from an `alias` (in the `ActiveFilters.js` component, and deleting that filter from state (removing the filter).

SO object:
```
{
path: { to: { object: 'value', key: 'value2' }}
}
```
with path: `['path', 'to', 'object']`

becomes:
{
path: { to: { key: 'value2' }}
}
```

thus deleting the filter

File: apps/projects/app/components/Shared/FilterBar/FilterBar.js [hunk 5]
```diff
@@ -47,14 +47,6 @@ const ActionLabel = styled.span`
 
 class FilterBar extends React.Component {
   state = {
-    filters: {
```
Schwartz10: Think this is a big cleanup win :)

apps/projects/app/components/Shared/FilterBar/prepareFilters.js
```diff
@@ -0,0 +1,86 @@
+import { BOUNTY_STATUS } from '../../../utils/bounty-status'
+
+/*
+  prepareFilters builds data structure for displaying filterbar dropdowns
+  data comes from complete issues array, issuesFiltered is used for counters
+*/
+const prepareFilters = (issues, bountyIssues) => {
```
Schwartz10: FYI no changes were made to this function - it's identical to the one stripped out of the `FilterBar` component

Schwartz10:
Closed for #702

apps/projects/app/components/Content/Filters.js
```diff
@@ -0,0 +1,115 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import { Text } from '@aragon/ui'
```
stellarmagnet: import 'theme' as well so we can use theme colors

apps/projects/app/components/Content/Filters.js
```diff
@@ -0,0 +1,115 @@
...
+            role='button'
+            style={{
+              cursor: 'pointer',
+              marginLeft: '18px'
+            }}
+          >
+            <Text
+              size='large'
+              weight='bold'
+              color='#80abe3'
```
stellarmagnet: color={theme.accent}

i think

apps/projects/app/components/Content/Filters.js
```diff
@@ -0,0 +1,115 @@
...
+          <div
+            onClick={this.props.disableAllFilters}
+            role='button'
+            style={{
+              cursor: 'pointer',
+              marginLeft: '18px'
+            }}
+          >
+            <Text
+              size='large'
```
stellarmagnet: suggest we use 'normal' instead

Issue #613, opened 2019-03-21
Range Voting bar padding issues


stellarmagnet:
Only occurs on landing page: 
![image](https://user-images.githubusercontent.com/2584493/54744354-1090a200-4bc7-11e9-9682-a1c019c0cfe9.png)

Issue #557, opened 2019-03-13
Reduce border height


stellarmagnet:
![image](https://user-images.githubusercontent.com/2584493/54296341-05c98200-45b5-11e9-809d-38a3e805517f.png)

Issue #553, opened 2019-03-13
Add padding at top of input field


stellarmagnet:
This below is from New Issue via projects app.

We need to add padding at top of input field, so the text isn't so close to the border

This seems to be a recurring problem with these larger inputs, let's check it globally

![image](https://user-images.githubusercontent.com/2584493/54286032-5b952e80-45a3-11e9-86bf-55c3ab0f53f7.png)

Issue #545, opened 2019-03-13
Radspec updates part (i)


stellarmagnet:
**Create account**
Current: Create a new account used for creating allocations
New: Create allocation account ‘_account_name_’
Example: Create allocation account ’2019 Nest Budget'

**Submit allocation**
Current: Sets the distribution proportion for this allocation
New: Create _amt_ allocation range vote for ‘_allocation_name_’
Example: Create 8 ETH allocation range vote for ‘Q1 Nest Grants’

Remove " ' " from Address Book radspec around addresses

Add something for changeBountySettings

![image](https://user-images.githubusercontent.com/2584493/54284340-0efc2400-45a0-11e9-9981-930d039d31de.png)

PR #694, opened 2019-03-31
ui tweaks incl #553, #545, #557, #613

stellarmagnet:
ui tweaks incl #553, #545, #557, #613

I also:
- enhanced the style of the badge on the Range Vote panel
- enhanced the style of the "show more" options link on Range Vote vote row


apps/range-voting/app/components/VoteRow.js
```diff
@@ -101,13 +101,21 @@ class VoteRow extends React.Component {
...
+                  pointerEvents: 'auto'
+                  }}                
                 onClick={() => this.setState({ showMore: !showMore })}
               >
                 {showMore
-                  ? 'Show less options...'
-                  : options.length - 2 + ' more options...'}
-              </ShowMoreText>
+                  ? 'Show less...'
+                  : ' + ' + remOptions + ' more'}
```
stellarmagnet: I couldn't get this to work without creating a const, not sure if there's a better way

Issue #241, opened 2018-10-27
New Allocation: If I populate an option field and do not click the "+" icon then it should still recognize it as a valid option


stellarmagnet:
Right now when I create a new allocation and I enter a value into an option field, and I don't click the "+" icon, it is not recognizing it as an option when you review the Allocation in the Range Voting app.

We need to recognize all options that are in the fields.

**Designs**
Single option: https://invis.io/7ER87SZC4GB
Multiple options: https://invis.io/9KR87UGTEBG

* If there is only a single option, do not display a minus icon next to the row for option. If there is more than one option, a minus icon will be displayed next to each row.
* Include an "+ Add Another" button.
* All options displayed in the UI will be considered required
* This design pattern applies for both if the Address Book is enabled or if it is not

Quazia:
This should be resolved in #244

stellarmagnet:
This issue returned

stellarmagnet:
As discussed, the expected design should be as follows for both Manual Entry and Address Book, where every exposed field should be considered required:
![image](https://user-images.githubusercontent.com/2584493/54621064-71a16400-4a67-11e9-9576-01e3eb831975.png)

The button should say "Add Another"

PR #699, opened 2019-03-31
241 revisit adding/removing allocations options

ottodevs:
Solves  #241


Quazia:
If I populate an option field and do not click the "+" icon then it should still recognize it as a valid option.

PR #701, opened 2019-04-01
 #688 bounties disappearing after clearing cache

### LINKED ISSUE #688: Bounty allocations are removed from issues after clearing cache
1) select some issues
2) allocate bounties to the issues
3) clear application cache

Result: the issues no longer have bounties

Expected: the issues should retain their bounties after cache is cleared

Schwartz10:
@Quazia I'm not sure if this PR solves every bounty misrepresented-status issue, but it should definitely help. I haven't been able to see the wrong bounty state after these changes, but i also wasn't able to replicate consistently beforehand. This includes clearing the cache and reloading issues from scratch too.

### DIFF
--- a/apps/projects/app/script.js
+++ b/apps/projects/app/script.js
@@ -53,8 +53,18 @@ const repoData = id => `{
     }
 }`
 
+const determineStateVars = (state) => {
+  const repos = (!!state && state.repos) || []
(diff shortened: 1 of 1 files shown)


Schwartz10:
Included in #702

PR #709, opened 2019-04-02
address book FE tweaks

stellarmagnet:
Enhanced the address book entries


apps/address-book/app/components/App/Entities.js
```diff
@@ -8,6 +8,7 @@ import {
   TableHeader,
   TableRow,
   Text,
+  theme,
```
stellarmagnet: i was gonna use some colors but i didnt.... not sure if we keep this in

PR #713, opened 2019-04-04
add length checks for execution script

### LINKED ISSUE #600: Look into potential double voting in RangeVoting
May be better to have another _resetVote() function; the current function is double counting the new values when there was a prior vote.

This issue should include tests for the contract to prevent this in the future.

### LINKED ISSUE #601: Add more bytes length checking in Range Voting
"Should probably do some length checks on the byte arrays"

These checks should also be tested.

topocount:
Resolve #600 : added message to tests to indicate where doublevoting is checked against
Resolve #601 : added a couple requires. might need tests for them


topocount:
E2E Tests are passing

PR #723, opened 2019-04-07
enhanced styles for allocations app

stellarmagnet:
I updated the styles and made the page responsive.

### DIFF
--- a/apps/allocations/app/components/Card/Account.js
+++ b/apps/allocations/app/components/Card/Account.js
@@ -71,7 +71,8 @@ const Account = ({
             {' ' + BigNumber(balance)
               .div(ETH_DECIMALS)
               .dp(3)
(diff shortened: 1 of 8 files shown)


apps/allocations/app/components/Card/Account.js
```diff
@@ -71,8 +71,7 @@ const Account = ({
             {' ' + BigNumber(balance)
               .div(ETH_DECIMALS)
               .dp(3)
-              .toString()}{' '}
-            {translatedToken}
```
stellarmagnet: Basically this is not working. "token" was removed from NewAccount, hence there is no token to translate.
I'm not sure if this is the preferred behavior for this page, but if translateToken and token isn't needed for Account.js then it should probably be removed throughout this file. @Quazia

PR #757, opened 2019-04-10
filtering by all statuses

rkzel:
FilterBar has ability to filer by bounty status. This PR is for showing all statuses plus ability to filter all funded / unfunded issues regardless of the stage they are at. This is how it should look like (HR was updated after taking this screenshot):

statuses


apps/projects/app/components/Content/Issues.js
```diff
@@ -159,14 +159,16 @@ class Issues extends React.PureComponent {
       // if there are no Status filters, all issues pass
       if (Object.keys(filters.statuses).length === 0) return true
       // should bountyless issues pass?
-
       const status = bountyIssueObj[issue.number] ? bountyIssueObj[issue.number] : 'not-funded'
       if ('not-funded' in filters.statuses && !bountyIssueObj[issue.number])
         return true
       // if issues without a status should not pass, they are rejected below
       if (status === 'not-funded') return false
```
Quazia: This line isn't needed, it'll hit the return false regardless of this line being here.

apps/projects/app/components/Content/Issues.js
```diff
@@ -159,14 +159,16 @@ class Issues extends React.PureComponent {
       // if there are no Status filters, all issues pass
       if (Object.keys(filters.statuses).length === 0) return true
       // should bountyless issues pass?
-
       const status = bountyIssueObj[issue.number] ? bountyIssueObj[issue.number] : 'not-funded'
       if ('not-funded' in filters.statuses && !bountyIssueObj[issue.number])
```
Quazia: This line is made redundant by `status in filters.statuses` as `if status in not-funded` `bountyIssueObj[issue.number]` is _necesarily_ false

Issue #729, opened 2019-04-07
Performance issues for Projects app (Issues tab)


stellarmagnet:
There are performance issues for the Projects app, Issues tab. It is most noticeable when trying to click the checkboxes when you have the planning suite repo integrated (clicking a checkbox takes about 5 seconds for me). For smaller repos, it still takes about 1.5 seconds. I think even that is not acceptable - in comparison, clicking a checkbox in Gmail takes 1.5 milliseconds (although they do not use a real html checkbox/button component).

The browser inspector tools say the issue is related to the javascript handlers.

![image](https://user-images.githubusercontent.com/2584493/55690209-8e59f900-598e-11e9-9a9f-2770ed82dc9b.png)

rkzel:
Quick explanation for the checkbox: its handler changes state of its parent component, list of Issues itself. When selecting an issue triggers re-rendering of the whole list, and it might involve reaching out to github to get a fresh list (via GraphQL query). Two things need to happen:
1. GraphQL needs to be moved to backend,
2. Unnecessary renders should be identified and disabled, with, most probably, memoization.

rkzel:
What got done:
1. Checkboxes are now slow only because they are animated, and their speed is not affected by the number of issues,
2. Re-rendering of Issue Cards has been limited,

What still should be done:
1. Moving query downloading issues out of render(). It is complicated and is not expected to improve performance much immediately, but it would open possibilities of preparing data before render() - filterbar objects for example, etc.
2. Issue Card needs to be reviewed  for potential minor improvements (which can become major wins for big repos with hundreds of Cards displayed),
3. https://blog.logrocket.com/rendering-large-lists-with-react-virtualized-82741907a6b3
4. ...

Issue #343, opened 2019-01-26
Ability to "show more" Project issues


stellarmagnet:
[Navigate to the Projects app, and click on the Issues tab. Make sure a project/repo with sufficient issues has been added.]

Right now we only display the last 100 issues of a connected repo

- We should enhance the Issues tab to include a "Show More" button that is displayed after 50 issues. 
- Clicking "Show More" will display the next 50 issues and so on. 
- The default sort should be descending order based on issue # (creation date).

![image](https://user-images.githubusercontent.com/2584493/52267396-358fc500-2939-11e9-8b09-423956466298.png)

PR #765, opened 2019-04-12
343 showmore issues

rkzel:
Addresses:
#343 - Show More Projects, using cursor and downloading batch at a time, adding it to already downloaded issues,
#729 - Performance of Issues. This is mostly about stopping constant re-rendering of individual Cards.


stellarmagnet:
To do: The current scrollY should be saved when "show more" is clicked so the scroll position remains the same after data is loaded, e.g.  scrollTo(savedScroll)

ottodevs:
Overall seems good but I'd rather if your solve merging conflicts and @stellarmagnet TODO note before going deeper and checking out locally 👍

rkzel:
Conflicts solved, and scrolling problem mitigated by setting default sorting by to creation date (it was sorting alphabetically by title before) - now it behaves as requested.

PR #768, opened 2019-04-14
Disable collaborators until we implement workaround the issue

ottodevs:
Addresses #766
Disabled the contributors field for projects until we solve how to access that data without push permission on the repo.

Seems related:
https://github.com/octokit/octokit.net/issues/1070

A workaround could be trying to get this data anonymously since it is public info


ottodevs:
I close this since https://github.com/AutarkLabs/planning-suite/pull/770 contains this single commit

Issue #527, opened 2019-03-09
Update Allocations app front-end to be responsive


stellarmagnet:
- [ ] Add hamburger menu
- [ ] Use + icon for "New Account" when the width decreases (see other apps)

Issue #680, opened 2019-03-30
Improve responsive behavior for filter bar


stellarmagnet:
We need to improve responsive behavior for filter bar in the Projects app. It should look more like this, and it should work well on mobile (as far as opening the menus and clicking the checkboxes). 
![image](https://user-images.githubusercontent.com/2584493/55762451-5670b580-5a63-11e9-83ee-ec031f28957f.png)

Basically we need to: 
* Improve/add icons (see below)
* Enhance the overflow/responsive behavior
* improve the behavior when one of the menus has a lot of options

Add a sort icon: 
https://www.dropbox.com/s/gkvyzk48k03syeq/sort_by_status.svg?dl=0

Improve the ellipsis icon: https://www.dropbox.com/s/g2qrhfoxasgz8os/dots.svg?dl=0

There should be a better arrow icon for the dropdowns: https://www.dropbox.com/s/qpbyda224m14dbe/arrowdown.svg?dl=0

PR #782, opened 2019-04-17
all-apps FE responsiveness review and refactoring

rkzel:
This addresses #527, but all apps were consider and subsequently refactored:
1. all apps now use AppView as their main container,
2. all know how to react to Wrapper Menu hiding,
3. all have top-nav buttons turning into + when screen size goes down,
4. in Projects split between App and AppContent is no more),
5. AppBar is used much better, as Aragon intended, consistently across whole TPS.

What remains: checking responsiveness of all content and testing.


rkzel:
Additionally:
- all tables and cards should dynamically adjust padding for small screens,
- panels are generally expected to stay correct even on "small phone screen" of 320px. Correct does not necessarily mean pretty, more like "readable". There are things like 30px padding in panels set in SidePanel component, preventing total control over panel space,
- FilterBar has separate ticket #680 for its menus, which on small screens span beyond screen borders (it wasn't the main problem, but it will be important but of that ticket),

rkzel:
Padding in Settings sorted out, rewards are in a better shape (no more errors).

Issue #787, opened 2019-04-18
TPS Deploy Todos


stellarmagnet:
- [x] 1.  **Range Voting**: If an account doesn’t have the Range Voting apps tokens, hide the voting sliders and submit button (instead, display the voting results - similar to the “closed vote” state). Below the results, in a blue <Info> component box say:

  * `This account cannot cast a vote because it did not hold any{' '} {tokenSymbol} at the time this vote was created. `

  * [Note: This info component is only displayed when the vote is open and the user didn’t have tokens at the snapshot.]

- [x] 2. **Range Voting**: Token balances should be based on snapshot value. Update the variables to reflect this.
    * Update the text to:
        * Your vote will be weighted by ## which was your {tokenSymbol} balance when the vote was created. No tokens will be spent to vote.
- [x] ~~3. **Range Voting**: Conditionally disable the Submit Vote button until “0 remaining”. Once 0 remaining, enable submit.~~
- [x] 4. **Allocations**: Update string for tokenTransferWarning (Vault instead of Finance app):
    * Since you are proposing an allocation with tokens, it will be withdrawn from the Vault if approved, since Accounts do not hold tokens.
    - [x] (Can we also display the vault balance of the selected token?) "Vault Balance: 4.55"
- [x] 5. **Projects**: Update Projects.sol so bountyCurrency is an address
- [x] 6a. **Kit**: Update PlanningSuite.sol so the init param for Projects app is ‘token’. 
- [x] 6b. **Kit**: Include Rewards in kit.
- [ ] 7. **New Allocation**: Unable to remove Address Book option (when there are multiple) [Fixed by Otto?]
- [x] 8. **New Allocation**: After hitting “new option” the default is the previously selected option. It should be “Select an entry” [Fixed by Otto?]
- [x] 9. **Range Voting panel**: Fix formatting of candidate support (see img). Include %
![image](https://user-images.githubusercontent.com/2584493/56378692-143b3700-620e-11e9-855f-406dd324ec99.png)

## Other deploy to-dos
- [x] Reduce the voting duration (we need to launch a test DAO with this very short, like 2 minutes. For the official one we launch, having 24 hours can work)
- [x] Rename the token symbol (should we just call it ‘spice’)
- [x] App descriptions (@stellarmagnet )
- [x] App screenshots (@stellarmagnet )

Kits
- [x] 1) the initialization param for projects will need to be changed to be "token" instead of "tokenSymbol"
[...]

PR #790, opened 2019-04-19
Hotfix dev errors

Quazia:
Arthur's changes
- [x] Addresses a lot of the small changes indicated in issue #787
- [x] Whenever I select multiple issues, the "Actions” button never becomes enabled. [From #805]
- [x] Null RepoID is stored after curating issues, which results in projects app not loading #800
- [x] Ensure default token is added in projects [#801]

Yalda's changes
- [x] Remove "Finance app" from our copy related to Projects app #802
- [x] Resolve Github URI issues so testing locally works
- [x] Renamed Range Voting to Dot Voting
- [x] Imported AverageRewardsTable in MyRewards


apps/projects/app/components/Content/Settings.js
```diff
@@ -95,13 +88,12 @@ class Settings extends React.Component {
...
     const expLevelsDesc = expLevels.map(l => fromUtf8(l.name))
     let expLevelsMul = expLevels.map(l => web3.toHex(l.mul))
-
     this.props.app.changeBountySettings(
       expLevelsMul,
       expLevelsDesc,
       web3.toHex(baseRate),
       web3.toHex(bountyDeadline),
-      bountyCurrencies[bountyCurrency],
+      this.props.tokens[bountyCurrency].addr,
```
stellarmagnet: This seems to be breaking updating the settings

PR #818, opened 2019-04-23
791 810 848 New Rewards Enhancements and Fixes

### LINKED ISSUE #791: Create reward with any snapshottable Token
Users should be able to use any snapshotted token as a reference token in Rewards, regardless of whether or not it is in the vault. We need to implement a text field that:
* accepts a token address
* verifies that what is entered is an address
* verifies that the address entered is a snapshottable token by test-calling `getBalanceAt` on the token and ensure it doesn't return an undefined value

### LINKED ISSUE #810: Various Rewards cleanup tasks
## Reward panels
- [x] Use token symbols instead of names, in both the dropdowns and summary text
- [x] Ensure that the dividend summary text is in fact the reference asset (incorrect here)

### Add data validation
- [x] The selected start date occurs before your reference asset `ANT` existed. Please choose another date.
- [x] You have specified a date range that results in `#` disbursements, yet our system can only handle `#` (sorry, not sorry). Choose an end date no later than `date`.
- [x] You have specified a reward for `500 ANT`, yet your vault balance is `100 ANT`. To ensure successful execution, specify another amount that does not exceed your balance.

### LINKED ISSUE #848: Reward Summary text seems wrong
The reward text appears to describe a "merit" reward despite being a "dividend"

The time also appears to be incorrect (should be the default 7 days after period ends). 

image

topocount:
Resolve #791 
Resolve #810 
Resolve #848


apps/rewards/app/utils/token-utils.js
```diff
@@ -73,3 +74,11 @@ export async function getTokenName(app, address) {
 
   return tokenName ? toUtf8(tokenName) : null
 }
+
+export async function getTokenStartBlock(app, address) {
```
Quazia: I think you can just import this here https://github.com/AutarkLabs/planning-suite/blob/dev/apps/rewards/app/store/app.js

apps/rewards/app/components/Panel/NewReward/NewReward.js
```diff
@@ -48,56 +84,194 @@ class NewReward extends React.Component {
...
+
+  handleCustomTokenChange = event => {
+    const { value } = event.target
+    const { network } = this.props
+
+    // Use the verified token address if provided a symbol and it matches
+    // The symbols in the verified map are all capitalized
+    const resolvedAddress =
+      !isAddress(value) && network.type === 'main'
+        ? ETHER_TOKEN_VERIFIED_BY_SYMBOL.get(value.toUpperCase()) || ''
```
Quazia: I know it's sort of an edge case but there are duplicates in this list and there isn't really any indication in the UI that this:
1. Only works on mainnet
2. Is pulling symbols off of the verified token list
I think maybe just some sort of warning messages around this functionality would be good.

apps/rewards/app/utils/token-utils.js
topocount: utils are normally imported straight from their file so it's easier to trace them.

apps/rewards/app/components/Panel/NewReward/NewReward.js
topocount: I've updated the UX so it only prompts for an address on any network that isn't mainnet. I've also made the prompts contextually adaptive

PR #827, opened 2019-04-27
new addresses for oauth services

rkzel:
Before merging we must coordinate reconfiguring GitHub and redeploying Projects App.


rkzel:
Merging. We are migrating now.

PR #837, opened 2019-04-29
581 dot voting contract tests

topocount:
This branch will contain all the most recent contract test updates.

# Dot Voting
- [x] Test line: 181, create a test where a user can't vote
- [x] Test line: 213, create a test where a candidate has already been added (duplicate candidate)
- [x] ~~Test line: 254, Test get candidate description~~ remove getCandidateDescription from contract
- [x] Test line: 268, Test isForwarder function
- [x] Test line: 290, Test canForward fail case, where the forwarder does not have perms for given scipt (will require an execution target with invalid permissions)
- [x] Test line: 444, Test case where a different type of execution script is passed in (any exec script with a different spec ID)
- [x] Test line:541 This is a tough one to test, you'll need to create an execution script and manually manipulate its length in the bytes
- [x] Test Lines: 605+ 615 Test cases where user votes with more tokens than they should. One of these requires may be unnecessary.


apps/dot-voting/test/dot-voting.test.js
```diff
@@ -19,9 +19,12 @@ const pct16 = x =>
 const createdVoteId = receipt =>
   receipt.logs.filter(x => x.event === 'StartVote')[0].args.voteId
 
-const castedVoteId = receipt =>
+const getCastedVoteId = receipt =>
```
Quazia: If you want to change receipt/event helpers to a getXXXX format can we change all of them?

apps/dot-voting/test/dot-voting.test.js
```diff
@@ -45,9 +48,9 @@ contract('DotVoting App', accounts => {
       aclBase.address,
       regFact.address
     )
-  })
+    //})
 
-  beforeEach(async () => {
+    //beforeEach(async () => {
```
Quazia: Just remove these

apps/dot-voting/test/dot-voting.test.js
```diff
@@ -342,7 +397,71 @@ contract('DotVoting App', accounts => {
...
+          false
+        )
+      }
+      const script = encodeCallScript([action])
+      return assertRevert(async () => {
+        await app.forward(script, { from: nonHolder })
+      })
+    })
+
+    it('rejects unequal script length and actual script length', async () => {
```
Quazia: Don't love the test string here, maybe "Rejects if supplied script length and actual script length are not equal"?

apps/dot-voting/test/dot-voting.test.js
```diff
@@ -19,9 +19,12 @@ const pct16 = x =>
 const createdVoteId = receipt =>
   receipt.logs.filter(x => x.event === 'StartVote')[0].args.voteId
 
-const castedVoteId = receipt =>
+const getCastedVoteId = receipt =>
   receipt.logs.filter(x => x.event === 'CastVote')[0].args.voteId
 
+const getVoteId = voteId =>
```
Quazia: I'd say this is more like a convert/parse than a get?

apps/dot-voting/test/dot-voting.test.js
```diff
@@ -522,21 +656,28 @@ contract('DotVoting App', accounts => {
 
         assert.equal(
           appleInfo[1].toNumber(),
-          voteTwo[0],
+          7,
```
Quazia: These values don't seem right.

apps/dot-voting/test/dot-voting.test.js
```diff
@@ -558,13 +699,14 @@ contract('DotVoting App', accounts => {
           holderVoteData1[2].toNumber(),
           'vote and voter state should match after casting ballot'
         )
+        await token.transfer(voter, 31, { from: nonHolder })
```
Quazia: Need an asset after the transfer.

apps/dot-voting/test/dot-voting.test.js
```diff
@@ -634,6 +786,8 @@ contract('DotVoting App', accounts => {
         assert.equal(canExecute, false, 'canExecute should be false')
       })
       it('holder can add candidates', async () => {
+        let newvote = await app.newVote(script, 'metadata', { from: holder50 })
+        voteId = createdVoteId(newvote)
         mango = accounts[5]
         await app.addCandidate(voteId, '0xbeefdead', mango, 0x1, 0x1)
         candidates.push(mango)
```
Quazia: I don't think this really needs to be pushed/popped here. We can just use the local object.

Issue #769, opened 2019-04-14
Issue detail bar needs to be enhanced


stellarmagnet:
We should use the standard app bar component for the Issue Detail bar on projects page (see Survey app example)

Right now the current implementation results in the arrow being in the wrong location, and the bottom line being drawn on the bottom of the page

PR #902, opened 2019-05-09
769 enhance issue detail bar

ottodevs:
Solves #769

And some minor issues, like cleaning missed console log at projects, and fixed that annoying chrome dev error with React Number Format at settings tab, related to: https://github.com/s-yadav/react-number-format/issues/272


ottodevs:
> Everything looks good, only thing to check is Details.js was modified upstream and you combined it into the IssueDetail.js file so can you confirm that there weren't any important changes in terms of responsive styling lost in deleting that file?

Just to clarify and for future reference, the IssueDetail was previously wrapping Details component as you guessed, and also the custom NavigationBar, but this one was removed after the refactor, so IssueDetail became just Details, then it made sense to just combine them into a single simple component

apps/projects/app/components/Content/IssueDetail/IssueDetail.js
```diff
@@ -1,49 +1,529 @@
...
-  componentWillUnmount() {
-    modalRoot.removeChild(this.el)
-  }
-
-  render() {
-    const { onClose, issue, onReviewApplication, onRequestAssignment, onSubmitWork, onAllocateSingleBounty, onReviewWork, onUpdateBounty } = this.props
-
-    return createPortal(
-      <div style={issueDetailStyle}>
-        <Navigation onClose={onClose} />
```
ottodevs: This was the previous IssueDetail that was handling the custom navigation bar, now is completely replaced - combined by it previous Detail child component (line 36 below)

Issue #881, opened 2019-05-07
Improve Projects List View


Quazia:
Implement the react-virtualized List component in order to improve the `Projects List View` https://bvaughn.github.io/react-virtualized/#/components/List

PR #904, opened 2019-05-09
[Blocked] 881 react-window implementation for large issues list

ottodevs:
Solves #881 

**UPDATE: Ready to review and tested with more than 500 issues, it works well, there are still some re-renders at projects app level, that should be solved when we refactor to use react hooks.**
 
Initial implementation, it still needs some additional work to make it more consistent
(Some rows are showing blank when scrolling fast trough the list)

The difficulty comes with handling each issue row individual props, trying memoizing approach for that

- Also fix prop camelCase at IconArrow.js


ottodevs:
> There's something weird going on with the scrollHeight. it _may_ be from merge but the cause is not instantly apparent to me. I fixed a quick issue where the scrollHeight got removed based on merge, but it doesn't look like it existed in state before this PR so I think there's also a decent chance the issue I'm seeing now was always present in this branch. I think this was _supposed_ to mainly deal with repo/projects view, but the issue is present in list view so make sure to check that. Please test and confirm the issue, I'll look more into if the problem was caused by a merge.

I think it is fixed with the recent commits 👍

ottodevs:
I detected an additional issue while reviewing the fixes:

the issue list virtualization may need some additional adjustment to allow dynamic heights, I detected some issues are overlapping because with current configuration the library expects every element inside the list to have the same hight
Is not that hard to change, is just I assumed that when I initially did the implementation, so I will change asap or submit a new issue about that

This should be the expected result:
image

Currently we have:
image

This does not seem so complex to fix, but will involve some research time and testing so whatever is more convenient, capture in other ticket or delay this PR review/merge until we fix that

apps/projects/app/components/Content/Issues.js
```diff
@@ -589,34 +641,55 @@ class Issues extends React.PureComponent {
...
+                          onUpdateBounty={this.handleUpdateBounty}
+                          onReviewWork={onReviewWork}
+                        />
+                      </div>
+                    )}
+                  </List>
+                </IssuesScrollView>
 
-                <IssuesScrollView>
+                {/* <IssuesScrollView>
```
Quazia: Remove this instead of commenting, this also orphans at least one function (showMoreIssues) so make sure to cleanup any unused functions and params along with this change.

ottodevs:
So, [react-window](https://github.com/bvaughn/react-window)  configuration should be modified to accept variable-size list elements, as shown here:
https://react-window.now.sh/#/examples/list/variable-size

apps/projects/app/components/Content/Issues.js
```diff
@@ -653,18 +652,16 @@ class Issues extends React.PureComponent {
                   <List
                     height={scrollHeight}
                     itemCount={preparedIssues.length}
-                    itemSize={92}
+                    estimatedItemSize={index =>
+                      preparedIssues[index].labels.totalCount > 0 ? 137 : 97
```
Quazia: Whether or not the item is funded also effects that size of the issue, right now adding funding to an issue appears to break the height.

chadoh: Great catches, @Quazia. I also noticed that the height of items wasn't updating when changing the filters or sorting, but I got that working. Check it out:

VariableSizeList behavior when filtering and sorting

---

However, I'm not sure how robust we can make the experience on *mobile*. Here's what it looks like now:

VariableSizeList behavior on narrow screens

---

 I think we have two options for fixing it:

1. Try to reverse-engineer the browser's layout engine for a given item, using the container width + content length to guess at how the content will flow unto multiple lines. I foresee this having endless bugs, and becoming out of date anytime we make subtle styling changes that don't seem immediately related to this list (example: if the next version of aragonUI tweaks what `size="normal"` means)
2. Re-design these cards so that they have a perfectly predictable height

I don't like either of these options, and I wonder if it's best to leave it as-is for now, and circle back to getting this pixel-perfect once we've had more time to play with it.

(But what do I mean by "as-is"? This PR as-is? Or the codebase as-is, without this PR? I haven't experienced the pain that this PR is trying to fix, so I don't know; I'll leave that up to someone else.)

chadoh: Note: I don't think there's an easy to configure `react-window` to update the item height with each item's "natural height". I've asked about this here: https://github.com/bvaughn/react-window/issues/263#issuecomment-501462501

Quazia: The main pain-point this PR is trying to resolve is having to click "show more" instead of having full scrolling for the issues. I think leaving this "as-is" for now is the best course of action as having list items spill over or render with the incorrect height is a bigger bug that this was meant to solve.

chadoh:
## To do

Copied from #966, since these things are specific to this branch:

- [x] Get rid of the rounded edges for the inner rows - It’s like two 1px borders are stacked and the border is 2px now. 
- [ ] Truncate issue name after two lines

e18r:
Hello @ottodevs @chadoh @Quazia, I'm going to work on this issue. From what I've gathered so far, I think [this](https://github.com/bvaughn/react-virtualized/blob/master/docs/CellMeasurer.md) can be a better solution to the current problem. What do you think?

chadoh:
I believe we've decided that we're not going to use this approach. Closing this to reduce the noise in our opened Pull Request list.

Issue #919, opened 2019-05-20
Modernize Projects


stellarmagnet:
- [x] Update to @aragon/api-react
- [x] Implement local identities

PR #936, opened 2019-05-28
919 Modernise projects app

ottodevs:
# Modernize projects app

## Summary

This pull request addresses the needed changes to modernize the Projects app to use recent versions of dependencies and Aragon framework modules.

Solves #919 , but it involves additional changes needed to perform that feature, like updating React and Babel.

Dependency updates were done on every app since lerna hoisting forces us to have the same versions of dependencies on some cases and the process to isolate each independent module needs is very time consuming, so I decided to just update the related dependencies on every app

## Tasks

- [x] Update Parcel, Babel, ESlint devDependencies and related packages 
- [x] Update babel config on every app 

- [x] Update react, react-dom and prop-types packages

- [x] Install missing @aragon/api-react dependency
- [x] Implement into needed Projects components
- [x] Replace networkProvider with useNetwork

- [x] Update @aragon/ui to 0.33 with custom labels support
- [x] Implement intermediate components (IdentityManager and LocalIdentityBadge)
- [x] Replace IdentityBadge occurrences with LocalIdentityBadge

### Misc related tasks/fixes
- [x] Fix some problematic rxjs imports that were causing some console error messages
- [x] Add missing dependencies for some apps (like address-book on allocations), probably deleted by error
- [x] Lint files with `npm run lint -- --fix` 
- [x] Regenerate package-lock.json with `npm run clean:commit`

### Preliminary QA
- [x] Reinstall everything from clean environment without console errors
- [ ] Pass full test script document manually
- [ ] Check both `npm run start:dev` and `npm run start` continue working correctly


apps/projects/app/App.js
```diff
@@ -148,14 +73,14 @@ class App extends React.PureComponent {
...
 
-  getChildContext() {
-    const { network } = this.props
-    return {
-      network: {
-        type: network.type,
-      },
-    }
-  }
+  // getChildContext() {
```
Quazia: Delete instead of comment

apps/projects/app/App.js
```diff
@@ -1,120 +1,49 @@
...
-  const json = await response.json()
-
-  return json.token
-}
+const ASSETS_URL = './aragon-ui'
 
 class App extends React.PureComponent {
   static propTypes = {
-    app: PropTypes.object.isRequired,
+    // is not required, since it comes async
```
Quazia: Could you expand on this?

apps/projects/app/index.js
```diff
@@ -10,53 +20,6 @@ import App from './components/App/App'
...
-  constructor(props) {
-    super(props)
-    this.state = {
-      app: new Aragon(new providers.WindowMessage(window.parent)),
-      network: {},
-      observable: null,
-      userAccount: '',
-    }
-  }
+// TODO: React strict
```
Quazia: Could you expand on this TODO

Quazia:
State management doesn't appear to be working in projects.

apps/projects/app/index.js
ottodevs: I wanted to try https://reactjs.org/docs/strict-mode.html but it has lot of potential errors right now, so it should be done after we refactor to spot any missing problem there.

apps/projects/app/App.js
ottodevs: app prop was renamed to api prop, previously app prop was a required prop (with isRequired) now, it cannot be required, because first it comes as null, then asynchronously gets a value.
Because at some initial point its value is undefined then it cannot be required anymore.

ottodevs:
I am confused about the last 2 comments, but since we were talking about this and the PR is merged I'll assume you gave an OK to this.

I take note of every comment and I will address the cleaning and polish into the next refactor task, that will involve migrating to hooks and cleaning some big components a bit, like App.js, Issues.js, IssueDetails.js into more reusable and testable pieces.

PR #953, opened 2019-06-03
Fund Issue Comments

e18r:
This is a proof of concept for a bot that posts bounty status updates to Github. Currently, the bot only posts updates when issues receive funding for the first time.

Things to consider:
* Doesn't include a URL since routing doesn't seem to be configured in the Planning Suite
* The comment is posted automatically using the account of the current user, rather than an independent (bot) account

### DIFF
--- a/apps/projects/app/App.js
+++ b/apps/projects/app/App.js
@@ -254,19 +254,20 @@ class App extends React.PureComponent {
     }))
   }
 
-  onSubmitBountyAllocation = async (issues, description) => {
+  onSubmitBountyAllocation = async (issues, description, post, resultData) => {
     this.closePanel()
 
     // computes an array of issues and denests the actual issue object for smart contract
     const issuesArray = []
     const bountyAddr = this.props.bountySettings.bountyCurrency
 
-    let bountyToken, bountyDecimals
+    let bountyToken, bountyDecimals, bountySymbol
 
     this.props.tokens.forEach(token => {
       if (token.addr === bountyAddr) {
         bountyToken = token.addr
         bountyDecimals = token.decimals
+        bountySymbol = token.symbol
       }
     })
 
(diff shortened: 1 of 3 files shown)


apps/projects/app/App.js
```diff
@@ -294,7 +294,13 @@ class App extends React.PureComponent {
       tokenArray,
       ipfsString,
       description
-    )
+    ).subscribe(
+      txHash => {
+        post({ variables: { body: `This issue has a bounty attached to it.\nAmount: ${issuesArray[0].size} autark\nDeadline: ${issuesArray[0].deadline}`,
```
Quazia: The bounty isn't necesarily always autark tokens, and I'd make sure that both the deadline and size are displayed in human readable formats. The decimals for the size will need to be inferred from the token.

e18r: @Quazia, I made the changes you requested:
* **Token Symbol**: it now changes with every different token, as you requested.
* **Deadline format**: it's a JS date object. In this case, I believe the `.toUTCString()` method to be the most human-friendly, because the Github message won't change depending on the user's timezone.
* **Amount format**: this raises some sort of UX dilemma. If the token has 18 decimals, as many ERC-20 tokens do, *and* the amount isn't round, what is more human-friendly? to show the full 18 decimals for the sake of precision (we're talking money after all), or to round the amount to some fixed number of decimals? I don't pretend to solve that here; instead I'll just follow the [common practice](https://github.com/AutarkLabs/planning-suite/blob/dev/apps/projects/app/components/Panel/FundIssues/FundIssues.js#L212).

Quazia: Should actually post on each issue:

```
        issuesArray.forEach(issue => {
          post({ variables: {
            body: `This issue has a bounty attached to it.\nAmount: ${issue.size.toFixed(2)} ${bountySymbol}\nDeadline: ${issue.deadline.toUTCString()}`,
            subjectId: issue.key } })
        }),
        err => console.log(`error: ${err}`)
```

Quazia:
Got it working locally, but when testing I realized you're only commenting on the _first_ issue. I've provided code to post to each issue. After that it should be good.

e18r:
Nice. How do I reproduce a situation in which the issue array has more than one issue?

Quazia:
Check multiple checkboxes in the issue view and use the action menu in the top right to fund issues.

e18r:
@Quazia I updated with the code you provided. It works well.

PR #959, opened 2019-06-07
Projects: Simplify "SearchFilterAction" row styling

chadoh:
This removes the Viewport component, which added an `onResize` listener to the page in order to alter the order of the markup.

With CSS Grid, we don't need to actually reorder the markup in order to change the placement of elements based on a media query. This accomplishes the same behavior without using `Viewport`.

Other styling/markup issues address while here:

* Make the "Clear Filters" button an actual button
* Vertically center the active filters and Clear Filters button
* Run `prettier` on all edited files

## Screenshot

SearchFilterAction responsive behavior

Here's the old behavior, for comparison:

SearchFilterAction old behavior

## Old vs New, narrow view

Old:

old

---

New:

new


apps/projects/app/components/Content/Filters.js
```diff
@@ -79,42 +96,39 @@ export default class Filters extends Component {
...
 
     return (
-      <div style={{
-        marginLeft: '8px',
-        marginTop: '10px',
-        flexDirection: 'row',
-        display: 'flex',
-        flex: '1',
-        flexWrap: 'wrap'
-      }}>
```
chadoh: The necessary change here was removing all of this styling expect for `display: flex`. This has been moved to the styled component called `Wrap`.

chadoh: All the other changes were either indentation-related or a result of running `prettier`.

apps/projects/app/components/Content/FilterTile.js
```diff
@@ -11,30 +12,38 @@ export default class FilterTile extends PureComponent {
...
           </div>
         </Badge.Identity>
-      </div>
+      </Tile>
     )
   }
 }
+
+const Tile = styled.div`
+  display: inline-block;
```
chadoh: This is the reason I had to had to edit this file—to add `display: inline-block` to the tile. This allows these tiles to all stack in side-by-side like this:

projects with many filters added

apps/projects/app/components/Content/Filters.js
```diff
@@ -79,42 +96,39 @@ export default class Filters extends Component {
...
+              />
+            )
+          })}
+        </div>
+        {Object.keys(filterAliases).length > 0 && (
+          <Button
+            mode="text"
+            size="small"
             onClick={this.props.disableAllFilters}
-            role='button'
```
chadoh: In general, I'm in favor of using `button` elements for buttons, rather than divs with `role="text"`. You may have noticed I left other such divs in place, though, in this PR. Why?

I tried to stay focused and only change things related to the styling issue at hand. This `div` wasn't vertically centered correctly. The button is.

chadoh:
Great catch, @rkzel. Thanks! I've updated the styles.

chadoh:
Just a note about this:

> disabling of the context menu

the logic about that didn't change, but the new code does make it much easier to follow the logic 😄

PR #968, opened 2019-06-10
Added the linter step

e18r:


e18r:
I forgot to include the command for the tests. Is it

    npm run test

?

PR #990, opened 2019-06-14
Added test dependencies

e18r:
I'm not sure why, but `npm run test` works when you want to run tests for all apps, but when you want to run tests only for the projects app, you `cd apps/projects` and then `npm run test` fails. This solves that.


Quazia:
Did you try running `npm run test:projects`? With a lerna repo you should really only be running commands from the root directory.

PR #992, opened 2019-06-14
[#926] killBounty support

### LINKED ISSUE #926: Contract change to support KillBounty interface
Currently, [Projects.sol](https://github.com/AutarkLabs/planning-suite/blob/dev/apps/projects/contracts/Projects.sol) does not support the [KillBounty](https://github.com/AutarkLabs/planning-suite/blob/dev/apps/shared/test-helpers/contracts/lib/bounties/StandardBounties.sol#L363) function in StandardBounties.sol.

In order for us to kill bounties from the Projects app, we need to connect this function to our Projects.sol contract.
This requires adding KillBounty to the Projects [Bounties interface](https://github.com/AutarkLabs/planning-suite/blob/dev/apps/projects/contracts/Projects.sol#L31).

In addition, we need an extremely simple function to clean up any local bounty state in Projects.sol and trigger the KillBounty function on the StandardBounties contract. This function should have a unique ROLE associated with it and should trigger an internal function. These functions should have the following signatures:
```
function removeBounties( bytes32[] _repoIds, uint256[] _issueNumbers) public auth(REMOVE_ISSUES_ROLE)
function _removeBounty( bytes32 _repoId, uint256 _issueNumber) internal
```
The [kit](https://github.com/AutarkLabs/planning-suite/blob/dev/apps/planning-suite-kit/contracts/PlanningKit.sol#L249) will also need to be updated to set the `REMOVE_ISSUES_ROLE`, this role will need to be added to the [arapp.json](https://github.com/AutarkLabs/planning-suite/blob/dev/apps/projects/arapp.json) file, and the function will need to be tested in [projects.test.js](https://github.com/AutarkLabs/planning-suite/blob/dev/apps/projects/test/projects.test.js).

e18r:
fixes #926 

* I had an issue with the refund functionality. As @Quazia helped me figure out, the StandardBounties contract was trying to refund the Projects contract, instead of the actual user. This could work in such a way, that the Projects contract is refunded and then refunds the user. Even though that might be the cleanest option, it requires knowledge about `AragonApp` contracts that I currently don't possess. Instead, I modified the StandardBounties contract so that a `refundee` address can be specified [on bounty creation](https://github.com/AutarkLabs/planning-suite/pull/992/files?file-filters%5B%5D=.json&file-filters%5B%5D=.sol#diff-8130d1a20d069683a6ed33da9ae39fa0R161); that's the user's address, which [will be refunded](https://github.com/AutarkLabs/planning-suite/pull/992/files?file-filters%5B%5D=.json&file-filters%5B%5D=.sol#diff-8130d1a20d069683a6ed33da9ae39fa0R382) directly if the bounty is killed. I'm not an expert, but I think this change doesn't carry security issues because only the bounty issuer can specify the refundee address.

* I didn't test killing bounties funded with ERC-20 tokens. I simply didn't know how to do that. Instead, I just added a [todo test](https://github.com/AutarkLabs/planning-suite/pull/992/files#diff-705875336e7af8299ec3f12c748ffb0dR926).

* I had to [comment out](https://github.com/AutarkLabs/planning-suite/pull/992/files#diff-705875336e7af8299ec3f12c748ffb0dR980) the testing of specific revert messages. They [work](https://ethereum.stackexchange.com/a/45613/7318) since Solidity 0.4.22, and in my machine they worked, but for some reason they didn't work in Travis CI. Is Travis CI using an outdated solidity version? I'm not sure.

### DIFF
--- a/apps/planning-suite-kit/contracts/PlanningKit.sol
+++ b/apps/planning-suite-kit/contracts/PlanningKit.sol
@@ -267,6 +267,7 @@ contract PlanningKit is KitBase {
 
         // Projects permissions:
         acl.createPermission(root, projects, projects.FUND_ISSUES_ROLE(), voting);
+        acl.createPermission(root, projects, projects.REMOVE_ISSUES_ROLE(), voting);
         acl.createPermission(voting, projects, projects.ADD_REPO_ROLE(), voting);
         acl.createPermission(voting, projects, projects.CHANGE_SETTINGS_ROLE(), voting);
         acl.createPermission(dotVoting, projects, projects.CURATE_ISSUES_ROLE(), voting);
--- a/apps/projects/arapp.json
+++ b/apps/projects/arapp.json
@@ -6,6 +6,11 @@
       "params": []
     },
     {
+      "name": "Remove funding from issues",
+      "id": "REMOVE_ISSUES_ROLE",
+      "params": []
+    },
+    {
       "name": "Add project repos",
       "id": "ADD_REPO_ROLE",
       "params": []
--- a/apps/projects/contracts/Projects.sol
+++ b/apps/projects/contracts/Projects.sol
@@ -32,6 +32,7 @@ interface Bounties {
 
     function issueBounty(
         address _issuer,
+        address _refundee,
         uint _deadline,
         string _data,
         uint256 _fulfillmentAmount,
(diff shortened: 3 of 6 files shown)


e18r:
I also tried to emit a debug event in StandardBounties, but I didn't manage to emit it. It's so hard to get debug information!

e18r:
Also, what do you think of this? https://sol-trace.com/

Issue #903, opened 2019-05-09
At projects app, responsive menu button seems to be wrong aligned


ottodevs:
Should be more like: 
![image](https://user-images.githubusercontent.com/5030059/57418943-babe9a80-7204-11e9-9a40-b421973687e9.png)

But it looks like:
![image](https://user-images.githubusercontent.com/5030059/57419212-d033c480-7204-11e9-88a3-a9b0cce3666e.png)

- [x] MenuButton position, margin and size should match other aragon apps
- [x] Check for every TPS app not just projects
- [x] MenuButton should appear as soon as the MenuPanel is hidden (it happens before the correct width right now)

I tried to do that myself quickly but just modifying Button padding didn't work, it seems another CSS rule is blocking the change

PR #993, opened 2019-06-15
Allocations: aragon/api-react and local Identities

rkzel:
This solves:

 - planning-suite#917 - Modernize Allocations

Both api-react and local identities were based on Projects App. Next modernized application should probably have local Identities moved to "/shared" folder available to all our apps.


apps/allocations/app/app-state-reducer.js
```diff
@@ -0,0 +1,11 @@
+function appStateReducer(state) {
+  const { github, repos } = state || {}
```
ottodevs: Is this correct? It seems that github and repos should not be present on the allocations app

apps/allocations/app/components/App/App.js
```diff
@@ -110,59 +111,76 @@ class App extends React.Component {
...
+      .requestAddressIdentityModification(address)
+      .toPromise()
+  }
+
   render() {
     const { panel } = this.state
     const { displayMenuButton = false } = this.props
     const PanelContent = panel.content
+
+    console.log(this.props)
```
ottodevs: Should we clean console.logs? Or at least put them into an intermediate environment conditional "log" function as John and you brilliantly did on the profiles app

apps/allocations/app/components/App/App.js
```diff
@@ -110,59 +111,76 @@ class App extends React.Component {
...
-            app={this.props.app}
-          />
-          <Payouts
-            payouts={
-              this.props.payouts !== undefined ? this.props.payouts : []
+        <IdentityProvider
+          onResolve={this.handleResolveLocalIdentity}
+          onShowLocalIdentityModal={this.handleShowLocalIdentityModal}>
+          <AppView
+            padding={0}
```
ottodevs: This will probably conflict with #903 PR, not an issue right now, just something worth to take care when merging

apps/allocations/app/components/App/App.js
```diff
@@ -110,59 +111,76 @@ class App extends React.Component {
...
-          title={(panel.data && panel.data.heading) || ''}
-          opened={panel.visible}
-          onClose={this.closePanel}
-        >
-          {panel.content && <PanelContent {...panel.data} />}
-        </SidePanel>
+          >
+            <Accounts
+              accounts={
+                this.props.accounts !== undefined ? this.props.accounts : []
```
ottodevs: the ternary is probably not needed here, just `this.props.accounts || []`

apps/allocations/app/components/App/App.js
```diff
@@ -110,59 +111,76 @@ class App extends React.Component {
...
+              }
+              onNewAccount={this.newAccount}
+              onNewAllocation={this.newAllocation}
+              app={this.props.api}
+            />
+            <Payouts
+              payouts={
+                this.props.payouts !== undefined ? this.props.payouts : []
```
ottodevs: the ternary is probably not needed here, just `this.props.payouts || []`

apps/allocations/app/components/Shared/LocalIdentityBadge.js
```diff
@@ -0,0 +1,89 @@
+import React from 'react'
```
ottodevs: If the code for this file is the same that the one used on Projects, I think it should be better to put it on shared/ui folder and import from there at both app, since we are going to put extra effort to de-duplicate these kind of files

apps/allocations/app/components/Shared/IdentityManager.js
```diff
@@ -0,0 +1,71 @@
+import React, { useCallback, useContext, useEffect, useState } from 'react'
```
ottodevs: If the code for this file is the same that the one used on Projects, I think it should be better to put it on shared/ui folder and import from there at both app, since we are going to put extra effort to de-duplicate these kind of files

apps/allocations/app/index.js
```diff
@@ -1,57 +1,13 @@
 import React from 'react'
 import ReactDOM from 'react-dom'
-import Aragon, { providers } from '@aragon/api'
+import { AragonApi } from '@aragon/api-react'
+import appStateReducer from './app-state-reducer'
 import App from './components/App/App'
+import '@babel/polyfill'
```
ottodevs: this import for polyfill should be always the first line on the code, to make sure the transpilation is applied to the whole file

apps/projects/app/App.js
```diff
@@ -529,13 +525,6 @@ class App extends React.PureComponent {
...
     const navigationItems = [ 'Projects', ...(issueDetail ? ['Issue Detail'] : []) ]
 
-    // return (
-    //   <Main assetsUrl={ASSETS_URL}>
-
-
-    //   </Main>
-    // )
-
     console.log(this.props)
```
ottodevs: Thanks for this cleanup!! it is really a best practice, could we remove the console.log as well? or put it on an special `log` function, only enabled on dev environment, as shown on profiles app :)

apps/allocations/app/app-state-reducer.js
ottodevs: Try instead `accounts` and `entries` ? (or better try to identify and confirm the needed state from previous allocations code) :)

apps/allocations/app/components/App/App.js
rkzel: Cleaned. A agree conditional debugging will be a great thing to add - although... prior to introducing it we need to discuss strategy and do it properly :)

apps/allocations/app/components/App/App.js
rkzel: ok!

apps/allocations/app/components/App/App.js
rkzel: Correct, thx.

apps/allocations/app/components/Shared/IdentityManager.js
rkzel: There are two files, so it made sense to move them into separate folder:

shared/ui/components/identity/

apps/allocations/app/components/Shared/LocalIdentityBadge.js
rkzel: Yes. I'll update Projects along with modernizing the rest of the apps.

apps/allocations/app/index.js
rkzel: Ok!

apps/projects/app/App.js
rkzel: Debugging TBD, but I'll be cleaning up as I go :)

apps/allocations/app/app-state-reducer.js
rkzel: As far as I can see there is no need for any variables there... Testing now. I'm going to leave it in place though, as standard and one-day-might-be-needed thing?

PR #994, opened 2019-06-17
Fix MenuButton placement & behavior in all apps

### LINKED ISSUE #903: At projects app, responsive menu button seems to be wrong aligned

chadoh:
Fixes #903

The hamburger/menu icon is now placed correctly, and appears at the correct moment, for all TPS apps. 

As part of this change, I simplified the styles and removed padding in many spots, instead preferring that all padding comes from aragonUI components. Some unexpected side effects:

* `AppBar` does not automatically add the same padding (`30px`) as the main `AppView` body, but `NavigationBar` does.
  * This means that Projects, which uses `NavigationBar`, does not need to add left padding to `AppTitle`, but all the other apps do (because none of them use `NavigationBar`). 
  * I would like to update aragonUI's `AppTitle` to add this padding, and `NavigationBar` to omit it. When this happens, all the spots where I've added `css="padding-left: 30px"` should go away, as well as using `className` in `AppTitle`
* I chose to simplify all of our CSS and defer to `AppView`, meaning that our mobile styles now all have `30px` of padding on the edges, same as the desktop styles. This is too much padding, but I think it's for the best.
  * We have fewer lines of code
  * All of our apps have consistent styling
  * When aragonUI is updated with better mobile styles, we'll be able to update easily

As I went, I also found some dead code that I removed. 

I recommend reading the commits in order.


ottodevs:
There are some conflicts that need to be addressed before merging this into dev, could you take care of those, please, @chadoh ?

ottodevs:
Hi @chadoh ! Did you check the comments from the PR? Also, there are some conflicts that need to be addressed, you can merge the PR after those are solved, I will try to solve the conflicts myself tomorrow to save some time if you are focused into other tasks :)

Let me know if you need assistance with the process 🌮

chadoh:
Conflicts resolved!

> did we already communicate the padding problems into an `aragon-ui` ticket?

I asked Pierre about this directly. aragonUI will have a fix for us soon, using new components. 

I'll merge once Travis finishes.

PR #1008, opened 2019-06-19
add script for running TPS without a client

Schwartz10:
@topocount @Quazia in case you ever need to run aragon with custom client this script adds support for that. 

I'm going to leave instructions on how to do so here: https://www.notion.so/autark/Running-TPS-with-local-repos-ad6df1dd9496441286f85f11cf0e5658


ottodevs:
LGTM, but I am a bit concerned about the amount of scripts we have on the package.json at some point it will be harder to remember all these than remembering the actual commands haha

For the moment this one will save some precious time for several use cases :)

ottodevs:
@Schwartz10 this PR is tagged as "In progress" is this accurate? otherwise we should merge the PR, I merged current dev branch, so it should be ready if no further changes are planned

PR #861, opened 2019-05-02
[Changes Requested] 587 e2e rewards

rkzel:
Set of tests is not complete, but it is still worth merging in, because it's a solid foundation for e2e test of all other apps.

PR #1010, opened 2019-06-19
Projects: move panel mgmt to React Context

### LINKED ISSUE #978: Projects: Move Panel management to React Context
This came out of #935.

One of the biggest areas of complexity in the Projects frontend code is the `App.js` file.

Part of what this file does is manage app-wide state, such as whether the side panel is open and what's shown in the side panel. This state and functions to modify it are then passed down through many layers of components.

One solution would be to use `redux` or a similar library to manage this sort of state. However, I'd like to avoid adding a new dependency if possible.

But I would like to accomplish two things:

* move as much as possible out of the definition of the App component, and after doing so move as much as possible into separate files
* make it easier to open side panels from anywhere in the app

I think using [React Context](https://reactjs.org/docs/context.html) might be a good way to reach these goals, but am open to other suggestions.

chadoh:
Fixes #978

This change slims down the `App.js` file in Projects by hundreds of lines, primarily by pushing logic to hooks and context.

## Reviewing

* you may find it useful to read the commits in order
* I suggest reviewing the file changes [with whitespace changes hidden](https://github.com/AutarkLabs/planning-suite/pull/1010/files?w=1)


ottodevs:
@chadoh the PR status is "In progress" is that accurate, or is it ready for review?

chadoh:
I am finding some issues with it still, maybe hold off on review. You can look over the code if you want, I don’t think the shape of it will change much, but don’t bother trying to run it locally.

apps/projects/app/components/Content/IssueDetail/IssueDetail.js
```diff
@@ -239,16 +240,7 @@ const activities = (
...
-  issue,
-  onReviewApplication,
-  onReviewWork,
-  onUpdateBounty,
-  onViewFunding,
-  onRequestAssignment,
-  onSubmitWork,
-  onAllocateSingleBounty,
-}) => {
+const detailsCard = ({ issue }) => {
```
Quazia: Remove curly brackets here since you're already de-structuring in IssueDetail.

ottodevs:
I will recheck this later, but initially I couldn't do the complete review because some blocking issues:

> If I fund 2 issues with 2 and 4 hours then I click on the first, but the issue detail does not load:
image

I think we need to review some props or issue members not getting passed correctly (`expLevel`)

Also I have some issues with Github authentication and 401 messages, but it eventually works, I dunno if that is related at all..

apps/projects/app/components/Panel/NewIssueCuration/NewIssueCuration.js
```diff
@@ -74,4 +76,68 @@ class NewIssueCuration extends React.Component {
   }
 }
 
-export default NewIssueCuration
+const onSubmitCuration = ({ closePanel, curateIssues }) => (
```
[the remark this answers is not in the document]
chadoh: I created #1095

apps/projects/app/components/Panel/NewIssueCuration/NewIssueCuration.js
```diff
@@ -74,4 +76,68 @@ class NewIssueCuration extends React.Component {
...
+  issues,
+  description
+) => {
+  closePanel()
+  // TODO: maybe assign this to issueDescriptionIndices, not clear
+  let issueDescriptionIndices = []
+  issues.forEach((issue, i) => {
```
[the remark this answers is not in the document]
chadoh: It's doing [the same thing it was before](https://github.com/AutarkLabs/planning-suite/pull/1010/files#diff-3fa6c42ccbaac15c982bbb6999c9097dL445). Do we need to change it now?

apps/projects/app/components/Content/IssueDetail/IssueDetail.js
chadoh: Oh, weird – note that it was doing that before, though! I guess this never worked?

chadoh: Oh, I see – `IssueDetail` used to receive `issue`, now it destructures `({ issue })`.

But it always _received_ `issue`, [not destructured](https://github.com/AutarkLabs/planning-suite/pull/1010/files#diff-1e61190978ae534ae80fe1f071b7e785L427)! Confusing.

Good catch, though. I'm still not sure if/why this worked before.

chadoh: Ok, right. `IssueDetail` should have called its args `props`, but it confusingly called them `issue`. And its `propTypes` are wrong – they should be nested inside `issue`. I'll clean this up. Thanks again!

chadoh:
> pulling the full context back to back (in usePanelManagement and then in the actual panels)

Note that the only `value` passed to `PanelContext.Provider` is `this.setPanel` from `App.js`. Given that `this.setPanel` is an object which never changes, none of the consumers of `PanelContext` should ever need to re-render (if I understand correctly how Context works). However, initially pulling in PanelContext may incur a one-time performance penalty; I have not measured this.

> I wish we had some concrete metrics to test loading times between different implementations as I don't think "this feels slow" is in any way possible to act on

I agree! That would be great. A quick search for how to automate such "performance regression checks" in our CI pipeline led me to [a blog post about using Lighthouse and Circle CI](https://stuartsandine.com/lighthouse-circle-ci/). I wonder if it makes to wait until we get our [e2e testing setup](https://github.com/AutarkLabs/planning-suite/pull/861) merged so we can explore Cypress a bit more, and see if there's a way to integrate Lighthouse or another performance-checker with our Cypress setup. 

This sort of research could easily take multiple days, though I think it would be worth it in the long-run. Want me to go ahead and make a ticket for it now?

> I see with NewIssue and NewProject are already structured as to consume state passed from usePanelManagement, do you envision all of the panels operating in this manner?

I'm not sure which pattern I like more here. I extracted `NewIssue` and `NewProject` from `App.js` later than most of the others, and hadn't come up with the idea of calling `useAragonApi` within `usePanelManagement` until then. 

Looking at it now, I think I'd prefer to get rid of that `useAragonApi` call within `usePanelManagement`, and have individual panels use it to get whatever state they need. No need for these two panels to get special treatment. I think it would be more explicit that way. I'll update that.

chadoh:
@ottodevs the error you ran into was caused by the invalid `detailsCard` arguments that Quazia noted. I've fixed this issue, and have not discovered any errors in my own local testing. Please re-test. 🙂

apps/projects/app/components/Panel/NewIssueCuration/NewIssueCuration.js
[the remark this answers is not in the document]
Quazia: Yeah, totally not relevant here it can be addressed in #1095 just this forEach is something special right now...

Quazia:
> I agree! That would be great. A quick search for how to automate such "performance regression checks" in our CI pipeline led me to a blog post about using Lighthouse and Circle CI. I wonder if it makes to wait until we get our e2e testing setup merged so we can explore Cypress a bit more, and see if there's a way to integrate Lighthouse or another performance-checker with our Cypress setup.
>This sort of research could easily take multiple days, though I think it would be worth it in the long-run. Want me to go ahead and make a ticket for it now?

I think capture a ticket and we can start looking into this post mainnet.

ottodevs:
I also commented with @chadoh about the -not that much annoying- `401 unauthorized` errors that we experienced on other branches as well, the issue is with the `useGithubAuth()` hook, line 17, maybe if we put a try-catch to handle that promise result we would have more control over the response, the same as my other comment, if we decide this is in the scope of this PR let's do it, should be trivial to add, otherwise I can capture the issue outside

PR #1047, opened 2019-06-24
Projects: simplify NewProject panel

chadoh:
This will make the effort in #1010 easier.

There is no way to open this panel when you are not authenticated with GitHub (at least not anymore). It therefore no longer needs the `status` or `onGithubSignIn` props.


chadoh:
@ottodevs @Quazia correct me if I'm wrong here, but I can't figure out a way to open the NewProject panel if you're not already authenticated. 

If I'm correct, then we can now remove all the code here. Let me know, so I can wrap up #1010!

stellarmagnet:
Once we decouple Github, we will need this feature. Something to keep in mind as you upgrade things

chadoh:
@stellarmagnet we'll need to revisit this component when we decouple from GitHub either way; I doubt the code being removed here would help us during that effort.

PR #1050, opened 2019-06-26
Discussions

Schwartz10:


Schwartz10:
Hey @chadoh! Thanks for adding those commits at the end. A couple things we should do before merging this in:

1. @Quazia should review the smart contracts briefly.
2. We should get a design and stub something a little nicer in the actual discussions app front-end (right now it looks like crap)
3. Fix the linter errors (I can take care of that unless you know of an easy way)
4. Hardcode as many dot votes as we want (maybe 100 to start?)

How does that sound to you? I feel like none of these things should take very long to complete.

I'll leave it up to you to squash and rebase these commits however you see fit. I'm fine with squashing them all into 1 commit if that's easiest.

apps/discussions/package.json
```diff
@@ -0,0 +1,67 @@
...
+    "compile": "aragon contracts compile",
+    "build:app": "parcel build ./app/index.html -d dist/ --public-url \".\" --no-cache",
+    "build:script": "parcel build ./app/script.js -d dist/ --no-cache",
+    "publish:patch": "aragon apm publish patch",
+    "publish:minor": "aragon apm publish minor",
+    "publish:major": "aragon apm publish major",
+    "publish:http": "npm run build:script && aragon apm publish major --http localhost:9999 --http-served-from ./dist",
+    "versions": "aragon apm versions",
+    "lint": "eslint . & solium --dir ./contracts",
+    "lint:fix": "eslint . --fix & solium --dir ./contracts --fix",
```
chadoh: Should these two lines use `&&` instead of `&`?

apps/discussions/package.json
```diff
@@ -0,0 +1,67 @@
+{
+  "name": "@tps/apps-discussions",
+  "version": "1.0.0",
+  "description": "",
+  "dependencies": {
+    "@aragon/api": "latest",
+    "@aragon/api-react": "latest",
```
chadoh: This differs from our other apps. Does that matter?

apps/discussions/app/App.js
```diff
@@ -0,0 +1,121 @@
...
+            Revise post
+          </Button>
+          <Button mode="secondary" onClick={hide}>
+            Hide post
+          </Button>
+          <Button
+            mode="secondary"
+            onClick={async () => {
+              const events = await api.pastEvents('0', 'latest').toPromise()
+              console.log(events, 'VENTS')
```
chadoh: @Schwartz10 what's this button for? I'd like to remove this `console.log` – can I remove the whole button?

apps/discussions/contracts/DiscussionApp.sol
```diff
@@ -45,12 +45,12 @@ contract DiscussionApp is IForwarder, AragonApp {
         post.author = msg.sender;
         post.postCid = postCid;
         post.discussionThreadId = discussionThreadId;
-        post.createdAt = now;
+        post.createdAt = now; // solium-disable-line security/no-block-members
```
ottodevs: Just for future reference, we are using the TimeHelpers from AragonOS for these cases, so you can perfectly replace `now` for `getTimestamp()` or `getTimestamp64()` depending on what data type is `post.createdAt` (uint or uint64), but these kind of changes will likely will be applied on future audits to these contracts, so no worries! just to share the info :)

NOTE: The danger will be there anywhere, but it should be prevented at contract design stage, by avoiding using block members for critical outcomes, since these values are gameable by miners

Source: https://github.com/aragon/aragonOS/blob/next/contracts/common/TimeHelpers.sol#L36
(TimeHelpers commands are automatically inherited by every AragonApp contract, and `now` is just an alias to `block.timeStamp`)

chadoh: > avoiding using block members for critical outcomes

Good to know about `getTimestamp` for AragonOS! In this case, I don't think the time at which a comment is created is a _critical outcome_, so I doubt it matters, but if I'd mistaken on that point I'd love for you to correct me!

Quazia: Yeah I mean they're functionally identical, the main advantage is in testing since you can create [a mock that overrides that call](https://github.com/aragon/aragon-apps/blob/master/apps/finance/contracts/test/mocks/FinanceMock.sol).

apps/discussions/contracts/DiscussionApp.sol
```diff
@@ -0,0 +1,125 @@
...
+        // should we limit the number of revisions you can make to save storage?
+        post.revisionCids.push(post.postCid);
+        post.postCid = revisedPostCid;
+        emit Revise(
+            msg.sender,
+            revisedPostCid,
+            discussionThreadId,
+            postId,
+            post.createdAt,
+            now // solium-disable-line security/no-block-members
```
Quazia: This is redundant since the time information should be accessible from the event either way.

apps/discussions/contracts/DiscussionApp.sol
```diff
@@ -0,0 +1,125 @@
...
+
+contract DiscussionApp is IForwarder, AragonApp {
+    using SafeMath for uint256;
+
+    event Post(address indexed author, string postCid, uint discussionThreadId, uint postId, uint createdAt);
+    event Revise(address indexed author, string revisedPostCid, uint discussionThreadId, uint postId, uint createdAt, uint revisedAt);
+    event Hide(address indexed author, uint discussionThreadId, uint postId, uint hiddenAt);
+    event CreateDiscussionThread(uint actionId, bytes _evmScript);
+
+    bytes32 constant public DISCUSSION_POSTER_ROLE = keccak256("DISCUSSION_POSTER_ROLE");
```
Quazia: Having a single role for all three actions isn't exactly ideal, what I would consider doing is splitting hide into two function calls with a seperate `HIDE_DISCUSSION_ROLE` and then keeping the one that exists but removing the role protections. As it is in most cases the auth check will have occured at the time of post, so this pattern only serves to _prevent_ a poster from hiding their own post if they've had posting permissions removed. I think it makes more sense for `hide` to be something you can _always_ do for your own posts and a separate role can do for other peoples posts.

apps/discussions/contracts/DiscussionApp.sol
```diff
@@ -0,0 +1,125 @@
...
+        uint discussionThreadId;
+        uint id;
+        uint createdAt;
+        bool show;
+        string[] revisionCids;
+    }
+
+    uint discussionThreadId;
+
+    mapping(uint => DiscussionPost[]) public discussionThreadPosts;
```
Quazia: Not sure how this will behave on an upgrade... this may need to be a mapping into another mapping instead of an array.

apps/discussions/contracts/DiscussionApp.sol
```diff
@@ -0,0 +1,125 @@
...
+    event CreateDiscussionThread(uint actionId, bytes _evmScript);
+
+    bytes32 constant public DISCUSSION_POSTER_ROLE = keccak256("DISCUSSION_POSTER_ROLE");
+    string private constant ERROR_CAN_NOT_FORWARD = "DISCUSSIONS_CAN_NOT_FORWARD";
+
+    struct DiscussionPost {
+        address author;
+        string postCid;
+        uint discussionThreadId;
+        uint id;
```
Quazia: I would specify unit256 for all params, and for indexes or anything else that's only being modified based on incrementation you can probably get away with a uint64.

apps/discussions/contracts/DiscussionApp.sol
```diff
@@ -0,0 +1,125 @@
+pragma solidity ^0.4.24;
+
+import "@aragon/os/contracts/apps/AragonApp.sol";
+import "@aragon/os/contracts/lib/math/SafeMath.sol";
+import "@aragon/os/contracts/common/IForwarder.sol";
+
+
+contract DiscussionApp is IForwarder, AragonApp {
+    using SafeMath for uint256;
```
Quazia: Doesn't look like you're using this anywhere.

PR #1051, opened 2019-06-26
Projects: Fix getIssuesForRepos query

chadoh:
GitHub's GraphQL API requires the `state` to either be `OPEN` or
`CLOSED`. The caps are required.


chadoh:
@Quazia I can't remember! I think the graphql query was returning an error instead of returning correct data.

Issue #920, opened 2019-05-20
Modernize Rewards


stellarmagnet:
- [x] Update to @aragon/api-react
- [x] Implement local identities

Issue #918, opened 2019-05-20
Modernize Dot Voting


stellarmagnet:
- [x] Update to @aragon/api-react
- [x] Implement local identities

Issue #916, opened 2019-05-20
Modernize Address Book


stellarmagnet:
- [x] Update to @aragon/api-react
- [x] Implement local identities

PR #1053, opened 2019-06-26
Modernized apps

rkzel:
This solves:

- Modernize Dot Voting #918
- Modernize Rewards #920
- Modernize Address Book #916

@ottodevs notes:
As part of the apps modernization, to ease the use of `aragonApi` hooks and to make it cleaner where possible, the App.js class components into functional components at AddressBook and Allocations (both were receiving changes to enable `network` context abstraction from `aragonApi` and to re-implement the needed `appStateReducer`, anyway the current reducer is just the minimal blueprint to make it work)

I also enabled `husky` hooks to aid spotting issues before hitting remote, so now, it will prevent commit if linter fails, and will prevent push if test fail... It also will prune and install deps on branch change (post-checkout hook), I hope this measures help with CI and with development workflow overall, but is just an initial setup that can be really easily customized.

These changes were attempts to find the previous issues, they were really useful for me and saved lot of time, but they are still opinionated, so we can rollback if we don't like them.


apps/address-book/app/index.js
```diff
@@ -1,48 +1,13 @@
+import '@babel/polyfill'
```
Schwartz10: Hmm this seems a little fishy to me. Why do you have to do this?

apps/dot-voting/app/app-state-reducer.js
```diff
@@ -0,0 +1,7 @@
+function appStateReducer(state) {
```
Schwartz10: What if state is null?

apps/address-book/app/app-state-reducer.js
```diff
@@ -0,0 +1,7 @@
+function appStateReducer(state) {
+  return {
+    ...state,
```
chadoh: Given that the `reducer` prop to `AragonApi` is optional, does it benefit us to have a no-op reducer at this point? Can we get rid of this whole file?

chadoh:
Should we address [this TODO](https://github.com/AutarkLabs/planning-suite/blob/b8aeae7596bba5db6b769be4a0fd912a126698f2/apps/dot-voting/app/components/AppView.js#L40-L46) while we're here?

> replace this component by @aragon/ui AppView when refactoring

apps/address-book/app/index.js
rkzel: To make it stop complaining about plugin-proposal-class-properties, as in Aragon apps:

https://github.com/aragon/aragon-apps/blob/master/apps/finance/app/src/index.js

However, it might have been a different issue - I retested, that error is gone. I removed these imports.

apps/address-book/app/app-state-reducer.js
rkzel: Yes. I guess there is no point in keeping code like that for future use - it can always be added when it's needed...

apps/dot-voting/app/app-state-reducer.js
rkzel: With removing appStateReducer from the equation this question became answered (indirectly).

apps/address-book/app/index.html
```diff
@@ -7,7 +7,7 @@
 
 <body>
 
-    <div id="root"></div>
+    <div id="addressbook"></div>
```
chadoh: Do we need to rename this?

apps/allocations/app/components/App/App.js
```diff
@@ -182,6 +182,6 @@ class App extends React.PureComponent {
 }
 
 export default () => {
-  const { api, appState } = useAragonApi()
-  return <App api={api} {...appState} />
+  const { api, appState, displayMenuButton } = useAragonApi()
+  return <App api={api} {...appState} displayMenuButton={displayMenuButton} />
```
chadoh: It looks like we need this, but I'm not sure why it got included on this PR. I guess it was missed during the #994 effort? Thanks for spotting this!

apps/address-book/app/components/App/App.js
```diff
@@ -1,17 +1,16 @@
...
 import NewEntity from '../Panel/NewEntity'
 import { networkContextType, AppTitle, AppTitleButton } from '../../../../../shared/ui'
+import { useAragonApi } from '@aragon/api-react'
+import { IdentityProvider } from '../../../../../shared/identity'
 
 class App extends React.Component {
   static propTypes = {
-    app: PropTypes.object.isRequired,
-    // TODO: Shape this
+    api: PropTypes.object,
```
chadoh: This is required, right? Won't things break if it's not present?

apps/dot-voting/app/App.js
```diff
@@ -20,7 +18,7 @@ const initialState = {
 
 class App extends React.Component {
   static propTypes = {
-    app: PropTypes.object.isRequired,
+    api: PropTypes.object,
```
chadoh: Same thing here; this is required, right?

apps/rewards/app/components/App/App.js
```diff
@@ -24,7 +24,7 @@ const convertApiUrl = symbols =>
 
 class App extends React.Component {
   static propTypes = {
-    app: PropTypes.object.isRequired,
+    api: PropTypes.object,
```
chadoh: Also required?

apps/dot-voting/app/Decisions.js
```diff
@@ -19,7 +19,7 @@ const EmptyIcon = () => <img src={emptyIcon} alt="" />
 
 class Decisions extends React.Component {
   static propTypes = {
-    app: PropTypes.object.isRequired,
+    app: PropTypes.object,
```
chadoh: Still required?

apps/dot-voting/app/Decisions.js
```diff
@@ -166,7 +166,7 @@ class Decisions extends React.Component {
                 icon={<EmptyIcon />}
                 title="You do not have any dot votes."
                 text="Use the Allocations app to get started."
-                actionButton={() => <div />}
+                onActivate={() => <div />}
```
chadoh: Was this just a typo you noticed while you were here?

apps/dot-voting/app/index.html
```diff
@@ -7,7 +7,7 @@
 
 <body>
 
-    <div id="root"></div>
+    <div id="dotvoting"></div>
```
chadoh: Do we need to rename this?

apps/projects/app/index.js
```diff
@@ -1,5 +1,3 @@
-import '@babel/polyfill'
```
chadoh: I love seeing code get deleted, and even more so whole packages! But I'm not sure why this is disappearing now, in particular. Mind explaining what enabled us to get rid of this here?

It looks like we're already importing it in every app's `script.js` – is that why?

I also noticed that it's in every app's `package.json` _except_ for `address-book`. Should we add it there while we're improving this behavior?

apps/address-book/app/components/App/App.js
rkzel: Actually no - initially there is no api, it is provided slightly later. If "isRequired" is set, there is a big warning in the console.

apps/allocations/app/components/App/App.js
rkzel: It was coming from a different place before 'api' was introduced. Now it needs to be here.

apps/address-book/app/components/App/App.js
chadoh: But the code in the component assumes that it's there, so it seems wrong to make it unrequired.

Maybe we could set `defaultProps` with an empty object, and add a comment there that `useAragonApi` doesn't provide `api` on initial load. Then we'll also have a reminder in the future to remove our `defaultProps` if `useAragonApi` fixes this inconsistency.

apps/address-book/app/index.html
rkzel: We don't need to rename it, but we had a mix of id=root and id=something in applications, so now we only have id=somethings. We could as well have id=roots :) What's the pro/con argument?

apps/dot-voting/app/Decisions.js
rkzel: No, this started appearing in warnings - actionButton being obsolete. It should be a separate PR, I know, but...

apps/dot-voting/app/index.html
rkzel: I cleaned up, made all standard. Not sure if there is any argument for or against? In any case sorry for the commit noise!

apps/address-book/app/index.html
chadoh: I like consistency! And this is a small change, so it may as well piggy-back on this one. I wasn't sure if the updated packages conflicted with the `id="root"`

apps/dot-voting/app/Decisions.js
chadoh: Oh, that seems related to me! We updated the library and fixed deprecation warnings. 👍

apps/projects/app/index.js
rkzel: At first I had errors in console about... something irrelevant. I was told it is because '@babel/polyfill' is missing, and it should be there (because it is present in Aragon apps already, etc). I just retested that and found the errors are still there. Therefore this import should stay in place in all apps.

ottodevs:
I think getting rid of `static childContextTypes = { network: networkContexType }` (For example, lines 17-23 on AddressBook app.js) Maybe that could be related to address book entries not working

apps/address-book/app/index.html
ottodevs: This should be just a naming opinion, my point about that is: I also like consistency, and semantically it sounds better for me to use the name of the app instead of the "cold" `root`

But I guess is really unimportant from the code / functionality perspective in any case

ottodevs:
Solved merge conflicts

apps/address-book/app/components/App/App.js
rkzel: Actually, it can't be provided with defaultProps, because it is provided (albeit, initially, null):
```
Warning: Failed prop type: The prop `api` is marked as required in `App`, but its value is `null`.
    in App (created by AppWrap)
    in AppWrap
    in AragonApi
```
This is after adding defaultProps and adding isRequired. So, what is being asserted now (I mean before I readded 'isRequired') is that if api is present and not null it should be an object. Soft requirement... We could in theory to this:

`PropTypes.oneOfType([null, PropTypes.object]).isRequired`

But it doesn't strike me as perfect solution. Is it the least wrong solution?

chadoh: I don't think that provides more information than leaving it optional. Thanks for looking into this. I'd like to see `@aragon/api-react` change this behavior in the future, but I think the way you implemented it initially is the best for now.

apps/dot-voting/app/Decisions.js
rkzel: No. The easiest thing to do now - because it is valid for the app to not be there initially - and because we probably don't want to make it overcomplicated with multi-level PropTypes (i.e., it can be `null`, it can be `object` - is to make it not required and just check that if it's present, it's an object.

apps/address-book/app/components/App/App.js
```diff
@@ -1,106 +1,79 @@
...
 
-  getChildContext() {
-    const { network } = this.props
-    return {
-      network: {
-        type: network.type,
-      },
-    }
+  const removeEntity = (address) => {
+    api.removeEntry(address)
```
chadoh: No need for this intermediate function. We can pass `onRemoveEntity={api.removeEntry}` instead.

apps/allocations/app/components/App/App.js
```diff
@@ -1,187 +1,140 @@
...
+      tokenAddress
     )
-    this.closePanel()
+    closePanel()
   }
 
-  onExecutePayout = (accountId, payoutId) => {
-    this.props.api.runPayout(accountId, payoutId)
+  const onExecutePayout = (accountId, payoutId) => {
+    api.runPayout(accountId, payoutId)
```
chadoh: Same thing here. The intermediate function doesn't really get us anything. Let's pass `api.runPayout` instead of `onExecutePayout`.

apps/allocations/app/components/App/Payouts.js
```diff
@@ -78,45 +87,54 @@ const PayoutsNarrow = ({ executePayout, data, tokens }) => (
             {payout.description}
           </PayoutDescription>
           <Text.Block size="small" color={theme.textTertiary}>
-            {showStatus(payout)}
+            {ShowStatus(payout)}
```
chadoh: Since this is a component, let's use jsx:

```jsx
<ShowStatus {...payout} />

File: apps/allocations/app/components/App/PayoutsTables.js [hunk 16]
```diff
@@ -1,35 +1,32 @@
...
+  listProps: PropTypes.object.isRequired,
 }
 
 const PayoutDescription = styled(Text.Block)`
   display: block;
+  /* stylelint-disable-next-line */
   display: -webkit-box;
   -webkit-line-clamp: 2;
+  /* stylelint-disable-next-line */
   -webkit-box-orient: vertical;
```
chadoh: We don't need these lines; styled-components adds backward-compatible styles for us. No need to disable the lines; just remove them.

apps/allocations/app/components/Card/Account.js
```diff
@@ -115,8 +117,10 @@ const CardTitle = styled(Text.Block).attrs({
   text-align: center;
   color: ${theme.textPrimary};
   display: block;
+  /* stylelint-disable-next-line */
   display: -webkit-box;
   -webkit-line-clamp: 2;
+  /* stylelint-disable-next-line */
   -webkit-box-orient: vertical;
```
chadoh: Same here. Remove browser-specific rules; they are added automatically by `styled-components`.

apps/dot-voting/app/App.js
```diff
@@ -65,51 +63,63 @@ class App extends React.Component {
...
   }
 
+  handleResolveLocalIdentity = address => {
+    return this.props.api.resolveAddressIdentity(address).toPromise()
+  }
+
+  handleShowLocalIdentityModal = address => {
+    return this.props.api
+      .requestAddressIdentityModification(address)
+      .toPromise()
```
chadoh: Do we have a ticket for moving all of this to the `IdentityProvider` component itself? It seems we want to do this for every app that uses it.

PR #1073, opened 2019-07-02
Address book add update entry

topocount:
Added an updateEntry function for compatibility with the Identity Provider API


apps/address-book/contracts/AddressBook.sol
```diff
@@ -72,13 +75,29 @@ contract AddressBook is AragonApp {
         emit EntryRemoved(_addr);
     }
 
+    /**
+     * @notice Remove address `_addr` from the registry.
```
stellarmagnet: is this radspec correct?

stellarmagnet: Can we return the old/new CID in the radspec? if this goes to the voting app as is, there is no context for the change

PR #1092, opened 2019-07-03
Dot Voting: fix "non-boolean attribute" error

### LINKED ISSUE #823: Warning when setting bool in Slider.js
RIght now we're getting a warning from Slider.js:
```
Warning: Received `false` for a non-boolean attribute `pressed`.

If you want to write it to the DOM, pass a string instead: pressed="false" or pressed={value.toString()}.

If you used to conditionally omit it with pressed={condition && value}, pass pressed={condition ? value : undefined} instead.
```

chadoh:
Fixes #823

This prop needs to be a number, but `state.pressed` is a boolean. As elsewhere in this component, we can coerce it to a number (1 or 0) using `Number(pressed)`.


ottodevs:
This file should just be a 1:1 copy of the original aragon-ui [Slider](https://github.com/aragon/aragon-ui/blob/master/src/components/Slider/Slider.js) but with some CSS changes (maybe at some point we can apply the CSS customization without overriding the whole file)

So probably the `pressed` issue should be handled the same way the did

Anyway, the changes seem ok and the warning message is gone, I've been trying the Slider extensively and I didn't observe anything wrong there, so I think the change is enough.

For future Slider refactor I 'd recommend to follow the changes upstream from @aragon/ui and migrate to use their component as soon as we can to resolve duplicities

PR #1098, opened 2019-07-05
Projects: cleanup getIssuesForRepos query

chadoh:
* Remove obsolete GET_ISSUES
* Don't use temporary variable
* Restructure getIssuesGQL to make it easier to read

### DIFF
--- a/apps/projects/app/components/Panel/NewIssue/NewIssue.js
+++ b/apps/projects/app/components/Panel/NewIssue/NewIssue.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types'
 import { Mutation } from 'react-apollo'
 import { useAragonApi } from '../../../api-react'
 import { Field, GU, TextInput, DropDown } from '@aragon/ui'
-import { NEW_ISSUE, GET_ISSUES } from '../../../utils/gql-queries.js'
+import { NEW_ISSUE } from '../../../utils/gql-queries.js'
 import { Form } from '../../Form'
 import { LoadingAnimation } from '../../Shared'
 import { usePanelManagement } from '../../Panel'
@@ -103,18 +103,12 @@ class NewIssue extends React.PureComponent {
 
     const id = selectedProject > 0 ? reposIds[selectedProject - 1] : ''
 
-    const reGet = [
-      {
-        query: GET_ISSUES,
-        variables: { reposIds },
-      },
-    ]
+    // TODO: refetch Issues list after mutation
 
     return (
       <div css={`margin: ${2 * GU}px 0`}>
         <Mutation
           mutation={NEW_ISSUE}
-          refetchQueries={reGet}
           variables={{ title, description, id }}
           onError={console.error}
         >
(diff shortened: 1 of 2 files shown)


apps/projects/app/components/Panel/NewIssue/NewIssue.js
```diff
@@ -100,17 +100,11 @@ class NewIssue extends React.PureComponent {
 
     const id = selectedProject > 0 ? reposIds[selectedProject - 1] : ''
 
-    const reGet = [
-      {
-        query: GET_ISSUES,
-        variables: { reposIds },
-      },
-    ]
+    // TODO: refetch Issues list after mutation
```
ottodevs: I wonder if this solution is better than the former `reget` for the time being? Would you like to put some more context on this change to help understand please?

apps/projects/app/components/Panel/NewIssue/NewIssue.js
```diff
@@ -100,17 +100,11 @@ class NewIssue extends React.PureComponent {
...
 
     return (
       <Mutation
         mutation={NEW_ISSUE}
-        refetchQueries={reGet}
```
ottodevs: It seems like something could be getting out-of-sync here by looking at the message (the stacktrace says it is happening on the new issue mutation)
image

ottodevs: This happens right after clicking the new issue creation button at the new issue panel

chadoh: Is that related to the `refetchQueries` prop? 

From what I could tell, we were originally fetching issues with this `GET_ISSUES` query. That would have made it so that when we called `refetchQueries`, Apollo cleared its cache for that query, refetched, and we got the newly-created issue to show up right away. However, when we added pagination, with the more complicated query, this `refetchQueries` prop never got updated, and the functionality broke.

Last I knew (early this year), expiring the cache for these sorts of complicated queries is more complex than you'd expect, with Apollo. That's why I put in a TODO instead of just updating it.

However, I'm not sure any of this would be related to the error that you're seeing.

apps/projects/app/components/Panel/NewIssue/NewIssue.js
chadoh: The `reget` stuff wasn't working anymore, ever since we stopped using `GET_ISSUES`. 

I think actually making this work may be somewhat complicated; do we have an existing ticket for fixing this?

stellarmagnet:
Super old -- I assume we don't need this anymore at this stage? If so, please close @chadoh

chadoh:
Still relevant, still awesome. @ottodevs I look forward to your re-review!

apps/projects/app/components/Panel/NewIssue/NewIssue.js
ottodevs: I checked and I confirm that it stopped to work prior to this issue, as you pointed out, so I created [a new ticket](https://github.com/AutarkLabs/open-enterprise/issues/1694) to track it.

I think the TODO note is also a good idea to have as a reminder.

PR #1100, opened 2019-07-08
Add CODEOWNERS

chadoh:
This file can be used [by GitHub] and [by Pull Panda] to help streamline
our code review process.

  [by GitHub]: https://help.github.com/en/articles/about-code-owners
  [by Pull Panda]: https://docs.pullpanda.com/products/assigner

### DIFF
--- a/docs/CODEOWNERS
+++ b/docs/CODEOWNERS
@@ -1,0 +1,12 @@
+# This is a comment.
+# Each line is a file pattern followed by one or more owners.
+
+# These owners will be the default owners for everything in
+# the repo. Unless a later match takes precedence,
+*       @ottodevs @Quazia @rkzel @topocount @chadoh
+
+# Order is important; the last matching pattern takes the most
+# precedence. When someone opens a pull request that only
+# modifies solidity files, only the following owners will be
+# requested for a review.
+*.sol    @Quazia @ottodevs @topocount


docs/CODEOWNERS
```diff
@@ -0,0 +1,12 @@
...
+
+# These owners will be the default owners for everything in
+# the repo. Unless a later match takes precedence,
+*       @ottodevs @Quazia @rkzel @topocount @chadoh @Schwartz10
+
+# Order is important; the last matching pattern takes the most
+# precedence. When someone opens a pull request that only
+# modifies solidity files, only the following owners will be
+# requested for a review.
+*.sol    @Quazia @Schwartz10 @topocount @rkzel
```
chadoh: @Quazia does this list look correct to you?

docs/CODEOWNERS
```diff
@@ -0,0 +1,12 @@
+# This is a comment.
+# Each line is a file pattern followed by one or more owners.
+
+# These owners will be the default owners for everything in
+# the repo. Unless a later match takes precedence,
+*       @ottodevs @Quazia @rkzel @topocount @chadoh @Schwartz10
```
chadoh: In addition to this global list of code owners, we could also add frontend code owners. Would that be useful? @ottodevs I know you would want to be on that list. Do we want that? And who all should be on such a list?

ottodevs:
I think this should be discussed better with the team to come with something consistent about the reviewers assignment, and what other indirect consequences will this `CODEOWNERS` file could add, is this defining or modifying anything related to the legal rights over the code or license or something like that?

I also would like to extract this info from the github api or a github hook instead of hardcoding the information manually, but I don't know if that is even possible nowadays

chadoh:
>  is this defining or modifying anything related to the legal rights over the code or license or something like that?

No, I don't believe "owners" is being used in the legal sense here. It's a casual, "who cares about X and wants to be part of conversations related to X changing?"

> extract this info from the GitHub API or a GitHub hook instead of hardcoding

I like having it recorded in a file. It seems straightforward and discoverable, and gives us visibility into any changes to this file via our review process.

docs/CODEOWNERS
Quazia: I think it should probably be @Quazia @ottodevs @topocount

docs/CODEOWNERS
Quazia: Generally I don't think it's good to rely on @Schwartz10 for reviews on TPS

Quazia:
I think using this tool is low risk so I'm in favor of moving forward with this. We can always just stop using it if we don't like it.

Issue #997, opened 2019-06-17
Projects: Clean up unnecessary re-renders


chadoh:
Spawned from #976 

## Prerequisites

- [Projects: Move Panel management to React Context](https://github.com/AutarkLabs/planning-suite/issues/978)
- [Projects: Move logic out of `App.js`](https://github.com/AutarkLabs/planning-suite/issues/979)

Working on these will ease the effort of reducing the number of unnecessary re-renders for the Projects app.

## Strategy

From there, some things to look at, in order of how easy they'll be to implement and maintain:

* Figure out if there are prop & state updates that can be prevented entirely – times when our app renders for some irrelevant reason, where we can be more strategic with how we update the state, or send PRs to aragonAPI to prevent prop updates
* Switch to React Hooks as much as possible, starting with the main App.js component
* Implement [`shouldComponentUpdate`](https://reactjs.org/docs/react-component.html#shouldcomponentupdate)

PR #1109, opened 2019-07-12
Projects: stub @aragon/api-react when not iframe

chadoh:
Precursor to #997

During development, it is sometimes convenient to load an app in the main browser window, outside of the iframe, such as when trying to use React Dev Tools to assess performance bottlenecks.

This breaks the connection to @aragon/api-react, which counts on a host page being available, and relays messages from that app into the iframe.

This sets up a stubbed @aragon/api-react, which only gets loaded in if

1. `NODE_ENV` is not production
2. the page is not in an iframe

It currently implements very few of the features of @aragon/api-react, but already enough to load the app and sign into GitHub.


sohkai:
This is really nice to see @chadoh!

I've been thinking about similar ways to improve DX for apps, and have had two primary lines of thought:

- You could add mock data and switch to displaying mock data when in a "non-connected" development mode
  - This is perhaps simpler, and similar to your approach of a mock
  - However, this is mostly only useful for layout changes, and not enough for an E2E test environment
- We could create a simpler version of the Aragon client that is only meant for development purposes, using the same `@aragon/wrapper` underneath the hood with the same RPC wiring, but with various things turned off (tx signing, option to turn off iframe sandboxing, etc.)
  - We can simulate the app in an environment similar to the Aragon client, but with great liberty to improve the development response/test cycle
    - Don't have to wait for ACL fetching / other slow downs, because we can use the app address directly
    - Can be configured to be bundled together with the app, via a different entry point, making it easy to automate staging builds

-----------

> During development, it is sometimes convenient to load an app in the main browser window, outside of the iframe, such as when trying to use React Dev Tools to assess performance bottlenecks.

On this, you can attach the React Dev Tools into the iframe by using [their electron app](https://github.com/facebook/react-devtools/tree/master/packages/react-devtools) and hooking it up in your index.js.

chadoh:
This PR was expanded to work with two separate apps, both Projects and now also Allocations.

Allocations demonstrates using complex `initialState` to show most of our UI components at first visit, while having a sparse set of `functions` for further manipulations to that state.

Projects demonstrates having a minimal "default state" (it looks like a lot, but it's what ends up in the state after first load, when the full dev env is running), and a function in `functions` which updates critical state.

I think we should merge this as-is and try it out. It can be extended fairly easily, and it's also now built in a way that will allow us to add a stubbed `useAragonApi` to our other apps for quick frontend development.

@Quazia @ottodevs @e18r do you feel good about merging this now?

ottodevs:
We have been talking about that, but probably summarising my opinions written here is a better way:

- For me, the frontend development should involve just the CSS work, layout of components and minor logic related to frontend like for animations.
- The desired workflow and the one that fits better IMHO would be mocking the state to achieve the desired design screen to be able to fine tune the layout and css there.
- To follow this workflow, you don't even need to have logic like click a button to open a panel, just set the `panel.visible` to `true` in your state
- We don't even need to define an `App.js` full state, you can instead define mocked states at sub-component levels, by temporarily replace the received props by the mocked state you need, once you finish the FE implementation you can revert the component to use its original props again.
- That said you still can define a full initial state in your root App component, if you want to, so the degree of freedom is high
- Adding logic for functions is out of scope of the frontend development, and requires making the work twice: implementing the actual logic in the background script and then mocking a modified logic that looks similar to the background script but is adapted, like mocking a function that returns and observable to be able to use `toPromise()` am I the only one that sees that this adds lot of cumbersome just to mimic something you don't really need?
- It is quite cool to have an app with its logic working fully detached from the aragon wrapper, but I honestly think that is not really useful, since is not adding value to the projects and you cannot generalize to every aragon app for others to profit from it

What I realize with this things is that we need to better document the correct way of developing an aragon app frontend, because it may result confusing right now, and also defining better proptypes (shaping proptypes to have a kind of "state schema") to exactly know what kind of state we need to define in every situation, both of these solutions seem helpful for me.

We could also hardcode some mocked state for reference, but again, I don't think we want to also mock the logic. Anyway Radek and me, and even Arthur, probably Kevin as well were eventuallyy also mocking state for our apps to aid with the frontend development, but at some point we decided to clean that mocked state from the apps and just define it locally when we need to...

If that mocked state becomes something really needed, we can discuss  adding some json files with mocked state to the apps for fun and profit, but again, I think mocking the aragon-api calls I see that a bit overengineered, even if that is easy, since mocking the state to achieve the design screen will always be easier than mocking a function that generates the same state you want...

For any other use case, we spent a lot of time integrating Cypress, and I still think we should put our efforts towards learning and using it instead.  

Sorry for my criticism!! I understand your intentions are really good, I just wanted to put my opinion on this ❤️

chadoh:
It sounds like we agree that we enjoy having the ability to stub the frontend, and we have two options.

# Option 1 (Otto's suggestion)

> * The desired workflow and the one that fits better IMHO would be mocking the state to achieve the desired design screen to be able to fine tune the layout and css there.
> * To follow this workflow, you don't even need to have logic like click a button to open a panel, just set the panel.visible to true in your state
> * We don't even need to define an App.js full state, you can instead define mocked states at sub-component level, by temporarily replace the received props by the mocked state you need, once you finish the FE implementation you can revert the component to use its original props again.
> * That said you still can define a full initial state in your root App component, if you want to

Downsides to this approach:

* Anyone who wants to stub the frontend needs to have a high degree of familiarity with the data structures required
* Rather than sharing effort by checking stubbing-related code into the repo, everyone duplicates the effort of stubbing out the frontend over and over

# Option 2 (this PR)

* Once you figure out how to stub the frontend, rather than stub your own local state, put it in a simple api-stubbing function or in the shared `initialState`, which gets checked into the codebase but does not get included in our production bundle
* If the frontend for the app you're working on already has stubbing logic, all you need to do is `cd` into its directory and `npm run start:frontend`

Downsides to this approach:

* Stubbing a function is more complicated than stubbing `initialState` or local state

   Honestly, it's not much more complicated, in most cases where I did this so far. (The most complicated one, for Projects, required cleaning up our actual Projects code a bit in order to write it, but that's a _benefit_, not a disadvantage!) If returning Observables feels awkward, we can do that at the `shared/api-react` level, since that's a better approximation of `@aragon/api-react` anyhow.

* These stubbed functions and initialState cannot be automatically generated, and will at times become out-of-sync with our actual contract & script.js logic.

In both of these cases, I think the benefit—having generic stubbed functions which provide a interactive frontend development environment, allowing anyone to run the frontend and quickly set up their own test data without figuring out opaque data structures and how to insert them into the app—far outweighs the disadvantages.

# Other cool possibility: prototyping!

Long-term vision: get the stuff in `shared/api-react` merged into `@aragon/api-react`. We need to test drive it for a while and work out some kinks, and maybe stub out more of the core aragonAPI functions.

I think a place this could really shine, which we might not have a chance to try out for a while, is _initial app development_. Rather than needing to get a whole blockchain+contracts+script.js+frontend working just to get started, this allows a team to quickly prototype an app frontend. This is huge! After locking in the frontend and deciding on exact UX flow, then a team can implement contracts and script.js logic to make it permanent.

chadoh:
Here's how this currently works with Dot Voting:

frontend-only demo

As you can see:

* it spins up in just a couple seconds
* you can create votes directly in the console; no need to switch to a different app to test the whole flow
* yes, not everything is stubbed, so there are some ugly console errors, but stubbing doesn't need to be complete to be useful

Issue #1124, opened 2019-07-18
Disable the github bot


stellarmagnet:
1) Disable github bot until we have deep linking
2) Regardless, always disable the Github in the dev environment (unless we are testing the bot)

Otherwise our repos get cluttered with "test" comments like below:

![image](https://user-images.githubusercontent.com/2584493/61445665-19636b80-a94e-11e9-8895-71695760ed7a.png)

ottodevs:
This issue seems to be solved, @stellarmagnet feel free to re-open if you feel something is missing, I close it for now.

PR #1133, opened 2019-07-22
Temporarily disable comment on issue funding

ottodevs:
Address #1124 
Tested locally, apparently no errors or malfunction when disabling this code block


ottodevs:
> I'd love a `TODO` comment, explaining why it's commented out and when we plan to re-enable it. No need for me to re-review when/if you add that, though.
> 
> Longer term, let's think of a way to only include this in production deploys. An environment variable, maybe.

Thanks! It makes sense, I added the TODO comment and a possible future workaround for consistency, but we probably need to plan a definitive solution in advance

PR #1142, opened 2019-07-23
Dot voting changes

Quazia:
Current dot-voting changes needed for internal audit.


ottodevs:
This branch is integrated into #1154 including the three commits on it with the same history, so I think this one can be eventually closed and deleted.

PR #1144, opened 2019-07-26
Lint fixes

rkzel:
After recent linting configuration changes we're getting tons of errors in different places. This PR addresses:
 - npm run lint
 - git precommit hook
Inside: all props are validated, most are even marked as isRequired. Some that are allowed to be not present were left alone. Minor code fixes were added where appropriate. Styling should have stayed exactly the same...


apps/allocations/app/components/App/Payouts.js
```diff
@@ -163,7 +168,7 @@ const PayoutsWrap = styled.div`
   ${breakpoint(
     'small',
     `
-    padding: 0 2rem 2rem 2rem;
+    padding: 0 0 2rem 0;
```
chadoh: In the description, you said:

> Styling should have stayed exactly the same...

but this looks like a change. What prompted it?

apps/allocations/app/components/App/PayoutsTables.js
```diff
@@ -4,21 +4,25 @@ import styled from 'styled-components'
...
-  </div>
+  return (
+    <div>
+      <Text.Block size="large" weight="bold" style={{ marginBottom: '10px' }}>
+        Allocations
+        {' '}
+        <Badge.Info>{props.data.length}</Badge.Info>
+      </Text.Block>
+
+      <List {...props} />
```
chadoh: This now passes _all_ props to `List`, including `list` and `data`. We didn't pass it those two props, previously.

Why change the argument/prop signature on this component? Which linting rule complained about the old one?

apps/allocations/app/components/Panel/NewAccount.js
```diff
@@ -28,6 +27,7 @@ class NewAccount extends React.Component {
   createAccount = () => {
     const { description, } = this.state
     if (isStringEmpty(description)) {
+      // eslint-disable-next-line no-console
       return console.info(
```
chadoh: I think this is an error message; maybe we can use `console.error` here. I think our rules may be configured to allow that, so we'd be able to get rid of the `eslint-disable-next-line` declaration.

apps/dot-voting/app/components/Card/Vote.js
```diff
@@ -62,25 +66,19 @@ class Vote extends React.Component {
     }
 
     return (
-      <StyledCard>
-        <div onClick={this.handleVoteClick}>
+      <StyledCard onClick={this.handleVoteClick} css="cursor: pointer">
```
chadoh: Did a linting rule complain about the missing `cursor: pointer`??

chadoh: Also, clever avoidance of the a11y linter 😉 

(Really, though, it's fine for now. We'll revisit this app soon enough!)

apps/projects/app/context/index.js
```diff
@@ -1,2 +1,3 @@
-export { default as GithubProvider } from './Github'
+/* eslint-disable import/no-unused-modules */
+//export { default as GithubProvider } from './Github'
```
chadoh: Can we delete this comment?

apps/rewards/app/components/Content/Overview.js
```diff
@@ -211,7 +222,8 @@ const tableType = [
 const Overview = ({ tokens, rewards, convertRates, claims, newReward, openDetails }) => {
   const rewardsEmpty = rewards.length === 0
 
-  console.log('reward props: ', rewards)
+  // For testing purposes uncomment the following line:
+  // console.log('reward props: ', rewards)
```
chadoh: I doubt this comment is useful. We probably all know how to add console logging when we need it! 😄

apps/rewards/app/store/events.js
```diff
@@ -29,7 +29,7 @@ export const handleEvent = async (state, event, settings) => {
       nextState = await onRefreshRewards(nextState, returnValues)
       break
     default:
-      console.log('[Rewards reducer] unhandled event:', event, returnValues)
+      // console.log('[Rewards reducer] unhandled event:', event, returnValues)
```
chadoh: Can we delete this comment, and with it the whole `default` section of the switch statement?

apps/allocations/app/components/App/Payouts.js
rkzel: This was a fix.. although every fix is a change, right? So this one was to make Payout table to be aligned to Accounts, as per:
Screenshot from 2019-07-29 13-05-19
It's not perfectly aligned on the design, but then again what we have does not perfectly reflect the design - still, 0 looks much better than 2em:
Screenshot from 2019-07-29 14-48-28

apps/allocations/app/components/App/PayoutsTables.js
rkzel: Linting didn't complain here, but props were not correctly passed and PayoutTable wasn't working - listProps was undefined. I didn't investigate too much, and the most concise way of making it work again was to tweak what was already there. I think only "list" is superfluous - I can exclude it, so code is more explicit..?

apps/dot-voting/app/components/Card/Vote.js
rkzel: Linting complained about having onClick slapped onto div without the div having "role" and/or something more official. Normally divs are (as you told me many times! :) just simple structural elements. Div with onClick needs to be equipped with some rules. The way above: code is shorter and lint stopped complaining. Thx :)

apps/rewards/app/components/Content/Overview.js
rkzel: True :)

apps/rewards/app/store/events.js
rkzel: Sure, but in this case it should be removed for all apps probably... ? Elsewhere we have, for example, this:
```
  default:
    break
```
Bad style?

apps/allocations/app/components/App/PayoutsTables.js
chadoh: Looks like we are calling this with:

```jsx
<PayoutsTable
  data={sortedPayouts}
  tokens={tokens}
  executePayout={executePayout}
  list={PayoutsNarrow}
/>
```

Originally, the code you've changed here had this:

```jsx
const PayoutsTable = ({ list: List, data, ...listProps }) =>
```

Which means `listProps` would have been an object like:

```js
{ tokens, executePayout }
```

Going back to `PayoutsNarrow`, it looks like it expects the following args:

```js
const PayoutsNarrow = ({ executePayout, data, tokens }) => (
```

Since the previous code was only expanding `listProps`, this component would only have gotten passed `executePayout` and `tokens`. It probably broke because it wasn't being passed `data`.

Rather than changing this whole function, I think we only needed to start passing `data` to `List`:

```jsx
<List data={data} {...listProps} />
```

apps/rewards/app/store/events.js
chadoh: Given that I've previously advocated for removing seemingly-superfluous code from the script.js setup, and that code turned out to have a purpose after all, I suppose it's best to leave the `default` as-is. Also, this is a linting change, so it's probably best to keep a narrow focus and address only the linting issues.

Let's remove just this console line, then. No need to leave it present-but-commented.

rkzel: Sure!

apps/allocations/app/components/App/PayoutsTables.js
rkzel: Ok, I see the point. There is one more posibility: `data`, `tokens` and `executePayout` are needed all the way through, but `PayoutsTable` needs data only to check its length. So how about:
```
const PayoutsTable = ({ list: List, ...listProps }) => (
  <div>
(...)
      <Badge.Info>{listProps.data.length}</Badge.Info>
(...)
    <List {...listProps} />
  </div>
)
```
This is even cleaner, at least to me - would you agree?

apps/allocations/app/components/Panel/NewAccount.js
rkzel: Yes. But also.. this is actually very valid error message for the user. I made it show up in the panel, where it belongs.

apps/projects/app/context/index.js
rkzel: Certainly.

apps/allocations/app/components/App/PayoutsTables.js
chadoh: Yes, I like that option, too.

PR #1156, opened 2019-08-05
Aragon DS upgrade: addressbook

rkzel:
This requires aragon-ui 1.0 to run. Addressbook should be fully working, one thing for review discussion is scrolling of DataView row caused by context menu.


stellarmagnet:
1. The entity labels need to be right-aligned, being directly to the left of the context menu button
image

2. The empty state card should be vertically & horizontally centered
image

3. There is an issue with the "type" dropdown in New Entity panel -- after you open the dropdown, you can't close it by clicking on it again. Feature or a bug?

4. As mentioned in the PR, there is an issue with the scroll. 
image

Whenever I removed `<ScrollWrapper>`, it seemed to solve the issue. What was the purpose of that wrapper?
image

PR #1160, opened 2019-08-06
Upgrade/aragon api

Schwartz10:


apps/dot-voting/app/store/init.js
```diff
@@ -30,10 +29,12 @@ export const initStore = addressBookAddress => {
...
     },
-    [
-      of({ event: INITIALIZATION_TRIGGER }),
-      // handle address book events
-      addressBookApp.events(),
-    ]
+    {
+      externals: [
+        {
+          contract: addressBookApp
```
ottodevs: I understand that we want to keep the working state in this branch, but this is extra work and not very efficient, this shouldn't be required since we are getting rid of the address-book integration in another PR that will be merged in hours/days

apps/dot-voting/package.json
```diff
@@ -29,8 +29,6 @@
     "test": "TRUFFLE_TEST=true npm run ganache-cli:test"
   },
   "dependencies": {
-    "@aragon/api": "1.1.0",
-    "@aragon/api-react": "1.0.0-beta.2",
```
ottodevs: We not add the new versions also here?

package.json
```diff
@@ -75,6 +75,8 @@
     "test": "lerna run test --no-bail"
   },
   "dependencies": {
+    "@aragon/api": "2.0.0-beta.5",
```
ottodevs: We shouldn't need to add those to the root monorepo!

apps/rewards/.eslintrc.js
```diff
@@ -51,6 +52,7 @@ module.exports = {
       }
     ],
     'no-undef': 'error',
+    'no-console': ['warn', { allow: ['warn', 'error'] }],
```
ottodevs: I guess is ok to have these additions, but that goes against the good practice of keeping the PR simple, clean and focused, I let you know because is common that you point these kind of things on your reviews @chadoh ! This is not really consistent on your side.

Anyway I am totally ok on keeping these changes, just commenting to know your point about that, if you don't mind.

apps/projects/package.json
```diff
@@ -29,8 +29,6 @@
     "test": "TRUFFLE_TEST=true npm run ganache-cli:test"
   },
   "dependencies": {
-    "@aragon/api": "1.1.0",
-    "@aragon/api-react": "1.0.0-beta.2",
```
ottodevs: I think it would be a good chance to add the new versions of `api` and `api-react` to every package using them, that will ease the integrations once merged to dev

stellarmagnet:
@ottodevs do we still need this PR? If not, can we close it

PR #1194, opened 2019-08-29
Address book design check

e18r:
1. Change list layout so that it corresponds with the design.

Before: 
screenshot_2019-08-28_12-38-03_815348116

After: 
screenshot_2019-08-29_11-25-09_465675758

Mobile version:
screenshot_2019-08-29_13-00-04_780942154

2. Don't show Header if there are no contacts.

Before:
screenshot_2019-08-28_12-35-58_754864643

After:
screenshot_2019-08-29_13-22-46_331035060

### DIFF
--- a/apps/address-book/app/components/App/App.js
+++ b/apps/address-book/app/components/App/App.js
@@ -6,6 +6,7 @@ import { Button, Header, IconPlus, Main, SidePanel } from '@aragon/ui'
 import { IdentityProvider } from '../../../../../shared/identity'
 import Entities from './Entities'
 import NewEntity from '../Panel/NewEntity'
+import { Empty } from '../Card'
 
 const ASSETS_URL = './aragon-ui'
 
@@ -40,23 +41,26 @@ const App = () => {
 
   return (
     <Main assetsUrl={ASSETS_URL}>
-      { entries.length > 0 && (
+      { entries.length === 0 && <Empty action={newEntity} /> }
+      { entries.length > 0 &&
         <Header
           primary="Address Book"
           secondary={
             <Button mode="strong" icon={<IconPlus />} onClick={newEntity} label="New Entity" />
           }
         />
-      )}
+      }
       <IdentityProvider
         onResolve={handleResolveLocalIdentity}
         onShowLocalIdentityModal={handleShowLocalIdentityModal}
       >
-        <Entities
-          entities={entries}
-          onNewEntity={newEntity}
-          onRemoveEntity={removeEntity}
-        />
+        { entries.length > 0 &&
+          <Entities
+            entities={entries}
+            onNewEntity={newEntity}
+            onRemoveEntity={removeEntity}
+          />
+        }
         <SidePanel onClose={closePanel} opened={panelVisible} title="New entity">
           <NewEntity onCreateEntity={createEntity} />
         </SidePanel>
--- a/apps/address-book/app/components/App/App.js
+++ b/apps/address-book/app/components/App/App.js
@@ -40,12 +40,14 @@ const App = () => {
 
   return (
     <Main assetsUrl={ASSETS_URL}>
-      <Header
-        primary="Address Book"
-        secondary={
-          <Button mode="strong" icon={<IconPlus />} onClick={newEntity} label="New Entity" />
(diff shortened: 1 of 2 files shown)


e18r:
Mobile version working

e18r:
Don't show Header if there are no contacts: https://drive.google.com/open?id=1TCrJfMm9a1oSKWuonyrqfvPPzW43R7d1

apps/address-book/app/components/App/App.js
```diff
@@ -40,12 +40,14 @@ const App = () => {
 
   return (
     <Main assetsUrl={ASSETS_URL}>
-      <Header
-        primary="Address Book"
-        secondary={
-          <Button mode="strong" icon={<IconPlus />} onClick={newEntity} label="New Entity" />
-        }
-      />
+      { entries.length > 0 && (
```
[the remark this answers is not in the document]
e18r: Yes. That makes a lot of sense.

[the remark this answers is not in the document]
e18r: OK. I consolidated the `if no entries` logic into `App.js`. However, I decided not to do it in two different `return` blocks as you propose, because when there are no entries you still need the `SidePanel`, so the code ended up being redundant. Instead, I made three "inline" checks.

In other words, I'm sacrificing clarity for succinctness. Let me know what you think.

apps/address-book/app/components/App/App.js
```diff
@@ -40,21 +41,26 @@ const App = () => {
...
   return (
     <Main assetsUrl={ASSETS_URL}>
-      <Header
-        primary="Address Book"
-        secondary={
-          <Button mode="strong" icon={<IconPlus />} onClick={newEntity} label="New Entity" />
-        }
-      />
+      { entries.length === 0 && <Empty action={newEntity} /> }
+      { entries.length > 0 &&
```
chadoh: Do we need to render `<Main>` and `<IdentityProvider>`, no matter what? 

If not, an early return would read more clearly, in my opinion:

```js
if (!entries.length) return <Empty action={newEntity} />

return (
  <Main>
    ...
```

If we do need to render them, then maybe we could do something like this:

```js
// at the top level of the file
const Wrap = ({ children }) => {
  const { api } = useAragonApi()

  const handleResolveLocalIdentity = address =>
    api.resolveAddressIdentity(address).toPromise()

  const handleShowLocalIdentityModal = address =>
     api.requestAddressIdentityModification(address).toPromise()

  return (
    <Main  assetsUrl={ASSETS_URL}>
      <IdentityProvider
      >
        {children}
      </IdentytProvider>
    </Main>
  )
}

// in App component
if (!entries.length) return <Wrap><Empty action={newEntity} /></Wrap>

return (
  <Wrap>
    <Header ... />
    <Entities ... />
    <SidePanel ... />
  </Wrap>
)

chadoh: Oh, I just saw your comment about always needing `SidePanel`. It's easy to rework my suggestion to put `SidePanel` in `Wrap`.

e18r: This is my code so far:

    const Wrap = ({ children }) => (
    <Main assetsUrl={ASSETS_URL}>
      { children }
      <IdentityProvider
      >
        { entries.length > 0 &&
          <Entities
            entities={entries}
            onNewEntity={newEntity}
            onRemoveEntity={removeEntity}
          />
        }
        <SidePanel onClose={closePanel} opened={panelVisible} title="New entity">
          <NewEntity onCreateEntity={createEntity} />
        </SidePanel>
      </IdentityProvider>
    </Main>
    )

    if (!entries.length) return (
    <Wrap><Empty action={newEntity} /></Wrap>
    )

    return (
    <Wrap>
      <Header
        primary="Address Book"
        secondary={
          <Button mode="strong" icon={<IconPlus />} onClick={newEntity} label="New Entity" />
        }
      />
    </Wrap>
    )

But I'm not sure how to get rid of that last inline check. Also, I'm not sure if this solution really contributes to clarity.

e18r:
I'm closing this and will create a new PR from a branch of AutarkLabs' repo instead of my own.

PR #1195, opened 2019-08-29
Enhance the mine cli script to be usable with Truffle

topocount:
As a small exercise I rewrote the mine macro script (for testing and other useful little stuff) to be usable with `truffle exec`

### DIFF
--- a/apps/shared/test-helpers/mine.js
+++ b/apps/shared/test-helpers/mine.js
@@ -1,30 +1,41 @@
-/// uncomment the Below block to use on aragon devchain without truffle
-
-//const Web3 = require("web3");
-//const PrivateKeyProvider = require("truffle-privatekey-provider");
-//const provider = new PrivateKeyProvider(
-//    "A8A54B2D8197BC0B19BB8A084031BE71835580A01E70A45A13BABD16C9BC1563",
-//    "http://localhost:8545"
-//  );
-//const web3 = new Web3(provider);
-
 const mineBlock = require("./mineBlock")
 const getBlock = require("./blockNumber")
 
-module.exports = async (cb) => {
-    const mineBlockFn = mineBlock(web3)
-    const getBlockFn = getBlock(web3)
+const getWeb3 = () => {
+    const Web3 = require("web3");
+    const PrivateKeyProvider = require("truffle-privatekey-provider");
+    const provider = new PrivateKeyProvider(
+        "A8A54B2D8197BC0B19BB8A084031BE71835580A01E70A45A13BABD16C9BC1563",
+        "http://localhost:8545"
+    );
+    return new Web3(provider);
+}
+
+// truffle script: can be run with `truffle exec`
+const mine = async (cb) => {
+    const root = this
+    let activeWeb3
+    let argIdx
+    if (typeof cb !== 'function') {
+        activeWeb3 = getWeb3()
+        argIdx = 2
+    } else {
+        activeWeb3 = web3
+        argIdx = 4
+    }
+    const mineBlockFn = mineBlock(activeWeb3)
+    const getBlockFn = getBlock(activeWeb3)
 
     const mineBlocks = async (blocks) => {
-        for (var i = 0; i < blocks; i++) {
+        const blockArr = new Array(Number(blocks)).fill(0)
+        blockArr.forEach(async () => {
             try {
                 await mineBlockFn()
             }
             catch(e) {
                 console.error('error: ',e)
             }
-        }
-        console.log('\nblock',await getBlockFn(), 'mined')
+        })
     }
 
     const mineToBlock = async (blockNumber) => {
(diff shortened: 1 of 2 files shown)


apps/shared/test-helpers/mine.js
```diff
@@ -1,43 +1,49 @@
-const Web3 = require("web3");
-const PrivateKeyProvider = require("truffle-privatekey-provider");
-const provider = new PrivateKeyProvider(
-    "A8A54B2D8197BC0B19BB8A084031BE71835580A01E70A45A13BABD16C9BC1563",
-    "http://localhost:8545"
-  );
-const web3 = new Web3(provider);
+/// uncomment the Below block to use on aragon devchain without truffle
```
chadoh: Did you mean to put three `///` at the beginning of this line?

apps/shared/test-helpers/mine.js
```diff
@@ -1,43 +1,49 @@
...
-        try {
-            await mineBlock()
-        }
-        catch(e) {
-            console.error('error: ',e)
+const mineBlock = require("./mineBlock")
+const getBlock = require("./blockNumber")
+
+module.exports = async (cb) => {
+    const mineBlockFn = mineBlock(web3)
```
chadoh: The indentation in this file differs from our others. Why is that?

More importantly, where does the `web3` variable come from, when the above code is commented out? If it's a global variable that's available thanks to Truffle, then instead of requiring someone to switch the commented code above, can we check if `web3` is defined and use the new style if so, and the old style if not?

apps/shared/test-helpers/mine.js
```diff
@@ -1,43 +1,49 @@
...
+    const getBlockFn = getBlock(web3)
+
+    const mineBlocks = async (blocks) => {
+        for (var i = 0; i < blocks; i++) {
+            try {
+                await mineBlockFn()
```
chadoh: We never use the `i` variable. I expected this `mineBlockFn` to receive a given block as an argument. How does it work, given that it receives no arguments? Does it do the same thing every single time?

Is there a way to use `Promise.all` so we don't have to await one `mineBlockFn` to complete in order to start the next one?

I see this isn't new code, but while we're here, can we use `blocks.forEach(block => {` instead of the `for` statement?

apps/shared/test-helpers/mine.js
```diff
@@ -1,43 +1,49 @@
...
-    console.log('\nblock',await getBlock(), 'mined')
-    process.exit()
-}
 
-if (!Number(process.argv[2])) {
-    mineToBlock(process.argv[3])
-}
-else {
-    mineBlocks(process.argv[2])
+    if (!Number(process.argv[4])) {
```
chadoh: `process.argv[4]` might be `''`, empty string? Is that correct?

ottodevs:
Hey @topocount did you have a chance to take a look to the related changes I made during Rewards internal audit?
https://github.com/AutarkLabs/open-enterprise/pull/1155/commits/75bcb44209ef52e38a94aa67aa09b9672097debf

Maybe this is what you are looking for, remember we are trying to get rid of test-helpers package

topocount:
> Hey @topocount did you have a chance to take a look to the related changes I made during Rewards internal audit?
> [75bcb44](https://github.com/AutarkLabs/open-enterprise/commit/75bcb44209ef52e38a94aa67aa09b9672097debf)
> 
> Maybe this is what you are looking for, remember we are trying to get rid of `@tps/test-helpers` package

We actually need this script for E2E testing Otto

apps/shared/test-helpers/mine.js
topocount: I can functionalize this, but mineBlockFn is a very primitive rpc command that accepts no arguments and just commands the devchain to mine a single block. The whole purpose of this script is to be able to run commands like:
`npm run mine to x` as in mine to block x and `npm run mine x` as in mine x number of blocks

apps/shared/test-helpers/mine.js
topocount: it might otherwise be a string, in which case `process.argv[5]` contains the block number we want to mine to

chadoh: Do you think it's best to coerce to `Number`, or would coercing to `Boolean` be a better way to explain what's happening?

apps/shared/test-helpers/mine.js
chadoh: Hmm, gotcha. So we could use `blocks.forEach(() =>` here, since we don't actually care about any of the info associated with a particular block.

apps/shared/test-helpers/mine.js
topocount: Yes because it's a metacomment about the commented code below

apps/shared/test-helpers/mine.js
topocount: The indendation is different from our app files, but it appears that the helper functions in the library are all node scripts with 4-spaced indentations

apps/shared/test-helpers/mine.js
topocount: coercing to a bool wouldn't work because if this is not a number, it'll still be a string of some length, since we'd be expecting  the `to` string in lieu of a non-numerical input

PR #1154, opened 2019-08-05
Perform cleanup and performance changes from internal audit on DotVoting app 

ottodevs:
- Includes boost to coverage config
- Includes changes to involved `@tps/test-helpers` contracts
- Progress and rest of needed changes defined into the notion report for this internal audit
- This PR targets `mainnet-audit-projects-3`, being rebased from it and forked from `dot-voting-changes`

Before:
image

After:
image

**NOTE: The high number of removed lines is related to cleaning `@tps/test-helpers` folder at commit 6066351, I deleted old/unused contract files, we only need to configure remaining allocations test-helpers imports to complete the cleaning*

PR #1196, opened 2019-08-29
Dot Voting: update contracts & related code

### LINKED ISSUE #1020: Web3 integration for revised Dot Voting contract
After the Dot Voting contract has been internally audited, and the new frontend components developed (if any), we should upgrade the web3 integration.

Note: should sync up on our Discussions strategy before this is done, as Dot Voting will be the app where we integrate Discussions

This includes adding the forwarder-api triggers to dot-voting.

chadoh:
Fixes #1020 

This is a minimal version of the contract-related changes from #1154, which we can get merged to `dev` quickly.

Foundational changes
--------------------

* add missing packages to package.json
* update solium (now called ethlint) configuration
* add solcover for test-coverage reporting (this required adding `.gitkeep` to the migration directory)
* remove unused files

The speed of coverage for dot-voting is BLAZING fast now (80s vs >400s) 🎉

Downside is we still need @tps/test-helpers package. Hopefully definitive solution when the 3 contracts (ADynamicForwarder, ScriptHelpers, IForwarder) get merged into `@aragon/os`.

The main event
--------------

* update DotVoting.sol from internal & external audit feedback
* update script.js to match new contract
* update UI to match new script.js


apps/dot-voting/package.json
```diff
@@ -63,15 +65,24 @@
     "eslint-plugin-promise": "^4.1.1",
     "eslint-plugin-react": "^7.13.0",
     "eslint-plugin-standard": "^4.0.0",
+    "eth-gas-reporter": "0.1.12",
+    "ethereumjs-abi": "0.6.5",
+    "ethlint": "^1.2.4",
```
topocount: @ottodevs Quazia and I ran into problems with Ethlint in CI for Discussions, and it looks like it's also causing problems in this branch. Can we delay these updates for another time and continue with solium for a bit longer? Also, is it okay if we continue to use the same soliumrc configuration we've been using in dev? I'd prefer to discuss style changes, especially of this magnitude

apps/dot-voting/app/store/events.js
```diff
@@ -28,14 +24,8 @@ export const handleEvent = async (state, event, settings) => {
...
-  case 'EntryAdded':
-    nextEntries = await onEntryAdded({ entries, addressBook }, returnValues)
-    nextState.entries = nextEntries
-    break
-  case 'EntryRemoved':
-    nextEntries = await onEntryRemoved({ entries, addressBook }, returnValues)
-    nextState.entries = nextEntries
-    break
+  case 'UpdateQuorum':
+  case 'UpdateMinimumSupport':
```
ottodevs: These 2 events are not handled right now, do they involve changes in the state?

apps/dot-voting/app/store/init.js
```diff
@@ -1,9 +1,6 @@
-import AddressBookJSON from '../../../shared/json-abis/address-book.json'
 import { app, handleEvent } from './'
 
-export const initStore = addressBookAddress => {
-  const addressBookApp = app.external(addressBookAddress, AddressBookJSON.abi)
-
+export const initStore = () => {
   const initialState = {
     votes: [],
     entries: [],
```
ottodevs: We might safely remove entries here

apps/dot-voting/app/App.js
```diff
@@ -88,7 +88,7 @@ const App = () => {
     votes = [],
     entries = [],
     voteTime = 0,
-    minParticipationPct = 0,
+    globalMinQuorum = 0,
     pctBase = 0,
   } = appState
```
ottodevs: I am not sure if the function below makes sense now that AddressBook integration is removed, we might want clean out in case this is not needed

apps/dot-voting/app/App.js
```diff
@@ -110,12 +110,12 @@ const App = () => {
...
       quorum: safeDiv(vote.data.minAcceptQuorum, pctBase),
       quorumProgress: getQuorumProgress(vote.data),
-      minParticipationPct: minParticipationPct / 10 ** 16,
+      globalMinQuorum: globalMinQuorum / 10 ** 16,
       description: vote.data.metadata,
       totalSupport: getTotalSupport(vote.data),
       type: vote.data.type,
     }
-  }, [ voteTime, getAddressLabel, pctBase, minParticipationPct ])
+  }, [ voteTime, getAddressLabel, pctBase, globalMinQuorum ])
```
ottodevs: getAddressLabel also seems part of the AddressBook integration, am I right?

apps/dot-voting/package.json
```diff
@@ -81,7 +92,7 @@
   "keywords": [],
   "license": "GPL-3.0",
   "lint-staged": {
-    "*.js": [
+    "app/**/*.js": [
```
ottodevs: Might it be useful to also lint other `js` files outside the `app` folder? like for example the test files, linter configs and truffle-config, just asking, I would like to understand better the motivation for this linter scope limitation :)

apps/planning-suite-kit/contracts/PlanningKit.sol
```diff
@@ -247,7 +247,7 @@ contract PlanningKit is KitBase {
         address root = msg.sender;
         addressBook.initialize();
         projects.initialize(registry, vault, token);
-        dotVoting.initialize(addressBook, token, 50 * PCT256, 0, 1 minutes);
+        dotVoting.initialize(token, 50 * PCT256, 0, 1 minutes);
```
ottodevs: This seems a really nice move!, but I think it is a good idea to also perform the same change to our production template at `kits/planning-suite/contracts/`

apps/dot-voting/app/store/events.js
chadoh: It's just a hook for us in the future. Right now both of these will "fall through" to the `default` case below.

apps/dot-voting/package.json
chadoh: I did not intentionally add this; I think it came from #1154. I'll remove it!

chadoh: I think this lint setting is why some of those unused variables weren't detected by the linter! Great find, thank you!

apps/planning-suite-kit/contracts/PlanningKit.sol
chadoh: @Quazia you want to handle that?

apps/dot-voting/app/store/events.js
chadoh: I could be convinced to take it out, but I like that this gives more visibility to what events are available in the contract.

apps/planning-suite-kit/contracts/PlanningKit.sol
chadoh: It was an easy change to make; never mind. See most recent commit; I changed it.

apps/dot-voting/app/store/events.js
topocount: I'm in favor of keeping them. It's like a comment-less todo

apps/dot-voting/contracts/test/mocks/DotVotingMock.sol
```diff
@@ -5,7 +5,7 @@ import "../../DotVoting.sol";
 
 contract DotVotingMock is DotVoting {
     // _isValuePct public wrapper
-    function isValuePct(uint256 _value, uint256 _total, uint256 _pct) external pure returns (bool) {
```
topocount: Why remove this? It shouldn't throw a compiler error since the internal function is also pure.

topocount:
@ottog I've approved this, but definitely request more changes if you notice anything else

apps/dot-voting/contracts/test/mocks/DotVotingMock.sol
chadoh: @Quazia this is a question for you

chadoh: Or maybe this is a question for you, @ottodevs 😄

PR #1199, opened 2019-08-30
Setup continuous deploy for Open Enterprise Apps

topocount:
This PR will enable us to publish dev builds of apps to rinkeby upon merging to our dev branch. Once we have a stable dev dao set up, the [deploy](https://github.com/AutarkLabs/open-enterprise/blob/20d347ccd03cbcf17b45416ae796f76ccfbe1022/shared/deployments/publish-cd.sh) script can be modified to upgrade all the apps in our dao. 

In the short term (for the next couple weeks) the dev dao should be handled manually until our workflow stabilizes and the deployment set up in this PR has been shown to be reliable


.travis.yml
```diff
@@ -15,7 +16,7 @@ node_js: '10.15' # equivalent to node 10.15.x
 cache:
   directories:
     - node_modules
-#    - $HOME/.npm
+    - /home/travis/.nvm/versions/node/v10.15.3/lib/node_modules/
```
chadoh: This looks brittle! Is there a more robust way to do it, using something closer to that `$HOME` variable?

.travis.yml
```diff
@@ -63,10 +64,27 @@ jobs:
...
+        on:
+          branch: dev
 
 notifications:
   email: false
+
+# Only build pushes to dev. All pull requests still build
+branches:
+  only:
+  - dev
```
chadoh: We've been having double-build problems on our PRs lately, with Travis configured to run on _every branch push_ and on _every push to a PR_. This means that pushes to branches associated with PRs build twice. Will this configuration fix that problem?

apps/address-book/arapp.json
```diff
@@ -23,6 +23,12 @@
       "wsRPC": "wss://rinkeby.eth.aragon.network/ws",
       "network": "rinkeby"
     },
+    "continuous-deployment": {
```
chadoh: Given how similar this is to the `rinkeby` setup above, can you explain more about why we need both?

apps/allocations/arapp.json
```diff
@@ -22,15 +22,21 @@
...
     },
-    "staging": {
+    "continuous-deployment": {
       "registry": "0x98df287b6c145399aaa709692c8d308357bc085d",
-      "appName": "allocations-staging.open.aragonpm.eth",
+      "appName": "alllocations-cd.open.aragonpm.eth",
+      "wsRPC": "wss://rinkeby.eth.aragon.network/ws",
+      "network": "rinkeby"
+    },
+    "rinkeby": {
```
chadoh: The order of these mismatches the order in Address Book's `arapp`. I'd prefer we keep them consistent.

apps/allocations/package.json
```diff
@@ -15,6 +15,7 @@
     "lint": "solium --dir ./contracts",
     "precommit": "lint-staged",
     "prepublishOnly": "truffle compile --all",
+    "publish:cd": "aragon apm publish major --files dist/ --environment continuous-deployment --apm.ipfs.rpc https://ipfs.autark.xyz:5001 --ipfs-check false",
     "publish:http": "npm run build:script && aragon apm publish major --http localhost:2222 --http-served-from ./dist",
```
chadoh: Same thing about ordering here

apps/address-book/arapp.json
ottodevs: > Given how similar this is to the `rinkeby` setup above, can you explain more about why we need both?

Note the app name is different for this environment, I guess that makes the need to separate it from rinkeby, I find that a good idea atm

.travis.yml
ottodevs: Agree with @chadoh is there a better "timeproof" way? What kind of issues did you find with the previous `$HOME` variable @topocount ?

.travis.yml
ottodevs: I recently enabled the pushes build again for some builds I wanted to test, related to `mainnet-audit-*` branches, since those were not directly targeting `dev`. I disabled it again, I think it is easy enough to enable-disable it on demand for things like that, do we need to have a static setting for this config to work @topocount ?

topocount: Yes this will resolve this and enable push-builds to only execute on merges to dev (as the comment on line 87 indicates). We'll be able to turn them on and not have them run concurrently with PR builds.

Also @chadoh even though it is resource-intensive, the double-builds provide some additional value in that the PR build has the parent codebase merged into the child branch. This is why PR builds won't execute if there are merge conflicts. The push-builds though just take the branch as-is and test it though. Executing both of these saves some time in terms of testing and debugging, but if the developer is already testing their code locally, the push-builds are mostly unnecessary

.travis.yml
topocount: I'll address each of these changes separately:

# Removing the `$HOME/.npm` line
This was commented out anyway, but I read in the TravisCI docs that [`.npm` is now cached by default](https://docs.travis-ci.com/user/caching/#npm-cache) so I figured we should remove it since this is officially documented

# caching global node_modules
This is definitely not pretty. I thought it'd be a good idea beause it would speed things up in terms of installing the aragonCLI globally for use with deployments. However, I've observed that it actually slows down the cache-warming step substantially (adding 3-4 minutes) so I'll go ahead and remove this line

apps/address-book/arapp.json
topocount: The concept of an `environment` exists separately from a `network` for this very purpose. You can have multiple namespaces (ie `appName`s) of a single app on a given network. For example, we have our public `rinkeby` deploy which is what is available in the App center, and we also have `staging`, which exists as our public demo-dao namespace. The `continuous-deployment` namesapace serves its own distinct purpose of giving us a distinct publishing space that doiesn't touch our public-facing deployments, while also existing on rinkeby and utilizing the same registry

chadoh: I was picturing this as our eventual goal:

* merges to `dev` automatically deploy to our `staging` environment, and anyone using our Rinkeby apps can get those changes immediately
* merges to `master` automatically deploy to our `production` environment; anyone using our mainnet apps can get those

I'm not certain we need the flexibility+overhead (they go together, in this case) of having two separate Rinkeby namespaces. Do you see having both as temporary?

topocount: I think our automated testing apparatus would have to be much more thorough to reach this point

apps/address-book/arapp.json
```diff
@@ -17,15 +17,21 @@
       "network": "rpc",
       "appName": "address-book.aragonpm.eth"
     },
-    "rinkeby": {
+    "staging": {
       "registry": "0x98df287b6c145399aaa709692c8d308357bc085d",
-      "appName": "tps-address-book.open.aragonpm.eth",
```
ottodevs: I hope we track this since we have the apps published on rinkeby environment, why was it needed to wipe this environment that is currently being used?

PR #1215, opened 2019-08-30
Address book custom type

e18r:
Added custom type option for contacts

Creating a new contact with custom type:
screenshot_2019-08-30_16-50-33_983489690

Listing contacts with custom types: 
screenshot_2019-08-30_16-50-10_401762476

*Note*: the design is vague on how to display the custom type badge.

### DIFF
--- a/apps/address-book/app/components/App/Entities.js
+++ b/apps/address-book/app/components/App/Entities.js
@@ -18,7 +18,7 @@ import { Empty } from '../Card'
 const ENTITY_TYPES = [
   { name: 'Individual', fg: '#76A4E5', bg: '#76A4E533' },
   { name: 'Organization', fg: '#F78308', bg: '#F7830833' },
-  { name: 'Project', fg: '#B30FB3', bg: '#B30FB333' },
+  { name: '_custom', fg: '#B30FB3', bg: '#B30FB333' },
 ]
 
 const entitiesSort = (a, b) => a.data.name.toUpperCase() > b.data.name.toUpperCase() ? 1 : -1
@@ -41,7 +41,7 @@ const Entities = ({ entities, onNewEntity, onRemoveEntity }) => {
         }
 
         renderEntry={([ name, entryAddress, entryType ]) => {
-          const typeRow = ENTITY_TYPES.filter(row => row.name === entryType)[0]
+          const typeRow = ENTITY_TYPES.filter(row => row.name === entryType || row.name === '_custom')[0]
           const values = [
             // eslint-disable-next-line react/jsx-key
             <EntityWrapper>
@@ -64,7 +64,7 @@ const Entities = ({ entities, onNewEntity, onRemoveEntity }) => {
               background={typeRow.bg}
               css="text-align: right"
             >
-              {typeRow.name}
+              {entryType}
             </Badge>
           ]
           return values
--- a/apps/address-book/app/components/Panel/NewEntity.js
+++ b/apps/address-book/app/components/Panel/NewEntity.js
@@ -4,11 +4,13 @@ import { Form, FormField } from '../Form'
 import { DropDown, TextInput } from '@aragon/ui'
 import web3Utils from 'web3-utils'
 
-const ENTITY_TYPES = [ 'Individual', 'Organization', 'Project' ]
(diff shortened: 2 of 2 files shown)


e18r:
@jayalaves thank you for your comments. Please check PR https://github.com/AutarkLabs/open-enterprise/pull/1214, where I put the AddressBadge within a separate column. As for the other two remarks, I don't quite follow, but maybe they belong to that other PR also.

e18r:
@chadoh @jayalaves This PR is ready for review. Custom type badges will have a single color for now, as discussed in keybase.

apps/address-book/app/components/App/Entities.js
```diff
@@ -41,7 +41,7 @@ const Entities = ({ entities, onNewEntity, onRemoveEntity }) => {
         }
 
         renderEntry={([ name, entryAddress, entryType ]) => {
-          const typeRow = ENTITY_TYPES.filter(row => row.name === entryType)[0]
+          const typeRow = ENTITY_TYPES.filter(row => row.name === entryType || row.name === '_custom')[0]
```
chadoh: Some thoughts:

* Strange that we do a `filter` and then `[0]` instead of `find`

* This find/filter logic feels strange to me. Like, "Where row name is the type, or just grab the single `_custom` one." I would expect us to save a whole array of custom types, with proper names. Then we merge the saved entity types with the hard-coded ones for `Individual` and `Organization`, and continue to find by name.

* I think it would simplify the code to use an object for ENTITY_TYPES, rather than an array.
   ```js
   { Individual: { bg: ..., fg: ... },
     Organization: { bg: ..., fg: ... },
     'Custom Label 1': { bg: ..., fg: ... },
     'Custom Label 2': { bg: ..., fg: ... },
     ...
   }
   ```

  Now that we're persisting custom types, rather than just using a hard-coded list, I would expect this object to come from `appState`.

apps/address-book/app/components/Panel/NewEntity.js
```diff
@@ -4,11 +4,13 @@ import { Form, FormField } from '../Form'
 import { DropDown, TextInput } from '@aragon/ui'
 import web3Utils from 'web3-utils'
 
-const ENTITY_TYPES = [ 'Individual', 'Organization', 'Project' ]
+const ENTITY_TYPES = [ 'Individual', 'Organization', 'Custom type…' ]
```
chadoh: Implementing custom types gives us a good chance to get rid of the duplicated logic we had between these two separate `ENTITY_TYPES` constants. If `ENTITY_TYPES` comes from `useAragonApi()` -> `appState`, it accomplishes that for us.

apps/address-book/app/components/Panel/NewEntity.js
```diff
@@ -4,11 +4,13 @@ import { Form, FormField } from '../Form'
...
 const INITIAL_STATE = {
   name: '',
   address: '',
   type: 'Individual',
+  isCustomType: false,
```
chadoh: If types are coming from appState, we don't need special handling here for `isCustomType`

apps/address-book/app/components/Panel/NewEntity.js
```diff
@@ -28,29 +30,44 @@ class NewEntity extends React.Component {
...
 
     if (Object.keys(error).length) {
       this.setState({ error: error })
     } else {
       this.setState(INITIAL_STATE)
-      this.props.onCreateEntity({ name, address, type })
+      this.props.onCreateEntity({
+        name,
+        address,
+        type: isCustomType ? customTypeName : type
```
chadoh: Rather than saving this as a simple `type` string on the entity, we probably want a separate api call for `createType`, where we pass it the name as well as the `bg` and `fg` colors. Then, when the result of that save comes back, we can show the new type in the drop-down.

However, implementing this requires not just code changes, but a UX/design discussion. I think we should close this PR for now, since the final version will probably look quite different, and circle back to this improvement once all of our apps are on mainnet.

e18r:
Hi @chad, thank you so much for your valuable feedback. I agree that we should close this PR for now and circle back later.

chadoh:
Sounds good! Thanks.

(Wrong Chad, btw 😄)

PR #1217, opened 2019-09-02
Icon for the delete entry in the context menu

e18r:
Before: 
screenshot_2019-08-28_12-41-39_657013967

After:
screenshot_2019-09-02_16-31-45_146466233


e18r:
Thank you, merged!

PR #1218, opened 2019-09-03
Fix/discussions

Schwartz10:


apps/discussions/app/App.js
```diff
@@ -1,41 +1,18 @@
 import React from 'react'
-import styled from 'styled-components'
-import { EmptyStateCard, GU, Header, Main, breakpoint } from '@aragon/ui'
-import nothingToSeeHere from './empty-state.png'
-
-const illustration = <img src={nothingToSeeHere} alt="" height="160" />
 
 function App() {
   return (
-    <Main>
```
ottodevs: If we get rid of the Main component, the app will not look like as an aragon app (css, fonts, etc)

apps/discussions/app/modules/DiscussionsApi.js
```diff
@@ -38,8 +38,11 @@ class Discussions {
   _pastEvents = () =>
     new Promise(resolve =>
       this.contract.pastEvents().subscribe(events => {
-        this.lastEventBlock = events[events.length - 1].blockNumber
-        resolve(events)
+        if (events.length > 0) {
+          this.lastEventBlock = events[events.length - 1].blockNumber
+          return resolve(events)
+        }
+        resolve([])
```
ottodevs: The shape result of the function seems different depending on event.length, one is returning the resolve result, the other is just resolving without a `return` is this the expected behavior?

apps/discussions/package.json
```diff
@@ -1,19 +1,20 @@
 {
   "name": "@tps/apps-discussions",
-  "version": "1.0.0",
+  "version": "0.0.1",
   "description": "",
   "dependencies": {
     "@aragon/api": "latest",
```
ottodevs: Since we are moving some dependencies around, it should be a good chance to fix this, that could break something if we don't control which version are we using for `@aragon/api` and `@aragon/api-react`
Shouldn't be better to use a more specific version to have more control over it?

ottodevs:
Can we close this @Schwartz10 ?

PR #1221, opened 2019-09-05
DV UI final fixes

rkzel:
Inside:
- app badges are now as per actual design,
- decimals needed to be brought back for Allocations vote balance to be displayed correctly,
- vote details tab closes after the vote has been executed,
- minor styling fixes.


stellarmagnet:
The detail tab should not close

apps/dot-voting/app/components/VoteDetails/index.js
```diff
@@ -54,8 +61,28 @@ const VoteDetails = ({ app, vote, userAccount, onVote }) => {
...
+          tokenContract.balanceOfAt(userAccount, snapshotBlock),
+          tokenContract.decimals(),
+          tokenContract.symbol()
+        )
+          .pipe(first())
+          .subscribe(([ balance, decimals, symbol ]) => {
+            setTokenData({
+              userBalance: balance,
+              decimals: decimals,
+              voteTokenSymbol: symbol,
```
chadoh: I don't think we ever use this value, `voteTokenSymbol`. If that's true, we can stop importing `tokenSymbolAbi` altogether.

apps/dot-voting/app/components/VoteDetails/index.js
```diff
@@ -54,8 +61,28 @@ const VoteDetails = ({ app, vote, userAccount, onVote }) => {
...
+      if (tokenContract && userAccount) {
+        combineLatest(
+          tokenContract.balanceOfAt(userAccount, snapshotBlock),
+          tokenContract.decimals(),
+          tokenContract.symbol()
+        )
+          .pipe(first())
+          .subscribe(([ balance, decimals, symbol ]) => {
+            setTokenData({
+              userBalance: balance,
```
chadoh: I don't see this, either. The function is called `loadUserBalance`, as if it's the most important attribute returned, but then we only ever use `tokenData.decimals`. I think we can simplify all of this logic a bit, and store `decimals` directly in state, rather than storing an object for `tokenData`.

apps/dot-voting/app/Decisions.js
```diff
@@ -155,6 +166,7 @@ const Decisions = ({ decorateVote }) => {
           app={app}
           vote={currentVote}
           userAccount={connectedAccount}
+          tokenContract={tokenContract}
```
chadoh: Looks like this is the only place that `tokenContract` is used. Given that `VoteDetails` can also pull `tokenAddress` out of `appState` and do all of this logic itself, can we move it all there?

apps/dot-voting/app/Decisions.js
```diff
@@ -107,11 +110,19 @@ const Decisions = ({ decorateVote }) => {
   const {
     votes,
     voteTime,
+    tokenAddress = '',
   } = appState
 
   const { layoutName } = useLayout()
   const theme = useTheme()
 
+  const tokenAbi = [].concat(tokenBalanceOfAbi, tokenDecimalsAbi, tokenSymbolAbi)
```
chadoh: This is a constant that is not reliant on component state/props. Let's move it outside of the component definition, right under the `import tokenDecimalsAbi from './abi/token-decimals.json'` part. That is, if we even need it anymore, given that it seems we only need `tokenDecimalsAbi`.

apps/dot-voting/app/components/VotingCardGroup.js
```diff
@@ -58,4 +63,26 @@ VotingCardGroup.propTypes = {
...
+  display: inline-flex;
+  white-space: nowrap;
+  color: rgb(109, 128, 136);
+  padding-top: 4px;
+  letter-spacing: -0.5px;
+  /* stylelint-disable-next-line property-no-vendor-prefix */
+  -webkit-box-pack: center;
+  justify-content: center;
+  /* stylelint-disable-next-line property-no-vendor-prefix */
+  -webkit-box-align: center;
```
chadoh: Do we need these two stylelint-ignored lines? Are they not auto-inserted for `justify-content: center`?

apps/dot-voting/app/components/VotingOption.js
```diff
@@ -18,13 +18,27 @@ const VotingOption = ({ valueSpring, label, percentage, color, threshold, userVo
...
+                border-radius: 24px;
+                font-size: 12px;
+                line-height: 1.5;
+                text-transform: uppercase;
+                font-weight: 600;
+                user-select: none;
+                color: ${theme.info};
+                background: ${theme.infoSurface.alpha(0.08)};
+                margin-left: 10px;
+                align-items: center;
```
chadoh: These styles seem pretty similar to `DiskTag` above. Is there a common component we could pull out, that both of these could then start from? It's a lot of styles that seem like maybe they should have a name.

apps/dot-voting/app/components/VotingCardGroup.js
```diff
@@ -58,4 +63,26 @@ VotingCardGroup.propTypes = {
...
+  align-items: center;
+  width: 18px;
+  height: 18px;
+  font-size: 12px;
+  font-weight: 600;
+  line-height: 20px;
+  background: rgb(220, 234, 239);
```
chadoh: Can this and `color` both use a theme style instead?

apps/dot-voting/app/components/VoteDetails/index.js
```diff
@@ -121,8 +163,10 @@ const VoteDetails = ({ app, vote, userAccount, onVote }) => {
 VoteDetails.propTypes = {
   app: PropTypes.object.isRequired,
   userAccount: PropTypes.string.isRequired,
+  tokenContract: PropTypes.object.isRequired,
   vote: PropTypes.object.isRequired,
   onVote: PropTypes.func.isRequired,
+  setCurrentVoteId: PropTypes.func.isRequired,
```
chadoh: We did not start passing this in, or using it anywhere.

apps/dot-voting/app/Decisions.js
```diff
@@ -84,18 +87,18 @@ const useFilterVotes = (votes, voteTime) => {
...
     voteOutcomeFilter: outcomeFilter,
     handleVoteOutcomeFilterChange: useCallback(
-      index => setOutcomeFilter(!index ? -1 : index),
+      index => setOutcomeFilter(index),
       [setOutcomeFilter]
     ),
     voteAppFilter: appFilter,
     handleVoteAppFilterChange: useCallback(
-      index => setAppFilter(!index ? -1 : index),
+      index => setAppFilter(index),
```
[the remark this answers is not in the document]
rkzel: Before -1 signified "no option chosen", resulting in "select an option" displayed as default option in DropDown. I changed it to 0, which is default now, and which points to the first real option, which is much more descriptive of what the filter does.

apps/dot-voting/app/components/VotingCardGroup.js
rkzel: At least in my browser all of that CSS was required and both ignores needed to be present, as they are very specific and have very specific targets.

apps/dot-voting/app/components/VoteDetails/index.js
rkzel: Ok. It is useful in store only it seems. Removed.

apps/dot-voting/app/components/VoteDetails/index.js
rkzel: Removed.

apps/dot-voting/app/components/VoteDetails/index.js
rkzel: Yes, balance is sourced directly from the contract in store... simplified.

apps/dot-voting/app/Decisions.js
rkzel: Done.

apps/dot-voting/app/utils/vote-styled-components.js
```diff
@@ -1,21 +1,31 @@
...
+    color: #30404F;
+    background-color: rgb(12,196,230,0.09);
+  `}>
+    <img src={badgeDetails[type].icon} alt="" />
+    <span css="margin: 0 10px; padding-top: 3px; height: 24px">{badgeDetails[type].text}</span>
+  </div>
 )
 
 GenerateBadge.propTypes = {
   type: PropTypes.string.isRequired,
```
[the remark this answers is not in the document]
rkzel: Nice! :)

PR #1226, opened 2019-09-06
updated empty card

rkzel:
Updated Empty Card as per Figma design - no other changes!


apps/address-book/app/components/Card/Empty.js
```diff
@@ -2,19 +2,18 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import styled from 'styled-components'
 import { Button, EmptyStateCard, GU, unselectable } from '@aragon/ui'
-import icon from '../../assets/empty-ab.svg'
```
chadoh: Looks like this was the only place where we used this file. Let's delete it.

apps/address-book/app/components/Card/Empty.js
```diff
@@ -2,19 +2,18 @@ import React from 'react'
...
+import emptyStatePng from '../../assets/no-contacts.png'
 
-const Icon = () => <img src={icon} alt="Empty entities icon" />
+const illustration = <img src={emptyStatePng} alt="" height="160" />
```
chadoh: ❤️

PR #1227, opened 2019-09-06
Allocations: Remove header on No budget page

e18r:
In this PR, the `if no entries` logic is consolidated into the `App` component.

Previously, the `Empty` card was displayed inside `Accounts` whenever the number of accounts was zero. Now, since `Header` is an element at the same level of `Accounts`, the logic for displaying the `Empty` card, along with hiding the `Header` was moved up to their parent component, `App`. This is similar to what I did for Address Book. See the related discussion [here](https://github.com/AutarkLabs/open-enterprise/pull/1194#pullrequestreview-281662871).

### DIFF
--- a/apps/allocations/app/components/App/Accounts.js
+++ b/apps/allocations/app/components/App/Accounts.js
@@ -5,18 +5,13 @@ import styled from 'styled-components'
 import { Badge, Text, Viewport } from '@aragon/ui'
 
 import { CARD_STRETCH_BREAKPOINT } from '../../utils/responsive'
-import { Account, Empty } from '../Card'
+import { Account } from '../Card'
 
 const Accounts = ({
   accounts,
-  onNewAccount,
   onNewAllocation,
 }) => {
   if (!accounts) return
-
-  if (accounts.length === 0) {
-    return <Empty action={onNewAccount} />
-  }
 
   return (
     <Viewport>
@@ -51,7 +46,6 @@ const Accounts = ({
 Accounts.propTypes = {
   // TODO: Create account shape
   accounts: PropTypes.arrayOf(PropTypes.object).isRequired,
-  onNewAccount: PropTypes.func.isRequired,
   onNewAllocation: PropTypes.func.isRequired,
 }
 
--- a/apps/allocations/app/components/App/App.js
+++ b/apps/allocations/app/components/App/App.js
@@ -1,10 +1,12 @@
 import React, { useState } from 'react'
+import PropTypes from 'prop-types'
 
 import { useAragonApi } from '@aragon/api-react'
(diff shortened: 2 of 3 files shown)


apps/allocations/app/components/App/Accounts.js
```diff
@@ -51,7 +46,6 @@ const Accounts = ({
 Accounts.propTypes = {
   // TODO: Create account shape
   accounts: PropTypes.arrayOf(PropTypes.object).isRequired,
-  onNewAccount: PropTypes.func.isRequired,
```
ottodevs: All changes on this file seem out of scope of what the PR title says, might we consider updating the PR title and/or comment to explain the full set of changes going here?

apps/allocations/app/components/App/App.js
```diff
@@ -122,6 +113,30 @@ const App = () => {
...
+    return <Wrap><Empty action={onNewAccount} /></Wrap>
+  }
+
+  return (
+    // TODO: Profile App with React.StrictMode, perf and why-did-you-update, apply memoization
+    <Wrap>
+      <Header
+        primary="Allocations"
+        secondary={
+          <Button mode="strong" icon={<IconPlus />} onClick={onNewAccount} label="New Account" />
```
ottodevs: This button is not present anymore in the designs, we should clarify if it is still needed, on that case, reflect the clarification in the PR comment for the reviewer to know. Otherwise, please remove the button

apps/allocations/app/components/App/Accounts.js
e18r: I hadn't even written a description of the changes, sorry. Please check out the updated description and let me know if it makes more sense.

apps/allocations/app/components/App/App.js
e18r: This PR is focused on the No budget page, in which the whole Header is not present anymore. In that sense, I believe it's out of scope to deal with that button here.

But in any case, the button is still present in the designs, except it's hidden inside the Actions menu.

apps/allocations/app/components/App/Accounts.js
ottodevs: Thanks for the clarification!

apps/allocations/app/components/App/App.js
ottodevs: Ok, I think this fits in the final checking task before mainnet deployment, where we could fine tune these kind of cosmetic details 👍

apps/allocations/app/components/App/Accounts.js
ottodevs: Thanks for the clarification!

PR #1228, opened 2019-09-09
Revert "Revert isValuePct to pure function at DotVotingMock"

chadoh:
Reverts c455c190a730a6f96a15b76b5fd6af9929d2318c

We [thought][1] we did not need this change, but tests are failing without it.

  [1]: https://github.com/AutarkLabs/open-enterprise/pull/1196#pullrequestreview-283984650


chadoh:
Merging without review, as tests on `dev` are currently failing.

PR #1229, opened 2019-09-10
Run only Allocations & dependencies

e18r:
Added commands and documentation to run only the Allocations app in
development mode along with its dependent apps (Address book & Dot
voting), thus saving precious developer time.


e18r:
@chadoh, regarding your feedback:

> I think I like this, 

Mee too!

> though I feel unqualified to comment on the template. @ottodevs @Quazia could one or both of you please take a look at that?
> 
> I guess my larger concern is that this seems to add a lot of complexity

Quite the opposite, my dear friend. If anything, it exposes some complexity that was hidden in an obscure script that gets called every time we run `npm run start:dev`. But being able to break that script down into four independent commands actually breaks that complexity down and makes it easier and faster to debug things.

> , all in service of running Allocations in a simpler way. This makes me wonder two things:
> 
> 1. Do we want a similar approach for other apps? How translatable is this work?

Yes, that's the idea. It is easily translatable, although I would need to create a few different templates.

>  2. What maintenance burden will this introduce? Is there a simpler way to accomplish the goal of working on only Allocations?

Yes, a bright colleague of mine is developing a way to mock the back end. That would definitely be a simpler and faster way to work only on a specific app's front end. But at some point you need to test the actual back end too, and this is the simplest way to do just that.

ottodevs:
I appreciate your intentions here @e18r but let me put my opinion on this.

What is the exact problem we are trying to solve here?
It can make allocations development a bit faster, I don't doubt about it, but at which cost?
We still we would need to add additional templates and npm scripts for each one of the 5 apps, that would add the complexity @chadoh is concerned about.

I think we have enough npm scripts in the package.json adding some new ones and forcing the developer to remember 4-5 scripts in a sequence for each app we need to develop sounds like crazy, even tho most of us already know these commands and what they do, so there is no clear advantage here.

The obscure script you refer to is a simple bash script that performs the same steps you mention: blockchain reset, publish, develop and start kit for every app in one single step, where is the complexity here? it transparently launches ganache-cli, with the option to reset it with just an environment variable, and saving the developer from typing every command manually and the need of opening up to 7 terminal windows if they want to develop more than 1 app at once, I think this is exactly the oposite from complexity, and it is how many complex projects solve these kind of cases.

And maintaining 5 new templates just for the sake of being able to develop them separately does not seem a realistic option for me.

Please, don't take my comments wrong, I really appreciate that you put some effort towards improving the whole team development flow, I just would like to be constructive here, the development experience is a cyclic topic in the team and it is very important for us, and for the whole aragon dev community, so most of us have spent lot of time thinking how to improve it.

We have in fact lot of mechanism to simplify some of our common development tasks, please don't hesitate to ask us if you don't understand anything relate to the development flow, or you find something could be improved or is blocking you in some way.

I honestly think our efforts to improve the DevX should be coordinated with other flock teams and focused into:
- Docker environments
- Aragen customization for our needs (I am working on a PoC for this during my free time, that will actually save a lot of time)
- TDD with cypress
- Convert our bash script to node to make it more readable since it seems lot of people get scared with bash scripts or don't have time to learn it
- Aragon/cli wrapper customized for open enterprise to selectively turn on/off some apps during development and do other tasks like republishing contracts with a nice friendly terrminal ui.

e18r:
@ottodevs Thank you for your feedback. You are far more experienced than me on this subject, so it is very valuable to me. 

> What is the exact problem we are trying to solve here?

The fact that running `npm run start:dev` takes several minutes to finish. This is no small problem. I actually think making this time shorter is the biggest productivity enhancement we can make to our development process right now.

> It can make allocations development a bit faster, I don't doubt about it,

Me neither. Saving as little as 10 seconds per build on any app can add up to several hours of developer time per month for the whole team.

> but at which cost?
> We still we would need to add additional templates and npm scripts for each one of the 5 apps, that would add the complexity @chadoh is concerned about.

Sure. But the time it would take to maintain the new templates would still be a tiny fraction of the time the whole team gained. Besides, if we're worrying about complexity, the main `package.json` is already unnecessarily complex. I actually just updated some commands that nobody was using. Instead of adding complexity, my proposal makes us more aware of what commands we actually need.

> I think we have enough npm scripts in the package.json adding some new ones and forcing the developer to remember 4-5 scripts in a sequence for each app we need to develop sounds like crazy, even tho most of us already know these commands and what they do, so there is no clear advantage here.

Let me add another NPM command that executes all the other NPM commands in a row so that you don't have to remember them. That way you can choose between executing them all at once or one by one, according to your needs.

> The obscure script you refer to is a simple bash script that performs the same steps you mention: blockchain reset, publish, develop and start kit for every app in one single step, where is the complexity here?

If something goes wrong in that bash script, there's no easy way to run the commands one by one in order to diagnose the problem. You end up running the whole thing multiple times. It *is* kind of obscure. Can you tell the path of that script off the top of your head? What's its name again? Something ganache? And how many lines long is it? Like 100? Honestly, my four npm commands are better. Sorry if you wrote that script, though. I don't mean no offense.

> it transparently launches ganache-cli, with the option to reset it with just an environment variable, and saving the developer from typing every command manually and the need of opening up to 7 terminal windows if they want to develop more than 1 app at once, I think this is exactly the oposite from complexity, and it is how many complex projects solve these kind of cases.

I really think the variable we want to optimize here is development time, not number of commands.

> And maintaining 5 new templates just for the sake of being able to develop them separately does not seem a realistic option for me.

To me it does. Let me take care of that.
 
> Please, don't take my comments wrong, I really appreciate that you put some effort towards improving the whole team development flow, I just would like to be constructive here, the development experience is a cyclic topic in the team and it is very important for us, and for the whole aragon dev community, so most of us have spent lot of time thinking how to improve it.

This improvement won't alter your development workflow. It will just add a new possibility that might end up being useful at some point.

> We have in fact lot of mechanism to simplify some of our common development tasks, please don't hesitate to ask us if you don't understand anything relate to the development flow, or you find something could be improved or is blocking you in some way.
> 
> I honestly think our efforts to improve the DevX should be coordinated with other flock teams and focused into:
> 
>     * Docker environments
> 
>     * Aragen customization for our needs (I am working on a PoC for this during my free time, that will actually save a lot of time)
> 
>     * TDD with cypress
> 
>     * Convert our bash script to node to make it more readable since it seems lot of people get scared with bash scripts or don't have time to learn it
> 
>     * Aragon/cli wrapper customized for open enterprise to selectively turn on/off some apps during development and do other tasks like republishing contracts with a nice friendly terrminal ui.

All of this is awesome! But I don't see how my proposal hinders any of this.

ottodevs:
The key is that the solution presented here is not even comparable with the script, since it is only focused for allocations.
The script is a fully fledged solution to launch development environment for the full suite of apps, tests for all apps and even coverage, without too much hassle.

Please tell me what failed for you with the script so we can help with that.

I have another solution for you, magnitudes of order simpler than your proposal, simply run: `npm start dev:allocations` that is all you need, then mock the initial react state into the appState object. You don't even need ganache-cli running for this workflow, but please, tell me if my presented solution does not work for you so we can help a move forward :)

docs/DEVELOPMENT_NOTES.md
```diff
@@ -48,6 +52,17 @@ This is the same as above, only all javascript files are published via IPFS.
...
+#### To run a specific app for development purposes
+
+The following commands will create a DAO with the Allocations app, and the apps it depends on, namely Dot Voting and Address Book.
+
+- `npm run devchain:reset` : Start a local blockchain
+- `npm run dev:allocations` : Start the development server (instead of IPFS)
+- `npm run publish:allocations` : Publish the contracts on the local blockchain
+- `npm run start:allocations` : Set up an Aragon DAO with the selected apps
+
+Note that you need to run each of these commands (except publish) on separate windows.
```
chadoh: If these could all be run with a single command from the project root (`npm run start:allocations` or `npm run dev:allocations` maybe), then I think it might be worth adding this change. 

How much time do you think using this script, vs using `start:dev`, would save?

chadoh:
@ottodevs said:

> simply run: `npm start dev:allocations`  (but the previous one that just launches allocations parcel server). That is everything you need, then mock the initial react state into the appState object.

This is what I've improved on in #1109, checking the "mock initial react state" stuff (which is far from straightforward) into our repo while avoiding having it become part of our production bundle.

I would like to see more discussion around this PR before we close it. Feedback from @Quazia & @topocount would be great.

docs/DEVELOPMENT_NOTES.md
ottodevs: I think it is not matter of "how much do you think". We should try to benchmark these things to have better metrics on the potential benefits.

ottodevs:
IMHO, this should be closed or put on hold state until we properly decide if that addition makes sense after the stubbing solution from Chad was merged last week.

I think in case there is some motivation to still merge this PR, we should discuss that here for the sake of having the reference in the own PR instead of the chats or external meetings where useful discussion could be disperse.

What are your thoughts @e18r ?

ottodevs:
I think the new template combined with the stubbing features introduced by @chadoh , superseded the functionality of this PR, so I will close for now after talking about with @e18r about it

PR #1233, opened 2019-09-11
Dot voting/gas hot fix

chadoh:
Creating this PR on behalf of @Quazia. It's in progress, but a PR helps with visibility.

### DIFF
--- a/apps/dot-voting/contracts/DotVoting.sol
+++ b/apps/dot-voting/contracts/DotVoting.sol
@@ -455,7 +455,7 @@ contract DotVoting is ADynamicForwarder, AragonApp {
         uint256 totalParticipation = voteInstance.totalParticipation;
         require(cKeys.length == supportsLength); // solium-disable-line error-reason
         require(oldSupportLength <= supportsLength); // solium-disable-line error-reason
-        _checkTotalSupport(_supports, voterStake); 
+        _checkTotalSupport(_supports, voterStake);
         uint256 i = 0;
         // This is going to cost a lot of gas... it'd be cool if there was
         // a better way to do this.
--- a/apps/dot-voting/contracts/DotVoting.sol
+++ b/apps/dot-voting/contracts/DotVoting.sol
@@ -35,7 +35,7 @@ contract DotVoting is ADynamicForwarder, AragonApp {
     bytes32 constant public ROLE_ADD_CANDIDATES = keccak256("ROLE_ADD_CANDIDATES");
     bytes32 constant public ROLE_CREATE_VOTES = keccak256("ROLE_CREATE_VOTES");
     bytes32 constant public ROLE_MODIFY_QUORUM = keccak256("ROLE_MODIFY_QUORUM");
-    bytes32 constant public ROLE_ROLE_MODIFY_CANDIDATE_SUPPORT = keccak256("ROLE_MODIFY_CANDIDATE_SUPPORT");
+    bytes32 constant public ROLE_MODIFY_CANDIDATE_SUPPORT = keccak256("ROLE_MODIFY_CANDIDATE_SUPPORT");
     string private constant ERROR_CAN_VOTE = "ERROR_CAN_VOTE";
     string private constant ERROR_MIN_QUORUM = "ERROR_MIN_QUORUM";
     string private constant ERROR_VOTE_LENGTH = "ERROR_VOTE_LENGTH";
--- a/apps/dot-voting/contracts/DotVoting.sol
+++ b/apps/dot-voting/contracts/DotVoting.sol
@@ -32,18 +32,13 @@ contract DotVoting is ADynamicForwarder, AragonApp {
 
     uint256 constant public PCT_BASE = 10 ** 18; // 0% = 0; 1% = 10^16; 100% = 10^18
 
-    bytes32 constant public CREATE_VOTES_ROLE = keccak256("CREATE_VOTES_ROLE");
-    bytes32 constant public MODIFY_QUORUM = keccak256("MODIFY_QUORUM");
-    bytes32 constant public MODIFY_CANDIDATE_SUPPORT = keccak256("MODIFY_CANDIDATE_SUPPORT");
-    bytes32 constant public ADD_CANDIDATES_ROLE = keccak256("ADD_CANDIDATES_ROLE");
-    string private constant QUORUM_SUPPORT_ERROR = "QUORUM_SUPPORT_ERROR";
-    string private constant VOTE_ID_RANGE = "VOTE_ID_RANGE";
-    string private constant CANDIDATE_RANGE_ERROR =  "CANDIDATE_RANGE_ERROR";
-    string private constant MIN_QUORUM_ISSUE = "MIN_QUORUM_ISSUE";
-    string private constant CAN_VOTE_ERROR = "CAN_VOTE_ERROR";
-    string private constant CAN_EXECUTE_ERROR = "CAN_EXECUTE_ERROR";
-    string private constant VOTE_LENGTH_ERROR = "VOTE_LENGTH_ERROR";
-    string private constant TOTAL_SUPPORT_ERROR = "TOTAL_SUPPORT_ERROR";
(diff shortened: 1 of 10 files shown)


apps/dot-voting/contracts/DotVoting.sol
```diff
@@ -36,7 +36,14 @@ contract DotVoting is ADynamicForwarder, AragonApp {
...
     bytes32 constant public ADD_CANDIDATES_ROLE = keccak256("ADD_CANDIDATES_ROLE");
-    string constant QUORUM_SUPPORT_ERROR= "Global quorum must be greater or equal to candidate support";
+    string private constant QUORUM_SUPPORT_ERROR = "QUORUM_SUPPORT_ERROR";
+    string private constant VOTE_ID_RANGE = "VOTE_ID_RANGE";
+    string private constant CANDIDATE_RANGE_ERROR =  "CANDIDATE_RANGE_ERROR";
+    string private constant MIN_QUORUM_ISSUE = "MIN_QUORUM_ISSUE";
+    string private constant CAN_VOTE_ERROR = "CAN_VOTE_ERROR";
+    string private constant CAN_EXECUTE_ERROR = "CAN_EXECUTE_ERROR";
+    string private constant VOTE_LENGTH_ERROR = "VOTE_LENGTH_ERROR";
+    string private constant TOTAL_SUPPORT_ERROR = "TOTAL_SUPPORT_ERROR";
```
chadoh: Not a big deal if this doesn't happen, but I'd prefer we sorted these sorts of lists alphabetically. I think it would make it easier to find a relevant error code to use.

I'd also prefer we standardized on `ERROR` or `ISSUE` instead of mixing the two.

Looking up to the ROLE constants, it also looks like we do not use consistent naming with those. Would it make sense to prepend all of these constants with `ERROR_` or `ROLE_`, rather than putting it and the end or, in some cases, omitting it entirely?

chadoh: Note that I am not proposing we change the strings, only the constant names. The strings could potentially omit the words "error" and "role" altogether, such as:

```sol
string private constant ERROR_CANNOT_VOTE = "Voting Forbidden"
```

I think I probably like the "error-code" style more than stripped-down error messages, but I wanted to throw it out as a possibility.

apps/dot-voting/contracts/DotVoting.sol
```diff
@@ -155,11 +162,11 @@ contract DotVoting is ADynamicForwarder, AragonApp {
     function getCandidate(uint256 _voteId, uint256 _candidateIndex)
     external view isInitialized returns(address candidateAddress, uint256 voteSupport, string metadata, bytes32 externalId1, bytes32 externalId2)
     {
-        require(_voteId < voteLength, "Vote ID outside of current vote range");
+        require(_voteId < voteLength, VOTE_LENGTH_ERROR);//, "Vote ID outside of current vote range");
         uint256 actionId = votes[_voteId].actionId;
         Action storage action = actions[actionId];
         uint256 candidateLength = action.optionKeys.length;
-        require(_candidateIndex < candidateLength, "Candidate index outside of current candidate range");
+        require(_candidateIndex < candidateLength); // solium-disable-line error-reason
```
chadoh: Why not use an error code here?

apps/dot-voting/contracts/DotVoting.sol
```diff
@@ -178,7 +185,7 @@ contract DotVoting is ADynamicForwarder, AragonApp {
     function setglobalCandidateSupportPct(uint256 _globalCandidateSupportPct)
     external auth(MODIFY_CANDIDATE_SUPPORT)
     {
-        require(globalMinQuorum >= _globalCandidateSupportPct, QUORUM_SUPPORT_ERROR);
+        require(globalMinQuorum >= _globalCandidateSupportPct); // solium-disable-line error-reason
```
chadoh: And here?

apps/dot-voting/contracts/DotVoting.sol
```diff
@@ -214,8 +221,8 @@ contract DotVoting is ADynamicForwarder, AragonApp {
     public auth(ADD_CANDIDATES_ROLE)
     {
         Vote storage voteInstance = votes[_voteId];
-        require(_voteId < voteLength, "Vote ID outside of current vote range");
-        require(_isVoteOpen(voteInstance), "Vote must be open");
+        require(_voteId < voteLength, VOTE_LENGTH_ERROR);
+        require(_isVoteOpen(voteInstance)); // solium-disable-line error-reason
```
chadoh: & here?

apps/dot-voting/contracts/DotVoting.sol
```diff
@@ -281,7 +288,7 @@ contract DotVoting is ADynamicForwarder, AragonApp {
     * @return True if the vote is elligible for execution.
     */
     function canExecute(uint256 _voteId) public view isInitialized returns (bool) {
-        require(_voteId < voteLength, "Vote ID outside of current vote range");
+        require(_voteId < voteLength, VOTE_LENGTH_ERROR); // solium-disable-line error-reason
```
chadoh: I don't think we need the `solium-disable-line` comment here

apps/dot-voting/contracts/DotVoting.sol
```diff
@@ -443,28 +450,25 @@ contract DotVoting is ADynamicForwarder, AragonApp {
...
         uint256[] storage oldVoteSupport = voteInstance.voters[msg.sender];
         bytes32[] storage cKeys = action.optionKeys;
-        uint256 cKeysLength = cKeys.length;
-        require(cKeysLength <= _supports.length, "More supports given than candidates");
-        uint256 i = 0;
+        uint256 supportsLength = _supports.length;
+        uint256 oldSupportLength = oldVoteSupport.length;
         uint256 totalParticipation = voteInstance.totalParticipation;
+        require(cKeys.length == supportsLength); // solium-disable-line error-reason
+        require(oldSupportLength <= supportsLength); // solium-disable-line error-reason
```
chadoh: Can we add error codes here?

apps/dot-voting/contracts/DotVoting.sol
```diff
@@ -474,11 +478,19 @@ contract DotVoting is ADynamicForwarder, AragonApp {
         voteInstance.voters[msg.sender] = _supports;
     }
 
+    function _checkTotalSupport(uint256[] _supports, uint256 _voterStake) internal {
+        uint256 totalSupport;
+        for (uint64 i = 0; i < _supports.length; i++) {
+            totalSupport = totalSupport.add(_supports[i]);
+        }
+        require(totalSupport <= _voterStake); // solium-disable-line error-reason
```
chadoh: And here?

apps/dot-voting/contracts/DotVoting.sol
ottodevs: I am in favour of using better messages while keeping the same constants names, I suggest using the same methodology as A1 does for their contracts, it should not be really hard to be a bit more expressive without using long strings, I think a good measure could be targeting something like 30 chars long strings (based on the average length for these reasons on other apps/contracts. also from A1)

apps/dot-voting/contracts/DotVoting.sol
Quazia: Gas limitations

apps/dot-voting/contracts/DotVoting.sol
Quazia: I'm fine to sort alphabetically and prepend ERROR_ or ROLE_ but I think the error message should just be the name, it's strictly an identifier and when I see an error the first thing I do is just key into the code, so being able to search for the actual `CONST` and get to the specific require is just quicker for me.

Quazia: https://github.com/aragon/aragon-apps/blob/master/apps/survey/contracts/Survey.sol#L25 A1 just uses the methadology of switching ERROR_ for the app name. That's fine with me but it's pretty clear what contract the error comes from. I don't want like, long sentence errors to me that's just not useful.

apps/dot-voting/contracts/DotVoting.sol
Quazia: Correct

PR #1235, opened 2019-09-12
Fix empty screen design for Allocations

ottodevs:
We missed some details like replacing the image and the text for the `EmptyStateCard`.
This PR fixes the problem


apps/allocations/app/components/Card/Empty.js
```diff
@@ -24,7 +24,6 @@ Empty.propTypes = {
 
 const EmptyWrapper = styled.div`
   ${unselectable};
-  flex-grow: 1;
```
ottodevs: I removed this rule because I used the same Empty card from AddressBook, to keep the same css on both, if it is needed for any reason we should re-add to both apps (and probably the rest of apps when their design is updated)

PR #1237, opened 2019-09-12
Allocations main page design update

e18r:
Updated the budget cards to match the design. In an effort to stub the back end, I ended up modifying the expected data that comes from web3, which looks like this currently:

    budgets = [{
      budgetId: '0',
      data: {
        name: 'Marketing',
        amount: String(80000.123856789012345678e18),
        currency: 'ETH',
        allocated: String(23000e18),
      }

Now I realise I should've asked you, @ottodevs, instead, what the new data is going to look like, because I'm not sure you have the ability to arbitrarily change the format. So I'll have to correct that depending on your feedback.

~~In order to ease the review, I didn't change the names of the files from Accounts.js and Account.js to Budgets.js and Budget.js. That's why the CI test will fail and if you need to test this locally you have to change the filenames.~~

~~Also, I don't think this should be merged in its current state because, since the web3 integration is missing, merging this into `dev` would break the Allocations app. I'm not sure what the workflow should be here, @ottodevs, maybe it's something we need to discuss. I believe that @chadoh's controversial PR might play a role in solving situations like this one.~~

~~But even if I'm putting this as a Draft, I still would like a thorough review from @chadoh and @ottodevs. Thank you.~~


ottodevs:
We should start by rebasing this PR and remove all commits except the one that is specific for allocations: b2363d0
The rest of commits are polluting the PR

apps/allocations/app/components/Card/Account.js
```diff
@@ -41,56 +39,64 @@ const Account = ({
...
-            {' ' + BigNumber(balance)
-              .div(ETH_DECIMALS)
-              .dp(3)
+          <StatsValueBig>
+            {displayCurrency(BigNumber(amount))}
+            <Text>{' ' + currency + ' per period'}</Text>
+          </StatsValueBig>
+          <StatsValueBig css={{ paddingTop: '24px' }}>
+            <ProgressBar
+              color="#1DD9D5"
```
chadoh: Let's use named colors from `theme`, if we can.

apps/allocations/app/components/Card/Account.js
```diff
@@ -41,56 +39,64 @@ const Account = ({
...
+              value={BigNumber(allocated).div(amount).toNumber()}
+            />
+          </StatsValueBig>
+          <StatsValueSmall css={{
+            color: '#212B36',
```
chadoh: & here

apps/allocations/app/components/Card/Account.js
```diff
@@ -41,56 +39,64 @@ const Account = ({
...
+            paddingTop: '8px',
+          }}>
+            {displayCurrency(BigNumber(amount).minus(allocated))}
+            <Text>{' ' + currency + ' below limit'}</Text>
+          </StatsValueSmall>
+          <StatsValueSmall css={{
+            color: '#637381',
```
chadoh: & here

apps/allocations/app/components/Card/Account.js
```diff
@@ -126,35 +134,27 @@ const CardTitle = styled(Text.Block).attrs({
...
 
 const StyledStats = styled.div`
   display: inline-block;
   text-align: center;
+  flex-grow: 1;
+`
+
+const StatsValueBig = styled.div`
+  font-size: 16px;
+  color: #637381;
```
chadoh: & here

apps/allocations/app/components/Card/Account.js
e18r: This color doesn't exist in the `theme`. Already asked Javi.

ottodevs:
I am currently implementing the changes here into the integration branch, this one should be ready to merge soon™️

apps/allocations/app/components/App/App.js
```diff
@@ -2,21 +2,50 @@ import React, { useState } from 'react'
...
 
 const nameSorter = (a, b) => a.data.name.toUpperCase() > b.data.name.toUpperCase() ? 1 : -1
 
 const App = () => {
   const [ panel, setPanel ] = useState(null)
+  const [ modal, setModal ] = useState({ visible: false, budgetId: null })
   const { api, appState } = useAragonApi()
-  const { accounts = [], balances = [], payouts = [] } = appState
+  const {
+    // backend stub, remove
```
ottodevs: We probably should not want to have this once it is presented for review unless the PR is work in progress

apps/allocations/app/components/App/App.js
```diff
@@ -2,21 +2,50 @@ import React, { useState } from 'react'
...
+          name: 'Hacktivism',
+          amount: String(38000.123856789012345678e18),
+          currency: 'DAI',
+          allocated: String(37200e18),
+          inactive: true,
+        }
+      },
+    ],
+    balances = [],
+    entries = [],
```
ottodevs: We are we re-adding entries? this was previously removed since it is part of the former address book integration, that now is becoming a dao wide IdentityProvider.

Please remove this regression.

apps/allocations/app/components/App/App.js
```diff
@@ -42,14 +71,21 @@ const App = () => {
     closePanel()
   }
 
+  const onSubmitDeactivate = (id) => {
+    console.log(`deactivating budget # ${id}...`)
+    //api.deactivateBudget(id)
```
ottodevs: Thanks for adding this here! I will take care of removing it during the integration process, this kind of practice really helps on this case.

apps/allocations/app/components/App/App.js
```diff
@@ -42,14 +71,21 @@ const App = () => {
     closePanel()
   }
 
+  const onSubmitDeactivate = (id) => {
```
ottodevs:
```suggestion
  const onSubmitDeactivate = id => {
```

apps/allocations/app/components/App/App.js
```diff
@@ -42,14 +71,21 @@ const App = () => {
...
+    //api.deactivateBudget(id)
+    closeModal()
+  }
+
   const onExecutePayout = (accountId, payoutId) => {
     api.runPayout(accountId, payoutId).toPromise()
   }
 
   const onNewBudget = () => {
+    const fundsLimit = '300000' // remove this!
```
ottodevs: 👍

apps/allocations/app/components/App/App.js
```diff
@@ -42,14 +71,21 @@ const App = () => {
...
     setPanel({
       content: NewBudget,
-      data: { heading: 'New budget', onCreateBudget }
+      data: { heading: 'New budget', onCreateBudget, fundsLimit }
```
ottodevs: If we remove fundsLimit, where will this value come from?

apps/allocations/app/components/App/App.js
```diff
@@ -69,10 +105,37 @@ const App = () => {
...
+      data: {
+        heading: 'Edit budget',
+        onCreateBudget,
+        editingBudget,
+        fundsLimit
+      }
+    })
+  }
+
+  const onDeactivate = (id) => {
```
ottodevs:
```suggestion
  const onDeactivate = id => {
```

apps/allocations/app/components/App/App.js
```diff
@@ -69,10 +105,37 @@ const App = () => {
     })
   }
 
+  const onEdit = (id) => {
```
ottodevs:
```suggestion
  const onEdit = id => {
```

apps/allocations/app/components/App/App.js
```diff
@@ -69,10 +105,37 @@ const App = () => {
...
+        fundsLimit
+      }
+    })
+  }
+
+  const onDeactivate = (id) => {
+    setModal({ visible: true, budgetId: id })
+  }
+
+  const onReactivate = (id) => {
```
ottodevs:
```suggestion
  const onReactivate = id => {
```

apps/allocations/app/components/App/App.js
```diff
@@ -82,17 +145,11 @@ const App = () => {
   const PanelContent = panel ? panel.content : null
 
   const Wrap = ({ children }) => (
-    <Main assetsUrl={ASSETS_URL}>
+    <Main assetsUrl={ASSETS_URL} scrollView={false}>
```
ottodevs: We don't really need to define `ASSETS_URL` since we are using the default value for it

apps/allocations/app/components/App/App.js
```diff
@@ -108,26 +165,71 @@ const App = () => {
...
+        state={modal}
+        onClose={closeModal}
+        onSubmit={onSubmitDeactivate}
       />
     </Wrap>
   )
 }
 
-// eslint-disable-next-line import/no-unused-modules
+const DeactivateModal = ({ state, onClose, onSubmit }) => {
```
ottodevs: I think this could definitely be moved to an external component to avoid overloading the main App component

apps/allocations/app/components/App/App.js
```diff
@@ -108,26 +165,71 @@ const App = () => {
...
+  }
+  return (
+    <Modal visible={state.visible} onClose={onClose}>
+      <div css={{ fontSize: '26px' }}>
+        Deactivate budget
+      </div>
+      <div css={{ marginTop: '32px' }}>
+        Deactivating this budget will immediately disable it once the decision is enacted. You may choose to reactivate this budget at any time.
+      </div>
+      <div css={{
```
ottodevs: Using divs for everything does not look really nice, and defeats the semantic web purposes. Try to define Styled components (we use this heavy library for some reason), if possible making them reusable, or rely instead into `Text` components from `aragon-ui` if you can (like the case here)

If the Aragon UI design system changes while we use their `Text` components, we don't need to do anything about it, but if we abuse `div`s and `css={{}}` everywhere then we need to invest lot of time maintaining and "modernizing" our apps just for a single css change in the aragon design system, please adopt the good practice now before it is too late.

Also makes our layouts difficult to read an maintain.

Also it is a wide accepted convention among web development industry to avoid using plain `div`s  for everything and it is against the World Wide Web modern standards

apps/allocations/app/components/App/App.js
```diff
@@ -108,26 +165,71 @@ const App = () => {
...
 }
 
-// eslint-disable-next-line import/no-unused-modules
+const DeactivateModal = ({ state, onClose, onSubmit }) => {
+  const deactivate = () => {
+    onSubmit(state.budgetId)
+  }
+  return (
+    <Modal visible={state.visible} onClose={onClose}>
+      <div css={{ fontSize: '26px' }}>
```
ottodevs: Please avoid using divs

apps/allocations/app/components/App/App.js
```diff
@@ -108,26 +165,71 @@ const App = () => {
...
+        Deactivate budget
+      </div>
+      <div css={{ marginTop: '32px' }}>
```
ottodevs: Each time you use a `div` Tim Berners-Lee kills a kitty

apps/allocations/app/components/App/App.js
```diff
@@ -108,26 +165,71 @@ const App = () => {
...
+          mode="negative"
+          onClick={deactivate}
+        />
+      </div>
+    </Modal>
+  )
+}
+
+DeactivateModal.propTypes = {
+  state: PropTypes.object.isRequired,
```
ottodevs: I don't think something like `{}` would be useful as a state, but it is in fact an object. What I mean is that we need to shape the kind of object that `state` expects to be.

https://reactjs.org/docs/typechecking-with-proptypes.html

These good practices involve a bit more of time (not really too much!) at the beginning, but it makes our development experience more pleasant in the future, and definitely a quick win!

apps/allocations/app/components/App/Budgets.js
```diff
@@ -0,0 +1,63 @@
...
+  return (
+    <Viewport>
+      {({ width }) => {
+        const screenSize = width
+
+        return (
+          <StyledAccounts screenSize={screenSize}>
+            {budgets.map(({ budgetId, data }) => (
+              <Budget
+                key={budgetId}
```
ottodevs: Could we have a better `key` alternative mechanism like using `uuids`? Using the same `budgetId` could make this brittle in some cases

apps/allocations/app/components/App/Budgets.js
```diff
@@ -0,0 +1,63 @@
...
+                id={budgetId}
+                name={data.name}
+                amount={data.amount}
+                currency={data.currency}
```
ottodevs: I personally think `token` is more appropriate for the space we are working on, but let's discuss this better to achieve a consensus before using one or another everywhere.

apps/allocations/app/components/Card/Budget.js
```diff
@@ -0,0 +1,213 @@
...
+  ContextMenuItem,
+  IconEdit,
+  IconPlus,
+  IconProhibited,
+  IconView,
+  ProgressBar,
+  Text,
+  useTheme,
+} from '@aragon/ui'
+import { displayCurrency } from '../../utils/helpers'
```
ottodevs: We shouldn't need to do this here, but instead separate the logic into our `app-state-reducer` to keep our components dumb and specialized into layout and style

apps/allocations/app/components/Card/Budget.js
```diff
@@ -0,0 +1,213 @@
...
+        </MenuContainer>
+        <CardTitle theme={theme}>{name}</CardTitle>
+        <StatsContainer>
+          <StyledStats>
+            <StatsValueBig theme={theme}>
+              <Text>Inactive</Text>
+            </StatsValueBig>
+          </StyledStats>
+        </StatsContainer>
+      </StyledCard>
```
ottodevs: This makes the developer so much easier to read and maintain the component instead of using a bunch of `div`s with `css` props spreading over several lines and lots of logic or conditions inside, please stick to this approach instead of what was done in the other components. ❤️

apps/allocations/app/components/Card/Budget.js
```diff
@@ -0,0 +1,213 @@
...
+            <IconProhibited />
+            <ActionLabel>Deactivate</ActionLabel>
+          </ContextMenuItem>
+        </ContextMenu>
+      </MenuContainer>
+      <CardTitle theme={theme}>{name}</CardTitle>
+      <StatsContainer>
+        <StyledStats>
+          <StatsValueBig theme={theme}>
+            {displayCurrency(BigNumber(amount))}
```
ottodevs: We should move these things to specialized logic management functions, we should not try to call any function inside the `return`

apps/allocations/app/components/Card/Budget.js
```diff
@@ -0,0 +1,213 @@
...
+
+  const reactivate = () => {
+    onReactivate(id)
+  }
+
+  if (inactive) {
+    return (
+      <StyledCard screenSize={screenSize}>
+        <MenuContainer>
+          <ContextMenu>
```
ottodevs: We are repeating lot of components here unnecessarily, please just put the differences inside the condition, and the rest of layout external to the condition.

apps/allocations/app/components/Card/Budget.js
```diff
@@ -0,0 +1,213 @@
...
+          </StyledStats>
+        </StatsContainer>
+      </StyledCard>
+    )
+  }
+
+  return (
+    <StyledCard screenSize={screenSize}>
+      <MenuContainer>
+        <ContextMenu>
```
ottodevs: We are repeating lot of components here unnecessarily, please just put the differences inside the condition, and the rest of layout external to the condition. (`inactive` condition)

Take advantage of react composability, create more specialized sub components, don't hesitate to use as much components as you need to, a fact: Facebook, the inventor of react, uses more than 50k components in their app!!!

Related and useful read: https://reactjs.org/docs/design-principles.html#stability

apps/allocations/app/components/Card/Budget.js
```diff
@@ -0,0 +1,213 @@
...
+      <StatsContainer>
+        <StyledStats>
+          <StatsValueBig theme={theme}>
+            {displayCurrency(BigNumber(amount))}
+            <Text>{' ' + currency + ' per period'}</Text>
+          </StatsValueBig>
+          <StatsValueBig css={{ paddingTop: '24px' }} theme={theme}>
+            <ProgressBar
+              color={String(theme.accentEnd)}
+              value={BigNumber(allocated).div(amount).toNumber()}
```
ottodevs: Let's try to use plain props here, I will take care of performing the calculations externally

apps/allocations/app/components/Card/Budget.js
```diff
@@ -0,0 +1,213 @@
...
+            />
+          </StatsValueBig>
+          <StatsValueSmall css={{
+            color: theme.content,
+            paddingTop: '8px',
+          }}>
+            {displayCurrency(BigNumber(amount).minus(allocated))}
```
ottodevs: This should be just an integer, or even a string, but not the result of a function called on the fly during render lifecycle

apps/allocations/app/components/Card/Budget.js
```diff
@@ -0,0 +1,213 @@
...
+            <Text>{' ' + currency + ' below limit'}</Text>
+          </StatsValueSmall>
+          <StatsValueSmall css={{
+            color: theme.contentSecondary,
+            paddingTop: '4px',
+          }}>
+            {BigNumber(amount)
```
ottodevs: Let's try to use something more useful here, like: `remainingPct` or `remainingBudgetPercentage` we will take care of performing the calculations/transformations outside instead.

apps/allocations/app/components/Card/Budget.js
```diff
@@ -0,0 +1,213 @@
...
+      </StatsContainer>
+    </StyledCard>
+  )
+}
+
+Budget.propTypes = {
+  id: PropTypes.string.isRequired,
+  name: PropTypes.string.isRequired,
+  amount: PropTypes.string.isRequired,
+  currency: PropTypes.string.isRequired,
```
ottodevs: These props will need some format changes, I will try to define them better, just a self-reminder for me

apps/allocations/app/components/Card/Budget.js
```diff
@@ -0,0 +1,213 @@
...
+  onDeactivate: PropTypes.func.isRequired,
+  onReactivate: PropTypes.func.isRequired,
+  screenSize: PropTypes.number.isRequired,
+}
+
+const StyledCard = styled(Card)`
+  display: flex;
+  margin-bottom: 2rem;
+  margin-right: ${props => props.screenSize < CARD_STRETCH_BREAKPOINT ? '0.6rem' : '2rem'};
+  box-shadow: 0px 2px 4px rgba(221, 228, 233, 0.5);
```
ottodevs: Do we have an aragon DS color definition for this shadow color?

apps/allocations/app/components/Card/Budget.js
```diff
@@ -0,0 +1,213 @@
...
+  border: 0px;
+  flex-direction: column;
+  justify-content: flex-start;
+  padding: 12px;
+  height: 264px;
+  width: ${props => props.screenSize < CARD_STRETCH_BREAKPOINT ? '100%' : BASE_CARD_WIDTH + 'px'};
+  transition: all 0.6s cubic-bezier(0.165, 0.84, 0.44, 1);
+  :hover {
+    cursor: pointer;
+    box-shadow: 0 9px 10px 0 rgba(101, 148, 170, 0.1);
```
ottodevs: Do we have an aragon DS color definition for this shadow color?

apps/allocations/app/components/Card/Budget.js
```diff
@@ -0,0 +1,213 @@
...
+  font-size: 26px;
+  font-weight: 400;
+  margin-top: 10px;
+  margin-bottom: 5px;
+  text-align: center;
+  color: ${({ theme }) => theme.content};
+  display: block;
+  /* stylelint-disable-next-line */
+  display: -webkit-box;
+  -webkit-line-clamp: 2;
```
ottodevs: We shouldn't use browser specific rules, since parcel transpiling process takes care of autoprefixing, and it seriously harms our cross-browser compatibility options.

Is there any alternative to get rid of these `webkit-` rules?

apps/allocations/app/components/Panel/NewBudget.js
```diff
@@ -1,36 +1,59 @@
...
+  constructor(props) {
+    super(props)
+    this.state =  INITIAL_STATE
+    if (props.editingBudget) {
+      this.state.name = props.editingBudget.data.name
+      this.state.nameError = false
+      this.state.amount = BigNumber(props.editingBudget.data.amount)
+        .div(ETH_DECIMALS)
+      this.state.amountError = false
+      this.state.currency = props.editingBudget.data.currency === 'ETH' ? 0 : 1 // change this!!
```
ottodevs: This is pending fix

apps/allocations/app/utils/constants.js
```diff
@@ -2,3 +2,9 @@ import { BigNumber } from 'bignumber.js'
 
 export const ETH_DECIMALS = new BigNumber(10e17)
 export const MIN_AMOUNT = new BigNumber(1e-18)
+export const STATUSES = {
+  0: 'Undergoing vote',
```
ottodevs: An Enum like structure, would seem better than an object for this case, what are your thoughts on this?

apps/allocations/app/api-react.js
```diff
@@ -40,6 +40,63 @@ const initialState = process.env.NODE_ENV !== 'production' && {
...
+        },
+        {
+          address: '0x6635F83421Bf059cd8111f180f0727128685BaE4',
+          amount: String(4e18),
+        }
+      ],
+      description: 'Some important stuff',
+      status: 0,
+      amount: String(15e18),
+      currency: 'ETH'
```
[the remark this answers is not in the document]
e18r: I'll let you change this, @ottodevs, since you have more clarity on the exact data structure required.

apps/allocations/app/api-react.js
```diff
@@ -40,6 +40,63 @@ const initialState = process.env.NODE_ENV !== 'production' && {
...
   }],
+  allocations:
+  [
+    {
+      date: String(new Date().getTime()),
+      budget: 'Marketing',
+      recipients: [
+        {
+          address: '0x75428BE833EFcAA951A466Ac58Db78A34B79104d',
+          amount: String(8e18),
```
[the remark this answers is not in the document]
e18r: I consider you to be the authority on all the data structure changes required, so I'm not gonna touch it for now.

apps/allocations/app/components/App/AllocationsHistory.js
```diff
@@ -0,0 +1,127 @@
...
+              color: theme.negative,
+              fontWeight: 600,
+            }}>
+            { displayCurrency(BigNumber(amount)) } { currency }
+          </div>
+        ]
+      }}
+      renderEntryExpansion={({ recipients, amount, currency }) => {
+        return recipients.map((recipient, index) => {
+          const allocated = BigNumber(recipient.amount).div(amount)
```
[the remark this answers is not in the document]
e18r: Totally agreed. But I'll let you change it if you don't mind. That way you can take control of the whole data flow. Next time, of course, I'll know how to do it.

apps/allocations/app/components/App/AllocationsHistory.js
```diff
@@ -0,0 +1,127 @@
...
+            </div>
+          )
+        })
+      }}
+    />
+  )
+}
+
+AllocationsHistory.propTypes = {
+  allocations: PropTypes.arrayOf(PropTypes.object).isRequired,
```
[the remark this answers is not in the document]
e18r: Awesome! Thank you for this explanation.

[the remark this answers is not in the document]
e18r: I now know how to do it, but again, I'll let you do it so that you're able to be in control of the whole data flow.

apps/allocations/app/components/App/App.js
e18r: It was meant for you to remove it after you get an idea of the data structures used

apps/allocations/app/components/App/App.js
e18r: I think it's the total amount the DAO has in its vault, minus the sum of the amount of every Budget. Maybe it should be an array of objects, each one representing the funds limit of every token.

apps/allocations/app/components/App/App.js
e18r: done, tested with frontend only

apps/allocations/app/components/Card/Budget.js
e18r: Yes we do. Thank you for pointing that out.

apps/allocations/app/components/Card/Budget.js
e18r: Not really. It's a bit similar to `theme.contentSecondary`, but it feels really odd to use that here.

apps/allocations/app/utils/constants.js
e18r: Definitely. Please feel free to update this.

apps/allocations/app/components/Panel/NewBudget.js
e18r: Yes. Please update this in accordance with the data structure.

apps/allocations/app/components/App/Budgets.js
e18r: Do we really need this? In what scenarios are `budgetId`s going to be duplicated? Maybe it should be a matter of ensuring upstream that they are unique.

apps/allocations/app/components/App/Budgets.js
e18r: I don't have any objections for this. Please feel free to update accordingly.

apps/allocations/app/components/Card/Budget.js
e18r: Got it. I'll do it like that next time. But this time I'll leave it to you so that my commits are still testable in a standalone way.

apps/allocations/app/components/Card/Budget.js
e18r: OK. Please update according to the data structure.

apps/allocations/app/components/App/Budgets.js
ottodevs: Is more related to re-renders, in the case an element gets removed and another takes the same id, it could cause some issues with React reconciliation, but as you point it is probably an unlikely scenario for this specific case and more a good practice to adopt :)

apps/allocations/app/components/Card/Budget.js
ottodevs: I will probably modify this part to take advantage of the appStateReducer transformations for this data

chadoh:
> What should be the strategy here? Do we wait until getting web3/allocations branch merged into dev or do we merge this one into web3/allocations first?

I prefer merging web3/allocations to `dev`, then retarget this one at `dev`. It keeps the history easier to read.

ottodevs:
> > What should be the strategy here? Do we wait until getting web3/allocations branch merged into dev or do we merge this one into web3/allocations first?
> 
> I prefer merging web3/allocations to `dev`, then retarget this one at `dev`. It keeps the history easier to read.

We did that and it didn't seem to work, closing this for now, as it seems that the changes were already integrated into other branchs, and the amount of change here don't seem something manageable.

Next time we should take a better approach for branching to avoid these situations.

PR #1239, opened 2019-09-13
Allocations: fix creating new account/budget

chadoh:
We're in a halfway state right now. We don't have new contracts yet, but we're starting to get new UI.

This keeps the new UI working with the current contracts.

The comparison of a raw JS Number object with `MIN_AMOUNT`, which is a `BigNumber`, was also failing intermittently.

### DIFF
--- a/apps/allocations/app/components/Panel/NewBudget.js
+++ b/apps/allocations/app/components/Panel/NewBudget.js
@@ -5,6 +5,7 @@ import styled from 'styled-components'
 
 import { Form, FormField } from '../Form'
 import { isStringEmpty } from '../../utils/helpers'
+import { BigNumber } from 'bignumber.js'
 import { MIN_AMOUNT } from '../../utils/constants'
 
 // TODO:: This should be votingTokens from account?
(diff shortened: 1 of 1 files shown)


chadoh:
I can fix the step while here. What’s a good step value, do you think? 0.1?

chadoh:
@ottodevs it looks like the browser still starts with the `min` value, when using the keyboard arrow keys to increment the value, even if a larger `step` is defined. We will either need to set `min="0"` and then enforce the `1e-18` minimum with JS-only, or leave it as-is. We can discuss with design later, but for now I added in a `step={0.1}` only.

e18r:
> @ottodevs it looks like the browser still starts with the `min` value, when using the keyboard arrow keys to increment the value, even if a larger `step` is defined. We will either need to set `min="0"` and then enforce the `1e-18` minimum with JS-only, or leave it as-is. We can discuss with design later, but for now I added in a `step={0.1}` only.

Setting a `step={0.1}` creates a HTML5 validation error whenever users input values that aren't multiples of 0.1. Users are still able to submit the form though.

chadoh:
Oh, I did not test that! Eek. Thanks for the catch, @e18r. Are you fixing that already in a separate PR?

e18r:
yeah, don't worry.

PR #1240, opened 2019-09-13
Dot Voting/api-react: more closely mirror actual

chadoh:
Comparing the data in this stub to a logged `appState` from App.js revealed some inconsistencies.


chadoh:
Yeah, I didn't expect the value fields to be strings, but when I looked at the actual state returned by our current script.js, they were strings. 

Let's for sure review again once our contracts are all updated!

PR #1244, opened 2019-09-14
[chore] Make git clean script options safer

ottodevs:
This is just a minimal correction to our clean script:

- Previously we were using `git clean -ffxd` for that, the issue with the lowercase `x` is that it also removes user created files and folders, so the developer can accidentally remove a bunch of work when cleaning, it happened to me lots of times.

- By using uppercase `X` in `git clean -fXd` those manually created files are kept, making the script safer.

source: https://git-scm.com/docs/git-clean#Documentation/git-clean.txt--X


apps/dot-voting/contracts/DotVoting.sol
```diff
@@ -444,7 +445,8 @@ contract DotVoting is ADynamicForwarder, AragonApp {
         uint256[] storage oldVoteSupport = voteInstance.voters[msg.sender];
         bytes32[] storage cKeys = action.optionKeys;
         uint256 cKeysLength = cKeys.length;
-        require(cKeysLength <= _supports.length, "More supports given than candidates");
+        // require(cKeysLength <= _supports.length, "More supports given than candidates");
+        require(cKeysLength <= _supports.length);
```
chadoh: I don't think these `require` changes should be here

ottodevs: Good catch, it was a mistake from my side, fixed now

PR #1246, opened 2019-09-18
README: add info about debugging contracts

chadoh:
Steps for troubleshooting (calls to) contracts using `truffle console`

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,78 @@ $ npm run start:dot
 
 **Detailed information in the [DEVELOPMENT_NOTES.md](/docs/DEVELOPMENT_NOTES.md) document.**
 
+### Debugging contracts
+
+If a call to a contract from the frontend fails, you may not get much information about the failure in the console. One of the best ways to troubleshoot is to tweak the contract code and try again, but restarting the whole development stack is time-consuming. Here's a way to make it faster:
+
+1. Record the info that was sent to the contract from the frontend (you can log the values to your browser console). Example:
+
+   ```js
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -48,6 +48,78 @@ $ npm run start:dot
...
+   ```
+
+   Or anything else listed in [the web3 docs](https://web3js.readthedocs.io/en/v1.2.1/web3-eth-contract.html)
+
+7. Make changes to your contract code
+
+8. Re-compile & deploy your contract
+
+   ```
+   ???
```
chadoh: @topocount what commands do we want here?

ottodevs: We should just remove 7 and 8 steps. Or replace by something like 
> ask a solidity developer to look into it if you think there is an issue with the contract

ottodevs:
> Steps for troubleshooting (calls to) contracts using `truffle console`

I think the title could be confusing, it is important to understand the difference between debugging frontend issues with contract calls and debugging contracts.

For the latter there are some really good approaches, use remix, tests, vs code debugging, truffle console is so good but not the only one, why would we want to stick to just one method? Each developer would have his own tricks, maybe this is better to fit a technical article on the blog commenting on the different options instead, but this is maybe another topic :)

ottodevs:
I would just put this PR on hold until the mainnet release is completed, I don't think we should  extra-time on making technical tutorials right now, but I definitely think this is a really nice idea to have in the future, maybe in a more visible place than the root README.md (we already have more specialized MD documents for technical topics)

chadoh:
Thanks for the input, @ottodevs. This is what I came up with while in a call with Kevin, for debugging the contract. Sounds like maybe there are better ways to do it, though. I'll close this without merging; I wanted to have it documented somewhere, and I don't mind having it in an unmerged PR for now.

Issue #1166, opened 2019-08-12
Disable Submit button in allocations if metamask isn't enabled


stellarmagnet:
We can already detect with api.accounts() whether the person has a connected account or not. If not, you can choose to disable the button

lcoenen:
I'll take this :)

lcoenen:
In the new version 0.8, Aragon One is using a dedicated panel to inform the user that there's no connected account. 

<img width="430" alt="image" src="https://user-images.githubusercontent.com/1237971/65344064-902b1800-dbce-11e9-8e10-ec420dfcbc07.png">

The behaviour is the same in Token, Voting and Finance, Address book, ...

ottodevs:
I would rely in this new feature from Aragon 0.8 client.

Just disabling the button could be confusing for the user that expects the button to be enabled after properly fill every input field, since the same behaviour (button disabled) also occurs when some fields are not validated against several rules, and all this happens without any visual advice besides the button being disabled all the time, so I would reconsider the story for this issue or just stick to how Aragon 0.8 manages it, as stated by @lcoenen (Thanks!)

Opinions @stellarmagnet ?

PR #1247, opened 2019-09-19
1166 disable submit buttons metamask

lcoenen:
This add `connectedAccount` as props of `newBudget` and `newAllocations` panels, and link the `disabled` props of `<Form` so the button is disabled when no users are connected.

See #1166 
ezgif com-video-to-gif(1)


ottodevs:
I commented on #1166 about this, the idea and code seems good, but we need to probably complete our migration to the 0.8.0 client version and then check if this is still needed and how we could improve the feedback for the user to avoid mixing this with input validations as button disabling reason without properly notify that.

Let's put this on hold for now if you don't mind @lcoenen

lcoenen:
Closing this as it is stale.

PR #1249, opened 2019-09-20
Use latest (experimental) versions of wrapper & api

chadoh:
* use @aragon/api from https://github.com/AutarkLabs/aragon-api
* use local path to `aragon` cli tool from under-review tweak at https://github.com/aragon/aragon-cli/pull/781
* point custom `aragon` cli at experimental for of aragon client, https://github.com/AutarkLabs/aragon/pull/27
* this custom aragon client uses our own experimental version of @aragon/wrapper, https://github.com/AutarkLabs/aragon-wrapper

This currently introduces a race condition during the `start:dev` run, where sometimes a contract gets deployed before the devchain has fully started.


apps/address-book/package.json
```diff
@@ -15,7 +15,7 @@
     "lint": "solium --dir ./contracts",
     "precommit": "lint-staged",
     "prepublishOnly": "truffle compile --all",
-    "publish:http": "npm run build:script && aragon apm publish major --http localhost:1111 --http-served-from ./dist",
+    "publish:http": "npm run build:script && aragon apm publish major --http localhost:1111 --http-served-from ./dist --propagate-content false --skip-confirmation true",
```
ottodevs: This is not related to using experimental versions of the wrapper, but to upgrading the `@aragon/cli` that is not needed at all for the scope of this task. I also spent lot of time extensively researching the update strategy, so I would like to ask you to revert these changes and wait until I present my solution to use the new cli.

To clarify again: We don't need to use the newest @aragon/cli versions for the scope of this PR

Just use the `--client-version` option to test Camino client version (that should be also reflected on the title/description of the PR if you decide to do so.

The client is a different thing from the wrapper and the api and `@aragon/cli` is different from `@aragon/aragon` aka **the client™️** and definitely unrelated to `@aragon/aragon.js` (the monorepo that contains the wrapper and api that the title mentions)

https://hack.aragon.org/docs/cli-main-commands#aragon-run

apps/planning-suite-kit/package.json
```diff
@@ -6,7 +6,7 @@
     "coverage": "SOLIDITY_COVERAGE=true npm run ganache-cli:test",
     "ganache-cli:test": "../shared/test-helpers/ganache-cli.sh",
     "lint": "solium --dir ./contracts",
-    "start:kit": "aragon run --template PlanningKit --template-init @ARAGON_ENS --client-version e08deb62080e17fe652fa83f60dc81a02c455c41",
+    "start:kit": "../../node_modules/@aragon/cli-monorepo/packages/aragon-cli/dist/cli.js run --template PlanningKit --template-init @ARAGON_ENS --client-version 5199d2f3356566a5bf5a5dc74ebc61a7d9b3d32a --client-repo https://github.com/AutarkLabs/aragon",
```
ottodevs: This is only working on your computer, please revert the line to use the global `aragon` package command instead of your locally downloaded `@aragon/cli`

The newest @aragon/cli has some compatibility issues and this change involves some sensible issues with the templates and some contracts, we should not just try to change this without proper testing.

ottodevs: We shouldn't try to merge non-working code

apps/projects/app/store/init.js
```diff
@@ -28,13 +19,23 @@ export const initStore = vaultAddress => {
...
+const initState = (vaultContract) => async (cachedState = INITIAL_STATE) => {
+  let nextState = await initializeTokens(cachedState, vaultContract)
+  const github = await app.getCache('github').toPromise()
+  if (github && github.token) {
+    nextState.github = github
+    initializeGraphQLClient(github.token)
+  }
+
+  return nextState
+}
```
ottodevs: All these changes going on into the projects store seem unrelated to the scope of the PR, they need to be separated and put into another request.

apps/address-book/package.json
ottodevs: Another alternative is using `npm link` if you need to test alternative repos locally, we've been doing this last year and it does not represent any difficulty, I think that is a better solution than merging these kind of changes (using special repo versions) into the dev branch, at least just for the sake of using a different `cli` that is not needed here

apps/planning-suite-kit/package.json
chadoh: This PR is still marked as draft. I had no intention of merging it with this line as it is 😄

apps/address-book/package.json
chadoh: We need a new version of the CLI in order to add the `--client-repo` option to the `aragon run` command, so that we can use our fork of the client.

> Another alternative is using `npm link` if you need to test alternative repos locally, we've been doing this during the whole year and it does not represent any difficulty

"Does not represent _any_ difficulty"? Maybe I'm doing it wrong; let's have a call so you can set me straight. I don't think it's a maintainable solution to _require_ all members of our team to do this (with specific branch of the client, linked to a specific branch of aragon.js) just to have baseline-functional apps during development. Also, when linking, I think I observed some script.js code gets minified, which results in much less actionable stack traces. This slows me down while working on web3 integration for Projects, and I don't think I can afford any slow-downs there. But again, maybe I configured something incorrectly in my environment, and there's a zero-hassle way to accomplish this.

I agree that needing to upgrade the CLI is unfortunate; I would love to avoid it if we can. Using `npm link` did not seem like a workable solution to me. Let's have a call and figure it out.

apps/projects/app/store/init.js
chadoh: We need these because the `api.cache` call went away and we need to use triggers instead. Maybe there's a simpler way to accomplish this; I followed the pattern started [in this PR](https://github.com/AutarkLabs/open-enterprise/pull/1160/files#diff-a62393c34897e310e6892a3aaf002ad2). Do you think it's important to make it more minimal?

README.md
```diff
@@ -19,9 +19,8 @@ Open Enterprise is a collection of Aragon apps that enable organizations to cura
 
 ## Development Setup
 
-Node.js LTS or greater required.
-
-- Note: @aragon/cli and truffle npm deps are automatically installed when bootstrapping.
+* Node.js LTS or greater required. You can use a tool like [`asdf`](https://asdf-vm.com/) or [`nvm`](https://github.com/nvm-sh/nvm) to manage versions of node.
```
ottodevs: That seems too obvious, and maybe adding `engine` field to `package.json` would help since we are not supporting every node version from LTS right now, add least we didn't test them all, we are in fact using 10.x branches in places like CI, so this statement seems not really accurate.
Said that I like the suggestions about the node versions managers, I didn't know about `asdf` myself, so I liked this message... I let to you the final decision on how to make up this based on my comment @chadoh or just keeping it as is, if you consider that.

apps/discussions/package.json
```diff
@@ -53,11 +53,11 @@
     "build": "npm run compile && npm run sync-assets && npm run build:app && npm run build:script",
     "build:app": "parcel build ./app/index.html -d dist/ --public-url \".\" --no-cache",
     "build:script": "parcel build ./app/script.js -d dist/ --no-cache",
-    "publish:cd": "aragon apm publish major --files dist/ --environment continuous-deployment --apm.ipfs.rpc https://ipfs.autark.xyz:5001 --ipfs-check false",    
+    "publish:cd": "aragon apm publish major --files dist/ --environment continuous-deployment --apm.ipfs.rpc https://ipfs.autark.xyz:5001 --ipfs-check false",
```
ottodevs: It seems you missed `--skip-confirmation true` on all these scripts, I don't see why would we need the option for http but not for cd, so this is probably missing it.

apps/address-book/package.json
```diff
@@ -17,7 +17,7 @@
     "prepublishOnly": "truffle compile --all",
     "publish": "aragon apm publish patch && sleep 3",
     "publish:cd": "aragon apm publish major --files dist/ --environment continuous-deployment --apm.ipfs.rpc https://ipfs.autark.xyz:5001 --ipfs-check false",
```
ottodevs: --skip-confirmation true seems to missing, I would ensure it is present on every `aragon apm publish`,  `aragon run` or `aragon start` call, to keep them working unattended

apps/shared/test-helpers/ganache-cli.sh
```diff
@@ -38,6 +38,7 @@ start_testrpc() {
 		aragon devchain --reset --port "$testrpc_port" &
 	elif [ "$DEV" = true ]; then
 		aragon devchain --reset --port "$testrpc_port" &
+		sleep 10 # wait for devchain to start TODO: modify cli to return rather than require interruption
```
ottodevs: No the most efficient way to do, but if that works, for now should be enough...
Hopefully we will get rid of this whole script sooner rather than later

chadoh: Exactly! I added similar comments to the commit message where I did this.

README.md
chadoh: Oh, gotcha, yeah, that makes sense. I was keeping the comment that had been there somewhat intact, "Node.js LTS or greater required" – I would have preferred to not change it at all, for now. Adding `engine` to `package.json` is on my list of things to do; maybe we can improve this line and add that in a follow-up PR?

apps/address-book/package.json
```diff
@@ -16,8 +16,8 @@
     "precommit": "lint-staged",
     "prepublishOnly": "truffle compile --all",
     "publish": "aragon apm publish patch && sleep 3",
-    "publish:cd": "aragon apm publish major --files dist/ --environment continuous-deployment --apm.ipfs.rpc https://ipfs.autark.xyz:5001 --ipfs-check false",
-    "publish:http": "npm run build:script && aragon apm publish major --http localhost:1111 --http-served-from ./dist",
+    "publish:cd": "yes | aragon apm publish major --files dist/ --environment continuous-deployment --apm.ipfs.rpc https://ipfs.autark.xyz:5001 --ipfs-check false --skip-confirmation true",
```
ottodevs: I guess if we add the skip confirmation option we shouldn't need the `yes` solution, but it shouldn't harm to have both for now, so why not?

chadoh: We do need both, unfortunately. With only `--skip-confirmation true`, it still gets stuck at this step:

    Couldn't reuse artifact due to mismatches, regenerate now? [y]es/[a]bort

I consider it a bug in aragonCLI that `--skip-confirmation` doesn't apply here, and I'd like to see one of us submit a patch soon.

PR #1250, opened 2019-09-20
Remove `concurrency: 1` option from lerna

chadoh:
This runs everything in parallel by default, which means we can remove our `--parallel` flags.

This results in much faster startup of our `start:dev` process, at the cost of making machines quite slow while it runs and having somewhat more chaotic logs.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,8 +1,5 @@
 {
   "command": {
-    "run": {
-      "concurrency": 1
-    },
     "bootstrap": {
       "progress": true
     }
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "dev": "apps/shared/test-helpers/develop-kit.sh",
     "devchain:reset": "aragon devchain --reset",
     "devchain": "lerna run devchain --stream",
-    "frontend": "lerna run frontend --stream --parallel",
+    "frontend": "lerna run frontend --parallel",
     "frontend:address": "cd apps/address-book && npm run frontend",
     "frontend:allocations": "cd apps/allocations && npm run frontend",
     "frontend:dot": "cd apps/dot-voting && npm run frontend",


chadoh:
@ottodevs I suspect you have the most familiarity with these sorts of tweaks. Do you know of a reason to avoid doing this?

ottodevs:
I guess because `aragon/cli` relies on `truffle` to publish the contracts to `apm` and `truffle`/`ganache-cli` may not be very good with deploying things in parallel?, I remember having issues in the past with that, but hey, if you tested it and that works it should be a nice improvement.

Another reason could be that following the output while publishing in parallel could be a bit chaotic, specially if some error happens during publishing.

I would like to personally test this a bit before merging if that makes sense, I think the improvement shouldn't be that much to justify merging this quickly without proper testing, do we have a benchmark on the benefits for this?

Let's discuss on these technical details if you don't mind!, thanks for taking the time to do this work!

package.json
```diff
@@ -61,7 +61,7 @@
     "mine": "node ./apps/shared/test-helpers/mine.js",
     "postinstall": "npm run bootstrap",
     "publish:apps": "lerna run publish",
-    "publish:http": "lerna run --scope='@tps/apps-*' publish:http --stream",
+    "publish:http": "lerna run --scope='@tps/apps-*' publish:http --stream --parallel",
```
ottodevs: I would probably try to remove concurrency setting from `lerna.json` config file instead, since lerna by default works in parallel, that config setting was specifically added for this npm script (you may quickly check this by searching "lerna run" recursively and find that there are more `--parallel` uses and just a single `--concurrency 1` use case

chadoh:
> ...if you tested it and that works it should be a nice improvement.
> 
> Another reason could be that following the output while publishing in parallel could be a bit chaotic, specially if some error happens during publishing.
> 
> I would like to personally test this a bit before merging if that makes sense, I think the improvement shouldn't be that much to justify merging this quickly without proper testing, do we have a benchmark on the benefits for this?

I haven't benchmarked it, I've only tested it a little bit. I tried it and noticed a drastic improvement, and didn't think the parallel output was worrisome. I thought maybe we could test it out by merging it in, and then see if there's a way we want to improve it from there. 

If that feels too reckless and you want to test it extensively before merging, we can do that too, though I'm not sure the rigor/strictness is worth slowing down our dev environments. If it will cause known bugs, then yeah, I don't want to merge it. If it's not optimal, well... neither is what we have now, so maybe the short-term trade-off favors merging this now and cleaning later.

I'll leave it up to you. I'm happy to close this for now, if that's what you think is best.

chadoh:
@ottodevs I've updated it to run things in parallel by default, but I'm observing this strange behavior where the `allocations` frontend process does not get started. Any idea why this might be the case?

chadoh:
If I run `lerna run frontend --stream`, it only runs four of the apps, excluding allocations.

if I run `lerna run frontend --stream --parallel`, it runs all five of the apps. Really strange.

stellarmagnet:
@chadoh is this PR still needed?

chadoh:
@ottodevs I think I last asked some questions to you. I still think this seems like a good improvement to make. Let me know if you have ideas of how to fix the problems I ran into.

apps/shared/test-helpers/develop-kit.sh
```diff
@@ -72,7 +72,7 @@ if multi_parcel_running; then
 fi
 
 start_multi_parcel() {
-	lerna run dev --log-level=silent --parallel "$@" &
+	lerna run dev --log-level=silent "$@" &
```
ottodevs: yeah, let's add the `--parallel` flag, now we measured its utility!
```suggestion
	lerna run dev --log-level=silent --parallel "$@" &
```

package.json
```diff
@@ -44,7 +44,7 @@
     "dev": "apps/shared/test-helpers/develop-kit.sh",
     "devchain:reset": "aragon devchain --reset",
     "devchain": "lerna run devchain --stream",
-    "frontend": "lerna run frontend --stream --parallel",
+    "frontend": "lerna run frontend --stream",
```
ottodevs: Yeah, let's re-add the parallel flag for a boost, but also let's remove `--stream` that seems redundant when using with `--parallel` (reference: https://github.com/lerna/lerna/tree/master/commands/run#--parallel))

```suggestion
    "frontend": "lerna run frontend --parallel",
```

lerna.json
```diff
@@ -1,8 +1,5 @@
 {
   "command": {
-    "run": {
-      "concurrency": 1
```
ottodevs: This seems safe to remove for the `run` command! Good move here.

package.json
chadoh: Strange that you saw a difference with this! It doesn't seem like it should affect it at all, to me, from what I understand of lerna. But it also won't hurt anything, and it seems like it might actually save some time – though I do wonder if you were just seeing some statistical noise, there.

chadoh:
Ok, @ottodevs, updated!

package.json
ottodevs: Yeah, it is just for cleanup purposes, as per lerna documentation `--parallel` flag already enables `--stream` so no need to add that flag again :)

PR #1245, opened 2019-09-17
Allocations design check: new allocation side panel

e18r:
fixes https://github.com/AutarkLabs/open-enterprise/issues/841
fixes https://github.com/AutarkLabs/open-enterprise/issues/892

screenshot_2019-09-17_14-48-21_120122364

PR #1256, opened 2019-09-24
Address book new entity

lcoenen:
This implement a new on-time validation on Address book app!

This is based on PR #1245, so please merge it first to see the change. 

- This adds on-time validation on the NewEntity Panel of the Address book.
- This adds a custom entity type on the same panel.

**Remarks**:

- The exact behaviour wasn't specified, so the system is taking the first error and showing it below the submit button.
- There is tons of logic to DRY up between the two different forms. We might want to integrate something like Final Forms into aragon-ui to abstract and uniform validation logic out of the forms. 

See https://youtu.be/m5qysCQm4Mg for a DEMO

**Edited**:

- We deleted the whole `touched` logic to better match with design decision. 
- Also, it have been rebased against `dev`. 

New demo: https://youtu.be/mKKZsuuzpOo

### DIFF
--- a/apps/address-book/app/components/Form/Form.js
+++ b/apps/address-book/app/components/Form/Form.js
@@ -2,14 +2,24 @@ import PropTypes from 'prop-types'
 import React from 'react'
 import { Button, Text, useTheme } from '@aragon/ui'
 
-const Form = ({ children, onSubmit, submitText, heading, subHeading }) => {
+const Form = ({
+  children,
+  onSubmit,
+  submitText,
+  heading,
+  subHeading,
+  disabled,
+  error
+}) => {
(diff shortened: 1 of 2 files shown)


lcoenen:
It looks like linting is failing. I don't get why, given that the error doesn't match the code on my branch. It's testing against dev for some reason.

image

https://github.com/lcoenen/open-enterprise/blob/5acbe7d0f61a97b837275890b9822fb1fed6988d/apps/address-book/app/components/Panel/NewEntity.js#L70

e18r:
@lcoenen please rebase your branch with `dev` in order to remove the unrelated commits. Thank you!

PR #1260, opened 2019-09-25
Fix creating Allocations/Dot Votes

chadoh:
At some point recently we started getting an error when trying to create an allocation:

    Uncaught (in promise) Error: invalid number value (arg="_amount", coderType="uint256", value=1000000000000000000)

The `String(balance)` fixes this issue, but another error hid behind it:

    Uncaught (in promise) Error: invalid address (arg="_token", coderType="address", value=undefined)

The initialization of tokenAddress fixes this problem.

Now creating allocations works again. Hooray!

### DIFF
--- a/apps/allocations/app/components/App/App.js
+++ b/apps/allocations/app/components/App/App.js
@@ -36,7 +36,7 @@ const App = () => {
       payoutId,
       recurring,
       period,
-      balance,
+      String(balance),
       tokenAddress
     ).toPromise()
     closePanel()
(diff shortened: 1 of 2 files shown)


chadoh:
@ottodevs @e18r I assume that the work you two are doing will render this obsolete soon. In the meantime, we need Allocations to function correctly so we can test the fixes I made to Dot Voting in #1236. This fixes the issues you two had been seeing so we can get those Dot Voting changes QA'd properly.

PR #1261, opened 2019-09-26
audit: remove 32 bytes from copy function

topocount:
Just a small change to comply with the audit warning about the extra 32 bytes in the copy length here: https://github.com/AutarkLabs/open-enterprise/commit/549073274f690f65aef2e01dd68dc25703cbffce#diff-9e59579be3cea1e79854a7134cc497eaR55

I've added an extra param to `ExecutionTarget.sol` to ensure that the tests work with more than a single additional static parameter


ottodevs:
> @ottodevs the contracts are locked post-audit :(

We shouldn't operate this way... one person addresses some changes, then the contract gets locked and sent to the auditors, without any previous internal review or any control... Anyway it's done, just a note to improve our processes in the future.

PR #1262, opened 2019-09-26
Rewards Design Update

e18r:
This PR is ready to be reviewed and merged. I'm setting it as a Draft simply because it can't be merged to `dev` without the web3 integration. Whenever you create a branch with your work on Rewards web3 integration, you let me know and I point this PR towards that branch and remove the Draft status.

* New Reward Side Panel:
Screenshot from 2019-10-01 10-53-20

* New Reward Side Panel review Reward page:
screenshot_2019-09-28_00-10-29_908130647

* Main page DataView:
screenshot_2019-09-29_15-33-16_642130326

* Empty card:
Screenshot from 2019-10-01 10-54-04

* View Reward Side Panel:
Screenshot from 2019-09-30 19-21-43

* Metrics:
Screenshot from 2019-09-30 21-46-35

* My Rewards tab:
Screenshot from 2019-10-01 11-52-37


jayalaves:
Nice work @e18r! Some fixes to make here and there:

- Missing a 24px spacing above the first input in New Reward Side Panel
- The empty card SVG illustration seems to overextend its container on the top
- That empty card design is not _exactly_ as designed, but if that's the fastest approach I think it's okay to approve until we ship an improved layout
- Metrics have to be left-aligned with the container, imagine you have 4 columns with the metrics placed from left to right and you'll have the fourth column empty

Let me know once you can check these out and see if any updates are possible within the time we have.

e18r:
Thank you so much for your feedback, @jayalaves! I think this is a very healthy work dynamic and I want us to keep doing it as much as we can. Regarding your comments:

> * Missing a 24px spacing above the first input in New Reward Side Panel

Nice catch! Solved

> * The empty card SVG illustration seems to overextend its container on the top

I thought that was cool, but solved

> * That empty card view is not _exactly_ as I thought of it, but if that's the fastest approach I think it's okay to approve until we ship an improved layout

Yeah, I think it would be important to detail this behavior in the design, especially in order to understand what exactly to show in each section when one section is empty but the other isn't.

> * Metrics have to be left-aligned with the container, imagine you have 4 columns with the metrics placed from left to right and you'll have the fourth column empty

Right now I'm using [`justify-content: space-evenly`](https://developer.mozilla.org/en-US/docs/Web/CSS/justify-content) to organize the metrics inside the container. I could set it to `start` and set a fixed margin, between them, but that wouldn't be very responsive. I could also create an extra empty metric to the right so that all the elements shift a bit to the left, but again, that wouldn't be very responsive either. So I'd prefer to wait for the design of the mobile screens to better understand how this should scale down.

jayalaves:
> I thought that was cool, but solved

Me too! But we need to be consistent with all the other designs :)

> Right now I'm using justify-content: space-evenly to organize the metrics inside the container.

Okay, that's a call to action for me to start picturing the mobile views. I'll get to it and we can revisit this topic.

@e18r Understood, thanks for the updates. I'll accommodate the new changes and specifications within this week such that we can implement them later on. Some additional comments on the above views:

- In the New Reward panel, the red stroke around the error input seems a little off, since it doesn't look like it's positioned in the same way as the blue stroke on the metric specifier next to it.
- Unclaimed rewards table missing a context menu per entry? How do you claim or view the reward details?

e18r:
Javi,

> * In the New Reward panel, the red stroke around the error input seems a little off, since it doesn't look like it's positioned in the same way as the blue stroke on the metric specifier next to it.

Gotcha. I'll try to correct it, although it can become a little hackish.

> * Unclaimed rewards table missing a context menu per entry? How do you claim or view the reward details?

Glad you asked. Please make the designs for that :)

e18r:
> Javi,
> 
> > * In the New Reward panel, the red stroke around the error input seems a little off, since it doesn't look like it's positioned in the same way as the blue stroke on the metric specifier next to it.
> 
> Gotcha. I'll try to correct it, although it can become a little hackish.

Corrected it. It was a Firefox-only thing, so I just disabled the red border so that it behaves like Chrome.

PR #1263, opened 2019-09-26
Dv supports value fix

Quazia:
It looks like we were sort of doing too much in an effort to get around the slider issues, BigNumber type issues, and the like. I've tested this with multiple values that add up to 100, multiple values that don't add up to 100, values of 100 with a value of 0, a single value of 100, a single non-100 value and they all seem to work.

Main issue seemed to have been either not presenting a BigNumber value, or presenting an exponential value and running into problems around the ether.js parser. Right now it just leverages a BigNumber library, does the calculation using that, and spits out a non exponentiated string value which works well with web3.

### DIFF
--- a/apps/dot-voting/app/components/VoteDetails/CastVote.js
+++ b/apps/dot-voting/app/components/VoteDetails/CastVote.js
@@ -8,7 +8,7 @@ import { Button, Text, useTheme } from '@aragon/ui'
 import tokenBalanceOfAbi from '../../abi/token-balanceof.json'
 import Slider from '../Slider'
 import Label from './Label'
-import { BigNumber } from 'bignumber.js'
+import { BN } from 'web3-utils'
 
 const ValueContainer = styled.div`
   box-shadow: 0 4px 4px 0 rgba(0, 0, 0, 0.03);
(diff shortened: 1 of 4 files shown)


apps/dot-voting/app/components/VoteDetails/CastVote.js
```diff
@@ -166,7 +153,7 @@ const CastVote = ({ onVote, toggleVotingMode, vote, voteWeights }) => {
           mode="strong"
           onClick={handleVoteSubmit}
           disabled={
-            voteOptions.reduce((sum, { trueValue = 0 }) => sum + trueValue, 0) === 0
+            voteOptions.reduce((sum, { trueValue = 0 }) => parseInt(sum) + trueValue, 0) === 0
```
chadoh: I think this is probably in the wrong place. We initialize `sum` to `0`, and then we add `trueValue` from each `voteOption`. I think we want to coerce `trueValue` to an int.

I just recently wrote this code, and I'd love ideas on how to make it less awkward. I don't like that we're doing this logic on every render pass; can you think of a more straightforward way to do this?

chadoh:
partial or full

@Quazia I noticed this behavior – if I vote with only 1 dot, it seems to _not_ use my full voting power (which I think makes sense), but the sliders still _say_ that I used my full voting power.

chadoh:
Fixed:

partial vote

PR #1264, opened 2019-09-27
package script improvements

topocount:
- add --no-cache flags in all remaining apps
- restore the use of clean:build while only keeping one -f flag instead of
two


package.json
```diff
@@ -25,8 +25,7 @@
     "build:script:projects": "lerna run build:script --scope=@tps/apps-projects",
     "build:script:rewards": "lerna run build:script --scope=@tps/apps-rewards",
     "clean:aragon": "rm -rf ~/.aragon && rm -rf ~/.ipfs",
-    "clean:build": "npm run clean -- --exclude=node_modules/",
-    "clean:commit": "rm -f package-lock.json && npm run bootstrap",
+    "clean:build": "git clean -fxd --exclude=node_modules/",
```
chadoh: Should this use `-fXd`?

ottodevs: We have 2 options her,  Chad's suggestion, or :
```suggestion
    "clean:build": "npm run clean -- --exclude=node_modules/",
```

 I would recommend my suggestion, because it centralizes the clean tags, to avoid using it wrong. The problem with lowercase -x is it will remove also user created files that are not tracked, becoming a dangerous script for the developer. That is how the script looked previously, what is the reason for replacing the old form with this new format @topocount ?

topocount: @ottodevs The problem with the capital "X" is that it disobeys the exclude param, rendering the function the same as the regular `npm run clean` script we have

topocount: https://github.com/AutarkLabs/open-enterprise/pull/1264#discussion_r329854509

ottodevs: I think found what we are looking for...
- It keeps untracked user created files (so I will not accidentally delete my new untracked `folderX` plenty of new features!)
- It removes everything else but `node_modules` so you can keep your installed modules while removing the generated cache and build garbage :)

```suggestion
    "clean:build": "npm run clean -- "-e !node_modules -e !**/node_modules/**/*"",
```

> Simply adding !node_modules and !**/node_modules/**/* to the exclude flag will keep that folder and node_modules on subfolders, as well at their content. It’s a double negative. You’re basically saying do not exclude node_modules.

The `\\!` negates the subject, that was the issue you probably were experiencing, it was a bit confusing to me to understand how it works.

Source:
https://git-scm.com/docs/gitignore#_pattern_format
https://git-scm.com/docs/git-clean#Documentation/git-clean.txt--eltpatterngt 

I submitted a commit to make the change, if you agree with that, and also Chad and the rest of people interested.

I also made the test by adding "-n" after a full `npm install`:

```sh
╰❱❱❱ npm run clean:build -- -n

> planning-suite-monorepo@ clean:build /Users/ottodev/box/projects/autark/open-enterprise
> npm run clean -- $npm_package_build_exclude_pattern "-n"


> planning-suite-monorepo@ clean /Users/ottodev/box/projects/autark/open-enterprise
> git clean -fXd "-e" "!node_modules" "-e" "!**/node_modules/**/*" "-n"

Would remove apps/discussions/package-lock.json
Would remove apps/projects/package-lock.json
```

ottodevs:
Another test, after running `npm run start:dev`:

```sh
╰❱❱❱ npm run clean:build -- -n

> planning-suite-monorepo@ clean:build /Users/ottodev/box/projects/autark/open-enterprise
> npm run clean -- $npm_package_build_exclude_pattern "-n"


> planning-suite-monorepo@ clean /Users/ottodev/box/projects/autark/open-enterprise
> git clean -fXd "-e" "!node_modules" "-e" "!**/node_modules/**/*" "-n"

Would remove apps/address-book/.cache/
Would remove apps/address-book/build/
Would remove apps/address-book/dist/
Would remove apps/allocations/.cache/
Would remove apps/allocations/dist/
Would remove apps/discussions/dist/
Would remove apps/discussions/package-lock.json
Would remove apps/dot-voting/.cache/
Would remove apps/dot-voting/dist/
Would remove apps/projects/.cache/
Would remove apps/projects/dist/
Would remove apps/projects/package-lock.json
Would remove apps/rewards/.cache/
Would remove apps/rewards/dist/
[...]
```

## No `node_modules` were harmed in the making of this test

PR #1267, opened 2019-09-30
Updated AddressBook radspec

stellarmagnet:


apps/address-book/contracts/AddressBook.sol
```diff
@@ -142,16 +142,16 @@ contract AddressBook is AragonApp {
...
      * @return contentId pointing to the IPFS structured content object for the entry
      */
     function getEntryIndex(address _addr) external view isInitialized entryExists(_addr) returns (uint256 index) {
         index = entries[_addr].index;
     }
 
     /**
-     * @notice Checks if an entry exists in the registry
-     * @param _entry the address to check
+     * @notice Checks if an `_entry` exists in the registry
```
topocount: should probably remove the "an" here, but there is no situation where this radspec will be evaluated, so that's kind of just semantics

PR #1222, opened 2019-09-05
Projects: update contracts for audits

chadoh:
This contains the updated contract for Projects, and all associated changes.

## Changes required for mainnet launch
- [x] updated event `BountyAdded` with new param `registryId`; update return values in event handler
- [x] new event `AssignmentRejected`
- [x] `initialize` no longer accepts `defaultToken`; in template, remove everything related to `defaultToken`
- [x] when running `start:dev`, as in `test.sh`, deploy `StandardBounties` contract to local blockchain
- [x] update `loadIssueData`; no longer returns `token` or `dataHash`
- [x] update `getSettings`; returns `bountiesRegistry` instead of `settings.bountyAllocator`
- [x] update `reviewSubmission`, now requires a sixth param, `tokenAmounts`
- [x] implement `submitWork`, which has been removed from `Projects.sol`, by using StandardBounties' `fulfillBounty` and listening for events from StandardBounties
- [x] removed `getSubmissionLength`, use StandardBounties `getBounty` instead
- [x] removed `getSubmission`, use StandardBounties `getBounty` instead. Inside the struct returned by `getBounty` there is a `fulfillments` array which contains all submissions & submitter info
- [x] when listening for events from StandardBounties, ignore all events for bountyIds that we don't care about 
- [x] update `addBounties`; input `tokenBounties` changed to `tokenTypes`

## Required for mainnet, but needs integration with some other in-progress work once it's all on dev
- [ ] ensure switching bounty token functions correctly
- [ ] remove hard-coded StandardBounties address from PlanningKit.sol

## Post launch improvements
- [ ] handle new event `BountyRemoved`
- [ ] handle new event `AwaitingSubmissions`
- [ ] update `getRepo`; use `openIssueCount` 
- [ ] use `updateBounty`
- [ ] use `removeBounties`
- [ ] use `addBountiesNoAssignment`
- [ ] unassign user when application or submission rejected, so same user can apply to work again

PR #1269, opened 2019-09-30
Web3/projects radspec

stellarmagnet:


apps/projects/contracts/Projects.sol
```diff
@@ -15,7 +15,7 @@ import "@aragon/os/contracts/lib/math/SafeMath.sol";
   */
 interface Bounties {
     /**
-     * @notice `_sender` will submit a fulfillment for bounty #`_bountyId` on Standard Bounties with the following info: `_data`
+     * @notice `_sender` submit a fulfillment for issue #`_bountyId` with the following info: `_data`
```
chadoh: > This transaction will eventually perform
> 
>   0xb4124cEB3451635DAcedd11767f004d8a28c6eE7 submit a fulfillment for issue #NNNN with the following info: ...
> 
> on 0xblahblahblah

The "_person_ submit a fulfillment" seems like a grammar issue.

apps/projects/contracts/Projects.sol
```diff
@@ -25,7 +25,7 @@ interface Bounties {
     ) external; //{}
 
     /**
-     * @notice `_sender` will update their fulfillment for `_bountyId` on Standard Bounties with the following info: `_data`
+     * @notice `_sender` updated their fulfillment for issue #`_bountyId` with the following info: `_data`
```
chadoh: Should this be past-tense? It's _about to happen_ at the time the user gets this message, right?

apps/projects/contracts/Projects.sol
```diff
@@ -359,8 +363,8 @@ contract Projects is AragonApp, DepositableStorage {
 // Repository functions
 ///////////////////////
     /**
-     * @notice Add repository to the Projects app
-     * @param _repoId id of the repo to be added
+     * @notice Add repository `_repoId`
```
chadoh: I don't think these repo IDs are human-readable, and I doubt it makes sense to put them in the radspec. I think they're just going to look sort of like ugly ethereum-address-style hashes.

apps/projects/contracts/Projects.sol
```diff
@@ -376,7 +380,7 @@ contract Projects is AragonApp, DepositableStorage {
     }
 
     /**
-     * @notice Remove repository from the Projects app
+     * @notice Remove repository `_repoId`
```
chadoh: Same thing here. Do we have some way to display the name of the repo at this point? Would we have to send extra data to the contract, which it wouldn't actually need to use?

apps/projects/contracts/Projects.sol
```diff
@@ -403,7 +407,7 @@ contract Projects is AragonApp, DepositableStorage {
 ///////////////////
 
     /**
-     * @notice Request assignment for issue `_issueNumber`
+     * @notice Request assignment for issue `_issueNumber` with application `_application`
```
chadoh: Is the IPFS hash of their application something the user will recognize (in which case it makes sense to show it here), or something we generate in the background?

apps/projects/contracts/Projects.sol
```diff
@@ -472,7 +476,7 @@ contract Projects is AragonApp, DepositableStorage {
     }
 
     /**
-     * @notice Review work for issue `_issueNumber`
+     * @notice `_approved ? 'Accept' : 'Reject'` work for issue `_issueNumber` with info: `_updatedSubmissionHash`
```
chadoh: Same question here

apps/projects/contracts/Projects.sol
```diff
@@ -673,7 +677,7 @@ contract Projects is AragonApp, DepositableStorage {
     }
 
     /**
-     * @notice Fund issues open to submissions from anyone: `_description`
+     * @notice Fund open submission issues: `_description`
```
chadoh: Why this change?

apps/projects/contracts/Projects.sol
stellarmagnet: Woops. Should be 
> `_sender` submitted a fulfillment for issue #`_bountyId` with the following info: `_data`

apps/projects/contracts/Projects.sol
stellarmagnet: Yeah, these are actually user facing and wont be forwarded.

Maybe it should just be
> Update fulfillment for issue #`_bountyId` with the following info: `_data`

apps/projects/contracts/Projects.sol
stellarmagnet: Maybe should be:

> Submitted a fulfillment for issue #`_bountyId` with the following info: `_data`

apps/projects/contracts/Projects.sol
stellarmagnet: Yeah but there is no other context for the voting app "Add repository" doesn't convey anything. I'm not sure how we can properly solve this, but it's a UX concern if we want people to be able to vote to add or remove a repo. I was hoping maybe we can provide with some kind of tool to look up the repo name via Github app.

apps/projects/contracts/Projects.sol
stellarmagnet: I was asking kevin if we can add metadata for the repo @topocount - but if it's IPFS we still have ugly hash until we develop an IPFS radspec resolver

apps/projects/contracts/Projects.sol
stellarmagnet: We may have to just put it in the docs so users know it's an ipfs hash. Ideally the client should be able to determine an IPFS hash based on it starting with a Q and it being a certain number of characters, and generate a link that is tied to the client's gateway

stellarmagnet: For now giving something is better than the users having no data at all in the voting app? i think?

apps/projects/contracts/Projects.sol
stellarmagnet: We are calling it "open submission" in projects UI, and it shortens the text.

chadoh: Would a hyphen make sense? "Fund open-submission issues"?

apps/projects/contracts/Projects.sol
topocount: I've decided to add a metadata field here, which we can presumable pass the repo name into from the github repo object we have in the frontend

stellarmagnet:
This is great but we will need an updateRepoMetadata action as well so
users can sync new repo names if they change and can also have things like
images associated w the repo. In future when we decentralize this can
manage everything around it like descriptions and other things

chadoh:
Closing; I believe @topocount incorporated these changes directly into #1222 already. Kevin, re-open this PR and correct me if I'm wrong!

PR #1273, opened 2019-10-01
Projects/script: don't update state unnecessarily

PeterMPhillips:
This improves the performance of our app in general, but is crucial now
that we listen to all events from the StandardBounties contract. We want
to ignore all of these events, leaving state unchanged.

We could improve this even further, as shown in the `github` handlers.
We only need to mutate the specific branch of our state tree that
requires updating, and then React's algorithm that calculates whether it
needs to render only gets calculated for the components that make use of
that specific branch. However, this gets us past the bottleneck
introduced by full-state-rewrites-on-every-event-from-StandardBounties
bottleneck, and that's the main improvement we need right now.

### DIFF
--- a/apps/projects/app/store/events.js
+++ b/apps/projects/app/store/events.js
@@ -39,7 +39,6 @@ import { app } from './app'
 
 export const handleEvent = async (state, action, vaultAddress, vaultContract) => {
   const { event, returnValues, address } = action
-  let nextState = { ...state }
 
   switch (event) {
   case REQUESTING_GITHUB_TOKEN: {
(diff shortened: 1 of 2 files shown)


apps/projects/app/store/events.js
```diff
@@ -48,96 +47,86 @@ export const handleEvent = async (state, action, vaultAddress, vaultContract) =>
...
+    return state
   }
   case REQUESTED_GITHUB_TOKEN_FAILURE: {
     return state
   }
   case REQUESTED_GITHUB_DISCONNECT: {
-    const { github } = INITIAL_STATE
-    nextState = { ...state, github }
-    return nextState
+    state.github = INITIAL_STATE
```
chadoh: Won't this create something like this?

```js
{
  repos: [],
  tokens: [],
  issues: [],
  bountySettings: {},
  github: {
    repos: [],
    tokens: [],
    issues: [],
    bountySettings: {},
    github: { status: STATUS.INITIAL, token: null, event: '' }
  }
}

chadoh:
@ottodevs 

> Hi @PeterMPhillips , thanks for this work, could you elaborate a bit on what issues is this PR solving and why we need these changes?

This would have been a good improvement to make regardless, but the specific reason we absolutely need it now is because we are listening to all events from the StandardBounties contract. If we update the state _on every single event_, whether we change anything or not, it's going to result in way too many unnecessary re-renders.

@PeterMPhillips this would be good info to add to the description!

Issue #1265, opened 2019-09-27
Address Book Tweaks


stellarmagnet:
- [x] On Address Book, right-align the Type column so it's closer to the context menu

![image.png](https://images.zenhubusercontent.com/5ab2c6f24b5806bc2bcbd667/4d2afb3f-fe29-43d2-a6f5-8920e05ac8e0)

Look at Tokens for inspiration -
![image.png](https://images.zenhubusercontent.com/5ab2c6f24b5806bc2bcbd667/53efc936-27ca-4cff-adf5-102515ba8819)

- [x] `force` the address for the LocalIdentityBadge (the custom label should not display)
![image.png](https://images.zenhubusercontent.com/5ab2c6f24b5806bc2bcbd667/61c2058a-a2a8-4acf-ad3c-42f0321fa868)

lcoenen:
The first one looks like a bug in `aragon-ui`. The difference between the Token Manager view and the Address Book is the last `<td>` element width. In the Token view, it's set to `1px`, while it's set to 0 in the Address Book. 

![image](https://user-images.githubusercontent.com/1237971/65893055-3df0b080-e39f-11e9-8b86-663a758366b6.png)
![image](https://user-images.githubusercontent.com/1237971/65893105-5234ad80-e39f-11e9-9a3d-19af13148c6c.png)

Indeed, setting the width to `1px` fixes the issue. It seems this property is set in [here](https://github.com/aragon/aragon-ui/blob/newstyle/src/components/DataView/TableView.js), lines 302-... (in the `<EntryRow>` component)

```
              width: ${compact
                ? '1px' // For some reason Blink tends to make 0 grow but not 1px
                : 'auto'};
```

The `compact` property is supposed to be set to `true` by `cellsFromEntry()` used in `Entry`, if it receive either `hasAnyExpensions` (which should be our case here - [see this line](https://github.com/aragon/aragon-ui/blob/2e57e6986f87d7e8182a00119ca0b372ee597812/src/components/DataView/TableView.js#L67) )

e18r:
@lcoenen Thank you for the explanation. Do you believe it can be fixed by means of a workaround? Also, we're not using the very last version of AragonUI at the moment. Does this get solved in the latest version?

lcoenen:
I'll checkout today what's going on exactly. Maybe we can just set the last
td of that specific page - it would be a hack. I'll let you know as soon as
I have more infos.

lcoenen:
[...]

PR #1274, opened 2019-10-01
Implement `forceAddress` in LocalIdentityBadge #1265

lcoenen:
Address books `App/Entities.js` now use `forceAddress` property with
<LocalIdentityBadge />.

The implementation have been copied from [Aragon IdentityBadge component](
https://github.com/aragon/aragon/blob/d754535d8487b94a651209464225d8507ae24a67/src/components/IdentityBadge/LocalIdentityBadge.js#L15)

See https://youtu.be/oxeWHqhoSUo for a demo

image
image

### DIFF
--- a/apps/address-book/app/components/App/Entities.js
+++ b/apps/address-book/app/components/App/Entities.js
@@ -46,6 +46,8 @@ const Entities = ({ entities, onRemoveEntity }) => {
           <LocalIdentityBadge
             key={entryAddress}
             entity={entryAddress}
+            shorten={true}
+            forceAddress
           />,
           <Tag
             key={entryAddress}
--- a/shared/identity/LocalIdentityBadge.js
+++ b/shared/identity/LocalIdentityBadge.js
@@ -36,12 +36,12 @@ function useIdentity(address) {
   return [ name, handleShowLocalIdentityModal ]
 }
 
-const LocalIdentityBadge = ({ entity, ...props }) => {
+const LocalIdentityBadge = ({ entity, forceAddress, ...props }) => {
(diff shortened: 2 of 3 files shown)


e18r:
@lcoenen Thank you for your contribution. I'm afraid that the version upgrade of AragonUI affected precisely LocalIdentityBadge. After that PR gets merged, Address Book will be using its own LocalIdentityBadge (not the shared one). Could you please apply your changes to this new version? You can rebase your branch with `address-book-aragon-ui-version-bump`. Thank you!

lcoenen:
This is broken, will fix tomorrow

shared/identity/LocalIdentityBadge.js
```diff
@@ -65,6 +65,7 @@ const LocalIdentityBadge = ({ entity, ...props }) => {
 
 LocalIdentityBadge.propTypes = {
   entity: PropTypes.string.isRequired,
+	forceAddress: PropTypes.bool
```
chadoh: indentation seems off

apps/address-book/app/components/App/Entities.js
```diff
@@ -46,6 +46,8 @@ const Entities = ({ entities, onRemoveEntity }) => {
           <LocalIdentityBadge
             key={entryAddress}
             entity={entryAddress}
+            shorten={true}
```
chadoh: We don't need the `={true}` here. If the prop `shorten` is given without a value, React takes the `true` as implicit.

chadoh:
Thanks for the good work, @lcoenen! I've merged without those changes; I'll update the code myself in a separate PR.

PR #1278, opened 2019-10-03
Request for help: Reputation template

ottodevs:
I want you guys to review this and work together into the integration because it is not really clear how to deal with these contracts.

For now I just added the contracts to the templates folder, but this needs more work to be something usable, and it is breaking the current template now (tests don't pass)

- Is WhitelistOracle a separate AragonApp? 
- In case it is (as it seems so), how will we deploy it, as a system app? frontend app? 
- How does it interact with TokenManager? Do we really need both installed to an DAO?
- What is the best approach to import it as node module while keeping a regular token-manager app into the template?

NOTE: I recommend you to go through each commit to see what I was progressively fixing, since the contracts here didn't work for me: https://github.com/AutarkLabs/aragon-apps/commit/7c192279c4348ce4138026155aae0a8e6bbeca1c#diff-2a9b6fae0c66f35810c66745c35ed2e3 (The link was taken from the hackmd templates specs document)


ottodevs:
Closing this as superseded by: https://github.com/AutarkLabs/open-enterprise/pull/1597

PR #1279, opened 2019-10-03
Projects: Initialize Jest; stub first test

chadoh:
Initial stab at adding testing to our reducer. Got held up on mocking calls to our contracts.

To use:

* `cd apps/projects`
* `npm run test:app` or `npm run test:app:watch`

This changes our `.babelrc` and removes `modules: false`. Jest requires that we avoid this setting. `npm run build:app` still works, which means this setting may have been here for no reason.

### DIFF
--- a/apps/projects/.babelrc
+++ b/apps/projects/.babelrc
@@ -1,6 +1,6 @@
 {
   "presets": [
-    ["@babel/preset-env", { "modules": false, "useBuiltIns": false }]
+    ["@babel/preset-env", { "useBuiltIns": false }]
   ],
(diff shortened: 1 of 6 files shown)


stellarmagnet:
@chadoh is this PR still needed? If yes, it's a draft - what is remaining?

chadoh:
The idea was to test the reducer in isolation, which would be much faster than full e2e tests. However, this ended up being more complex than initially expected, since we were going to need to stub more and more calls to the contract. I think we should focus on the testing that Kevin is adding in #1677 instead.

PR #1289, opened 2019-10-03
Projects: support funding issues with DAO token

chadoh:
If you go to Projects -> Settings and choose to fund bounties with your DAO-specific token, this updates the FundIssues logic to actually support that choice.

one issue funded with ETH, one with Autark


chadoh:
@topocount do you know of a better way to do this?

PR #1290, opened 2019-10-03
Fixed require statement in setDistribution and improved testing coverage

PeterMPhillips:
Also did `npm link` on the tps-helpers package which may have caused the errors @topocount was seeing


apps/allocations/package.json
```diff
@@ -51,6 +51,7 @@
     "@babel/plugin-proposal-class-properties": "^7.4.4",
     "@babel/preset-env": "^7.4.5",
     "@babel/preset-react": "^7.0.0",
+    "@tps/test-helpers": "^0.0.1",
```
chadoh: `@tps/test-helpers` is not a real repository. For now, we're requiring files explicitly from the `PROJECT_ROOT/shared` directory, using terrible file paths like `import X from '../../../../../../shared/utils'`

PeterMPhillips: Right, I was getting this error without it `Error: Could not find @tps/test-helpers/contracts/factory/EVMScriptRegistryFactory.sol from any sources; imported from .../open-enterprise/apps/allocations/test/Spoof.sol`. The test references `@tps/test-helpers/artifacts.js` to import a few contracts. I added `@tps/test-helpers` to package.json and then called `npm link ../shared/test-helpers`. Don't know if that is how we want to do things but it solved the problem. Also, it seems the only real blocker is the StandardBounties contract, everything else can be imported from `@aragon` packages. We could include a StandardBounties package, it looks like there is one available on npm: https://www.npmjs.com/package/bounties

PeterMPhillips: @chadoh I just attempted to replace references to `@tps` with relative paths e.g. `../../shared` which works fine until I try to run a coverage test. Because solidity-coverage copies the files over to `coverageEnv` the relative paths are no longer correct

ottodevs:
Just a note: `web3/allocations` is already merged, please don't target that branch again.
I change this to target `dev` instead

apps/allocations/test/allocations.test.js
```diff
@@ -4,7 +4,8 @@ const {
   DAOFactory,
   EVMScriptRegistryFactory,
   Kernel,
-  MiniMeToken
+  MiniMeToken,
+  StandardBounties
```
ottodevs: Let's remove this unneeded import, allocations is not going to use StandardBountes, please, look at the correct `Allocations.test.js` file in `dev` branch and rebase from it before messing up with further changes

apps/shared/test-helpers/artifacts.js
```diff
@@ -1,10 +1,11 @@
...
   DAOFactory: artifacts.require('./factory/DAOFactory'),
   EVMScriptRegistryFactory: artifacts.require(
-    './factory/EVMScriptRegistryFactory'
+    './contracts/factory/EVMScriptRegistryFactory'
   ),
-  Kernel: artifacts.require('./kernel/Kernel'),
-  MiniMeToken: artifacts.require('./lib/minime/MiniMeToken'),
+  Kernel: artifacts.require('./contracts/kernel/Kernel'),
+  MiniMeToken: artifacts.require('./contracts/lib/minime/MiniMeToken'),
+  StandardBounties: artifacts.require('./contracts/lib/bounties/StandardBounties'),
```
ottodevs: Don't edit this file without asking first, because we are trying to get rid of this package, these changes are not needed and introduce regressions into the whole repository

apps/allocations/package.json
chadoh: Hmm, I'm not sure how to do this with contracts. @topocount @ottodevs can you advise?

I don't think `npm link` will work, as that is local to your machine.

ottodevs: I commented on this, please let me know if something is not clear

ottodevs:
I close here in favour of: https://github.com/AutarkLabs/open-enterprise/pull/1282
Please, move any work from this PR not present on the above, if there is still something relevant here.

Good work overall! I am sorry but things were a bit messed with the different branches.

PR #1292, opened 2019-10-04
Web3/allocations

topocount:
This is an initial integration of web3/allocations.

Remaining tasks:

- [ ] integrating the remaining budget gauge in the Budget cards
- [ ] hooking up the status in allocations history to update when an `_runPayout` is executed
- [ ] Getting the budget to render correctly in the allocations history (very easy win)
- [ ] Exposing the runPayout Button for allocations that haven't executed every available time yet


apps/allocations/app/components/App/AllocationsHistory.js
```diff
@@ -12,9 +13,18 @@ import PropTypes from 'prop-types'
...
 import { displayCurrency } from '../../utils/helpers'
+import { match } from 'minimatch'
 
 const AllocationsHistory = ({ allocations }) => {
   const theme = useTheme()
+  const { balances = [] } = useAppState()
+  const getTokenSymbol = inputAddress => balances.reduce((matchingSymbol,
+    { symbol, address }) => {
+    if (inputAddress === address) return symbol 
+    else return matchingSymbol
```
chadoh: ahhh I'm breaking the policy of "re: style comments, add a linting rule or gtfo", but breaking up those args onto multiple lines makes my eyes cross. How about this:

```js
const getTokenSymbol = inputAddress => balances.reduce(
  (matchingSymbol, { symbol, address }) => inputAddress === address ? symbol : matchingSymbol,
  ''
)
```

Wait, now that I can see it better, isn't it more straightforward to use `find`?

```js
const getTokenSymbol = inputAddress => {
  const matchingBalance = balances.find(({ address }) => inputAddress === address)
  return matchingBalance ? matchingBalance.symbol : ''
}

File: apps/allocations/app/components/App/AllocationsHistory.js [hunk 2]
```diff
@@ -36,7 +46,7 @@ const AllocationsHistory = ({ allocations }) => {
       renderEntry={({
         date,
         budget,
-        recipients,
+        recipients = [ 'test', 'hi' ],
```
chadoh: We probably shouldn't commit this 😄

apps/allocations/app/components/App/AllocationsHistory.js
```diff
@@ -48,51 +58,51 @@ const AllocationsHistory = ({ allocations }) => {
           recipients.length === 1 ? '1 entity'
             : recipients.length + ' entities',
           description,
-          <Status key={index} code={status} />,
+          <Status key={index} code={2 } />,
```
chadoh: more style comments i can't stop myself the mismatched spacing here 😬

apps/allocations/app/components/App/AllocationsHistory.js
```diff
@@ -48,51 +58,51 @@ const AllocationsHistory = ({ allocations }) => {
...
+      //          fontSize: '12px',
+      //        }}>
+      //          { displayCurrency(BigNumber(recipient.amount)) } {' '}
+      //          {token} {' • '}
+      //          { allocated.times(100).dp(0).toNumber() }{'%'}
+      //        </div>
+      //      </div>
+      //    )
+      //  })
+      //}}
```
chadoh: Can we delete this? Is it enough to have it in the commit history?

apps/allocations/app/components/App/App.js
```diff
@@ -38,7 +38,16 @@ const App = () => {
...
+    console.log('budgetId: ', budgetId, ' balance: ', balance)
+    console.table({
+      addresses,
+      description,
+      budgetId,
+      recurring,
+      period,
+      balance,
+      tokenAddress,
+    })
```
chadoh: Should we leave this in? Should we add `eslint-disable-line` statements, and a comment explaining why we want this in shipped code for now?

apps/allocations/app/components/App/App.js
```diff
@@ -90,15 +99,15 @@ const App = () => {
     })
   }
 
-  const onNewAllocation = (address, description, id, balance) => {
+  const onNewAllocation = (id, description, balance, token) => {
+    console.log('budgetId: ', id)
```
chadoh: Same question here

apps/allocations/app/components/App/App.js
```diff
@@ -90,15 +99,15 @@ const App = () => {
...
     setPanel({
       content: NewAllocation,
       data: {
         heading: 'New Allocation',
         subHeading: description,
-        address,
         balance,
-        balances,
+        balances: [token],
```
chadoh: The confusion of the word "budget" and "token" here seems like something we should talk about at some point. This seems like it might indicate that it's about time to reshape our state tree and rename some of its attributes. But we can have that conversation later.

apps/allocations/app/components/Card/Budget.js
```diff
@@ -1,6 +1,7 @@
 import PropTypes from 'prop-types'
 import React from 'react'
 import styled from 'styled-components'
+import BN from 'bignumber.js'
```
chadoh: This seems dishonest. If we call it `BN`, let's `import { BN } from 'web3-utils'`. Otherwise, let's call it `BigNumber`

apps/allocations/app/components/Card/Budget.js
```diff
@@ -48,7 +49,9 @@ const Budget = ({
   const reactivate = () => {
     onReactivate(id)
   }
-
+  const tokenAmount = BN(amount).div(BN(10).pow(token.decimals))
+  const tokensRemaining = tokenAmount.minus(BN(allocated).div(BN(10).pow(token.decimals)))
+  console.log('budget token: ', tokensRemaining.toString())
```
chadoh: Do you want to merge this `console.log` statement?

apps/allocations/app/store/allocation.js
```diff
@@ -9,7 +9,7 @@ import { app } from '../../../../shared/store-utils'
 
 export const updateAllocations = async (allocations, { accountId, payoutId }) => {
   const newAllocations = Array.from(allocations || [])
-
+  console.log('payout info: ', await app.call('getPayout', accountId, payoutId).pipe(first()).toPromise())
```
chadoh: Same question here.

apps/allocations/app/store/allocation.js
```diff
@@ -73,24 +73,28 @@ const getAllocation = async ({ accountId, payoutId }) => {
...
         distSet,
         period,
         recurrences,
-        startTime,
-        token,
-      }) => ({
+        startTime
+      },
+      description
+      ]) => ({
```
chadoh: 😖 

This is _really_ hard to understand. I think some indentation cleanup might be all it takes to clarify.

apps/allocations/app/store/events.js
```diff
@@ -38,11 +38,11 @@ const eventHandler = async eventData => {
       ...state,
       accounts: await updateAccounts(state.accounts, returnValues.accountId),
     }
-  // case 'SetDistribution':
-  //   return {
-  //     ...state,
-  //     allocations: await updateAllocations(state.allocations, returnValues)
-  //   }
```
chadoh: This seems like a comment we can definitely get rid of.

apps/allocations/contracts/Allocations.sol
```diff
@@ -438,6 +441,28 @@ contract Allocations is AragonApp {
         }
     }
 
+    /**
+     * @param _accountId The account ID of the budget to be calculated
+     */
+    function getRemainingBudget(uint64 _accountId) public view returns (uint256) {
+        Account storage account = accounts[_accountId];
+        if (!account.hasBudget) {
+            return MAX_UINT256;
```
chadoh: If the account has no budget, it has the maximum budget possible?

apps/allocations/contracts/Allocations.sol
```diff
@@ -520,26 +545,7 @@ contract Allocations is AragonApp {
...
-
-        // A budget decrease can cause the spent amount to be greater than period budget
-        // If so, return 0 to not allow more spending during period
-        if (spent >= budget) {
-            return 0;
-        }
-
-        // We're already protected from the overflow above
-        return budget - spent;
+        return getRemainingBudget(_accountId) >= _amount && vault.balance(account.token) >= _amount && _amount > 0;
```
chadoh: We removed the `_` from the beginning. Why's that?

apps/allocations/package.json
```diff
@@ -35,6 +35,7 @@
     "@aragon/apps-vault": "4.1.0",
     "@aragon/os": "4.2.0",
     "@aragon/ui": "1.0.0-alpha.19",
+    "@tps/test-helpers": "0.0.1",
```
chadoh: Do we want this? See [discussion here](https://github.com/AutarkLabs/open-enterprise/pull/1290#discussion_r331277577).

apps/allocations/contracts/Allocations.sol
topocount: It's a public function now

topocount: ah wait I'll just make a public wrapper around this private function. Thanks for calling this out

apps/allocations/contracts/Allocations.sol
topocount: Yeah no budget means that the only limitation is an overflow. You might be thinking of a disabled budget, where  hasBudget is set to `true` and the budget value itself is set to 0

apps/allocations/app/components/App/AllocationsHistory.js
topocount: This will be implemented pretty soon in the future, but possibly not for the initial mainnet release. It's not uncommon for A1 to leave pending stuff commented out in their prod code. Not saying we should follow it on that principle, but I think it has it's uses

PR #1293, opened 2019-10-04
Final Rewards Tweaks

Quazia:
Fix time issues with Rewards, add metrics for Rewards. Still need to plug in the claim rewards button.


apps/rewards/app/app-state-reducer.js
```diff
@@ -20,16 +21,17 @@ function appStateReducer(state) {
...
     state.rewards = state.rewards  || []
     state.claims = state.claims  || []
     state.rewards = state.rewards.map(reward => {
       if(reward.isMerit){
         reward.rewardType = ONE_TIME_MERIT
         reward.dateReference = new Date()
-      } else if (reward.occurances === 1){
+      } else if (reward.occurances.toString() === '1'){
         reward.rewardType = ONE_TIME_DIVIDEND
+        reward.dateReference = new Date(reward.endDate)
```
chadoh: Did you check this and verify that `reward.endDate` will be in the correct format? Do we need to multiply by 1000?

apps/rewards/app/app-state-reducer.js
```diff
@@ -39,8 +41,9 @@ function appStateReducer(state) {
       reward.amountToken = amountToken.symbol
       return reward
     })
-    state.myRewards = state.myRewards || []
-    let metric = calculateAverageRewardsNumbers(state.rewards, state.claims, state.balances, updateConvertedRates(state.balances))
+    state.myRewards = state.rewards.filter(reward => reward.userRewardAmount > 0)
+    let convertRates = updateConvertedRates(state.balances)
+    let metric = calculateAverageRewardsNumbers(state.rewards, state.claims, state.balances, convertRates)
```
chadoh: Why `let` instead of `const`?

apps/rewards/app/app-state-reducer.js
```diff
@@ -58,8 +61,26 @@ function appStateReducer(state) {
...
+        unit: 'USD',
+      },
+      {
+        name: 'All time rewards obtained',
+        value: myMetric[2].toString(),
+        unit: 'USD',
+      },
+    ]
     state.amountTokens = state.amountTokens || []
+    console.log('end recuders:', state)
```
chadoh: Do you think it's useful to leave this in for now? Do we need an `eslint-disable-line` comment?

apps/rewards/app/components/App/App.js
```diff
@@ -161,6 +162,7 @@ class App extends React.Component {
     const tenBN =  new BN(10)
     const decimalsBN = new BN(reward.amountToken.decimals)
     reward.amount = amountBN.mul(tenBN.pow(decimalsBN))
+    console.log(reward)
```
chadoh: Same question here.

chadoh:
Looks like there are some linting errors introduced here. These may have resulted from the rebase with `dev`.

apps/rewards/app/app-state-reducer.js
Quazia: This is correct, no need to multiply it by 1000.

apps/rewards/app/app-state-reducer.js
Quazia: Bad habit, this can be changed I'll switch it.

PR #1295, opened 2019-10-04
Address linting issues

chadoh:
Our linter(s) must have been misconfigured for a while, allowing various errors to get into the codebase. This addresses them all.


chadoh:
This did fix linting errors, which revealed that some Rewards tests are failing. Let's fix those in a separate PR.

PR #1298, opened 2019-10-05
NewAllocation Panel Enhancements 

topocount:
...and integration of the Payout Executions
... and merging in the final form of the dot voting contract fix


apps/allocations/app/components/App/AllocationsHistory.js
```diff
@@ -52,51 +54,36 @@ const AllocationsHistory = ({ allocations }) => {
           <div key={index}>
             {'# ' + accountId}
           </div>,
-          recipients.length === 1 ? '1 entity'
+          recipients.length === '1' ? '1 entity'
```
chadoh: 🧐

Is `recipients` an array? Why are we comparing `length` to a string?

apps/allocations/app/components/Card/Budget.js
```diff
@@ -90,40 +91,38 @@ const Budget = ({
...
-              color: theme.contentSecondary,
-              paddingTop: '4px',
-            }}
-          >
-            <Text>{`${tokensRemaining.div(tokenAmount).times(100)}% remaining`}</Text>
-          </StatsValueSmall>
-        </StyledStats>
-      </StatsContainer>
-    </StyledCard>
+        </React.Fragment>
```
chadoh: Looks like `ContextMenuItem` is the only child of `React.Fragment` here, which makes `React.Fragment` unnecessary.

apps/allocations/app/components/Card/Budget.js
```diff
@@ -90,40 +91,38 @@ const Budget = ({
...
+      }
+    >
+      <StatsValueBig theme={theme}>
+        {displayCurrency(BigNumber(amount))}
+        <Text>{' ' + token.symbol + ' per period'}</Text>
+      </StatsValueBig>
+      <StatsValueBig css={{ paddingTop: '24px' }} theme={theme}>
+        <ProgressBar
+          color={String(theme.accentEnd)}
```
chadoh: Do we need to coerce `theme.accentEnd` to string?

apps/allocations/app/store/account.js
```diff
@@ -2,17 +2,22 @@
...
 
 import { app } from '../../../../shared/store-utils'
+import { combineLatest } from 'rxjs'
 
 /// /////////////////////////////////////
 /*    Allocations event handlers      */
 /// /////////////////////////////////////
 
 export const updateAccounts = async (accounts, id) => {
+  //Array.prototype.findIndex()
```
chadoh: Is it useful to keep this comment here?

apps/allocations/app/store/allocation.js
```diff
@@ -101,13 +107,27 @@ const getAllocation = async ({ accountId, payoutId }) => {
         recurrences,
         token,
         date: new Date(startTime*1000),
-        recipients: { length: Number(recipientsLength) },
-        status: 2 //Approved will be made dynamic once the execution handler is integrated
+        recipients,
+        status: recipients.every(r => r.executions === recurrences) ? 3 : 2, //Approved will be made dynamic once the execution handler is integrated
```
chadoh: Can we remove this comment now?

Also, it would be great to assign these numbers to some constants, so we can give them meaningful names.

apps/dot-voting/app/store/votes.js
```diff
@@ -91,24 +93,28 @@ const loadVoteDataAllocation = async (vote, voteId) => {
...
-        // const symbol
-        // const tokenAddress = '0x' + vote.executionScript.slice(794, 834)
-        // if (tokenAddress === ETHER_TOKEN_FAKE_ADDRESS) {
-        //   symbol = 'ETH'
-        // }
-        // else {
-        //   symbol = await getTokenSymbol(app, tokenAddress)
-        // }
+        const allocationsAddress = '0x' + vote.executionScript.slice(10,50)
+        const allocationsAccountId = parseInt(vote.executionScript.slice(514, 578), 16).toString() //should be 514, 578
```
chadoh: The comment seems to repeat the code itself. An explanation of what these numbers mean might be nice. Better yet, rather than explain them with a comment, explain them by declaring them as well-named constants.

PR #1300, opened 2019-10-08
Create Authenticator app; update Projects

chadoh:
We currently have a React app deployed at `https://tps.autark.xyz/tps-auth` which does basically what this new Authenticator does. Here's how it works:

1. Projects opens a popup which loads the GitHub OAuth login page
2. We have [a GitHub OAuth App](https://github.com/organizations/AutarkLabs/settings/applications/953918), where we provide an Authorization Callback URL. This ~is~ was configured to that `tps.autark.xyz/tps-auth` page.
3. This `/tps-auth` page receives a `code` from GitHub, which it posts back to Projects using `opener.postMessage`
4. Projects uses this `code` to hit `https://tps.autark.xyz/authenticate`, which holds onto the private key for our GitHub OAuth App and can thus load an API token for the user, so Projects can start making requests as this user.

This PR seeks to optimize three aspects of the current approach:

1. Code visibility/maintainability: rather than having a deployed app with seemingly no corresponding GitHub repository, this PR adds it right to our `open-enterprise` repo. Authenticator is a key piece of our infrastructure and deserves to be treated as such. This PR also simplifies Authenticator, using a simple html page rather than a complex React app.
2. Design: rather than a sans-serif "loading" message, this PR adds a simple loading animation while the user is logged in, with a subtle technical status.
3. Security: loading Authenticator at a URL we own provides no additional security. Why? Because any app, not just Projects, could open a GitHub OAuth page and provide our CLIENT_ID. GitHub would then redirect back to our Authenticator, regardless of where it lives, and (with the implementation at tps.autark.xyz), Authenticator blindly passes the `code` it received back to `opener`. From there, a malicious app can easily POST that code to our `https://tps.autark.xyz/authenticate` endpoint and receive a token for the user. And then that malicious app could post comments as them and update their code, and it would look like these GitHub actions were mediated by us.

## Unfortunately, the security goal could not be reached

Read the commit messages for a log of the three attempted approaches and why they don't work.

In the fourth commit, this PR implements a solution that's very similar to the app that's deployed at `tps.autark.xyz/tps-auth` – it still reaches the first two goals stated above, but all security checks have been removed. This Authenticator will blindly pass the `code` it receives back to the page that opened it. 

## Let's think through the exact vulnerability here

We are making it super easy for any ~app developer~ malicious hacker to create an app (an Aragon app or any other kind) that opens our GitHub login page. Let's call our hypothetical malicious hacker Paige. Paige creates an app that borrows our code for opening a GitHub login page; [see such an example app here](https://github.com/chadoh/open-enterprise-oauth-hack). When the users of her app click to log in to GitHub, this is what they see:

Open Enterprise by AutarkLabs wants to access your public repositories

Paige is now stealing the trust that these users have in AutarkLabs. Once they grant her app permission, via our OAuth App, her app can now read & write to their GitHub repositories. Maybe she posts spam or defamatory comments on behalf of these users, or maybe she does something more subtle: maybe she commits code to their repositories that compromises the security of their applications.

If a user suspects something and they check the OAuth apps they've authorized in their GitHub settings, they'll see only our app:

You have granted 22 applications access to your account

If we were able to patch the security hole here, Paige could still do all of these things, but she could not use our name to do it. If she wanted to borrow our name, she could make a GitHub OAuth app that looks similar, but when the user went to their GitHub settings, they would see two similar apps:

You have granted 23 applications access to your account

## Where do we go from here?

For now, given our small user base, we probably don't need to worry too much about the Paiges of the world wanting to borrow our brand to give their schemes more believability. Maybe we can tolerate the risk that comes with shipping this vulnerable version of Authenticator until we can implement a more robust solution. 

Is it worth mentioning the risk to our users, when they sign into GitHub within Projects?

NOTE: Autark will never ask for your GitHub credentials outside of this Projects app. If another app posing to be from AutarkLabs asks you to sign into GitHub, don't trust it!

Maybe? I'm not sure. The risk seems low, and the messaging seems awkward, so I'm leaning toward having no warning, for now.

Long term, how do we patch this vulnerability? I can think of a couple possibilities:

1. Have the user sign a request that this Authenticator reads, which (somehow) cryptographically proves that the user is coming from a legitimate version of the Projects app. Downsides: how exactly this would work remains to be worked out, and it seems like it would result in a suboptimal user experience.

2. Stop doing the OAuth handshake directly from Projects; do it from the Aragon client instead. Just as the Aragon client is responsible for managing a user's on-chain identity, this approach would also make it responsible for managing a user's Web2 identities.

   Global Settings: OAuth Authorizations

   Aragon Apps like Projects would then request access to a user's GitHub credentials via aragonAPI. If the user has not yet given Aragon access to their GitHub, the Aragon client would kick off that handshake. If they've already granted access to Aragon but not Projects, they would be asked if they want Projects to have access to their GitHub credentials.

   This option seems like it fits better with the security model of Aragon in general. 

   Downsides: it's a big effort, with a long timeline. Is this in scope, for the Aragon client?

Please, offer other suggestions!


apps/authenticator/public/index.html
```diff
@@ -0,0 +1,93 @@
...
+          document.getElementById('status').innerText = 'cannot authenticate'
+        }, 500)
+      }
+
+      var code = new URLSearchParams(window.location.search).get('code')
+      console.log('code: ' + code)
+
+      // assuming window.opener is the Projects app, `pathname` will look like `/ipfs/Qmblahblahblah`
+      // we need to check that the `Qmblablahblah` part is a known version of Projects
+      var callingApp = window.opener && window.opener.location.pathname.split('/')[2]
```
chadoh: This line doesn't work with `localhost` due to same-origin restrictions.

chadoh: Also doesn't work when `opener` refers to an app in an iframe, because the iframe's origin is hidden from the popup. If the Aragon client included `allow-same-origin` on the iframe's `sandbox` property, we might be able to get around this, but that comes with significant security implications and is [not under consideration](https://github.com/aragon/aragon/blob/fcab70b3a30d6e5a4e33c2382f6cbc4c917da31a/src/security/configuration.js#L22).

sohkai:
> Stop doing the OAuth handshake directly from Projects; do it from the Aragon client instead. Just as the Aragon client is responsible for managing a user's on-chain identity, this approach would also make it responsible for managing a user's Web2 identities.

> This option seems like it fits better with the security model of Aragon in general.

The main concern I have is how we'd control the OAuth permissions model between installed Aragon Apps. If you authenticate via the Aragon client, would any app have permission to use that OAuth credential? That would definitely be bad.

If we assume the storage model in the Aragon client is tight*, in that no apps can read data from other apps or the client itself, could we then build robust and secure APIs around whether an app could access a certain credential or not?

*This is difficult to really say, but the easiest way around it would be to encrypt this credentials store via a user-given password or just their ETH private key

chadoh:
@sohkai 

> The main concern I have is how we'd control the OAuth permissions model between installed Aragon Apps. If you authenticate via the Aragon client, would any app have permission to use that OAuth credential? That would definitely be bad.

I don't think that makes sense, no. 

> If we assume the storage model in the Aragon client is tight*, in that no apps can read data from other apps or the client itself, could we then build robust and secure APIs around whether an app could access a certain credential or not?

I was picturing a minimal-viable version working like this, from the user perspective: Say I've given the Aragon client access to my GitHub credentials, but no specific app yet. Now I want to use Projects. Projects could ask, via aragonAPI, for access to my GitHub credentials, and the Aragon client would ask me if I approve. Sort of like how location or microphone access works for a web app in the browser. "Projects would like to access your GitHub credentials. Yes or no?" If I then use a different app which also wants access to my GitHub credentials, that app would also need to ask for permission via aragonAPI, and I would see a similar popup/permission box. "App X would like access to your GitHub credentials. Yes or no?"

As far as how Aragon client stores and secures that info, I haven't thought about that yet. It would need to store the credentials themselves, as well as a list of apps to which I've granted access. It would be great if signing into Aragon on a new device unlocked the same OAuth settings, but that doesn't seem necessary for an initial version.

It's also worth noting that different apps might need different amounts of access to a GitHub profile. Maybe Projects only needs the ability to read data ("Projects would like access to **read public repository data** for your GitHub account"), while another app needs the ability to post comments ("App X would like access to **read public repository data** and **post comments** for your GitHub account"). Without further research, I'm not sure if the Aragon-client-managed model is compatible with such a minimally-permissioned approach.

ottodevs:
I appreciate your efforts here to improve the authentication process, but I am not sure about some of your assumptions, I would like to see a reproduction of the attacks you mention. I don't think the `code` or Autark app identity can be easily spoofed, I am looking forward to have better examples supporting your comments and I am not sure about taking them as true facts without any kind of proof other than generic screenshots.

The `oauth` process happens in 2 major steps, one is the code step and the other one is exchanging the code by the final github token, so as I see this, we are not really fully solving the goal, we still need to have a centralised server to prevent exposing the github client secret.

I like the solution you applied here, but is solving some security concerns to bring others.
The explanation on why we where using the previous solution was the result of a series of trade-offs, it initially was part of the projects app, so it has written in react, then it was split from projects app, and the natural iteration makes sense to be removing the react and keeping it pure js + html.

We might want to have more dialog before advancing on this

Personally, I don't think it would be a good idea to talk about implementing this on the aragon client, the current use case is so specific and narrow, in any case what we would might to have in the aragon client in a future should be a generic oauth protocol handling imho, but it seems that is not really something to worry about for now, since Projects is the only app using something like this atm.

chadoh:
@ottodevs check out this demonstration: https://github.com/chadoh/open-enterprise-oauth-hack

stellarmagnet:
@ottodevs I think you were evaluating this recently, is that still pending?

chadoh:
@ottodevs good points. I like the idea of moving Authenticator to its own repository; I've done so in the final commit here. See https://github.com/AutarkLabs/authenticator

I think we're good to merge this then. Do you agree?

I'll check in with @rkzel about removing the old authenticator process from our server, since we're no longer using it.

apps/projects/app/App.js
```diff
@@ -24,21 +24,22 @@ import {
...
-import { getToken, getURLParam, githubPopup, STATUS } from './utils/github'
+import { getToken, githubPopup, STATUS } from './utils/github'
 import Unauthorized from './components/Content/Unauthorized'
 import { LoadingAnimation } from './components/Shared'
 import { EmptyWrapper } from './components/Shared'
 import { Error } from './components/Card'
 import { DecoratedReposProvider } from './context/DecoratedRepos'
 import usePathSegments from './hooks/usePathSegments'
 
+let popupRef = null
```
ottodevs: Please, could you elaborate on why is it better to declare this outside the component and if you properly checked no potential issues could happen because putting this variable outside of the normal react lifecycles management.

What was wrong about having it in a React useState hook?

apps/projects/app/App.js
```diff
@@ -61,40 +62,39 @@ const App = () => {
...
-      // TODO: Optimize the listeners lifecycle, ie: remove on unmount
-      window.removeEventListener('message', handlePopupMessage)
+  const handlePopupMessage = useCallback(message => {
+    if (!popupRef) return
+    if (message.source !== popupRef) return
+
+    popupRef = null
 
-      const code = message.data.value
+    async function processCode() {
```
ottodevs: Why define a function inside the callback, using variables from the outer scope (`message`) just call that function from the callback context? Shouldn't be enough to put the logic directly in the callback?
Is it related to the try-catch? Also the function is being declared `async` but the call is `sync` the general structure of this callback confuses me a bit.

apps/projects/app/App.js
chadoh: When state changes, the component is re-rendered. We do not need that here. We only want to keep a reference to the pop-up so that we can verify incoming messages.

We could use `useRef`, but that is also more ceremony than we need. A global variable within this file is already encapsulated to this file, it is not accessible to other files because of how modules work. That’s perfect; that’s all we need.

apps/projects/app/App.js
chadoh: A function passed to `useCallback` cannot return anything (except another function, for the use-case of removing an event listener)

An `async` function returns a promise – that's part of the behind-the-scenes magic of `async/await`

So if you want to use `async/await` within a `useCallback`, you have to define it inline like this and then call it. We are doing this in a few spots in our code now.

An alternative would be using regular Promise notation:

```js
const handlePopupMessage = useCallback(message => {
  if (!popupRef) return
  if (message.source !== popupRef) return

  popupRef = null

  getToken(message.data.value)
    .then(token => {
      setGithubLoading(false)
      api.emitTrigger(REQUESTED_GITHUB_TOKEN_SUCCESS, {
        status: STATUS.AUTHENTICATED,
        token
      })
    })
    .catch(err => {
      console.error(err)
      setGithubLoading(false)
      api.emitTrigger(REQUESTED_GITHUB_TOKEN_FAILURE, {
        status: STATUS.FAILED,
        token: null,
      })
    })
}, [api])
```

Do you like this more, @ottodevs?

ottodevs: I think it is enough to have it the way you did, as it seems consistent with other similar patterns in our codebase. I really appreciate your explanation about this change! Thank you!

PR #1308, opened 2019-10-09
Rewards: 'View' and 'Claim' options in ContextMenu

e18r:
* Overview context menu:
Screenshot from 2019-10-09 13-14-48

* My rewards context menu:
Screenshot from 2019-10-09 13-14-55


apps/rewards/app/components/Content/MyRewards.js
```diff
@@ -1,8 +1,13 @@
 import PropTypes from 'prop-types'
 import React from 'react'
 import styled from 'styled-components'
+import moment from 'moment'
```
chadoh: I see you've already started using `moment` throughout Rewards, but we do not have `moment` in our `package.json` dependencies, and in all other places where we've been doing date calculations, we've been using `date-fns`. Personally, I love `date-fns`, while I've had very frustrating experiences with `moment`, so I'm really glad we don't use `moment`! However, if you think there's a case to be made that we switch to `moment` instead, I think we should have that discussion in an isolated PR later. For now, I think this PR needs to avoid using `moment`, and I think we need a separate PR that removes `moment` from the places where you've added it already.

apps/rewards/app/components/Content/MyRewards.js
```diff
@@ -133,14 +130,15 @@ const renderOneTimeMerit = (reward) => {
...
+    endDate,
   } = reward
   const displayAmount = (
     <Text color={String(theme.positive)}>
       +{amount} {amountToken}
     </Text>
   )
-  const disbursementDate = addDays(new Date(dateEnd), 1).toDateString()
+  const disbursementDate = (new Date(endDate)).toDateString()
+  console.log(disbursementDate)
```
chadoh: Can we remove this?

apps/rewards/app/store/events.js
```diff
@@ -33,5 +33,7 @@ export const handleEvent = async (state, event, settings) => {
   }
 
   nextState = { ...state, ...nextState }
+  nextState.convertRates = await updateConvertedRates(state)
+  console.log(nextState.convertRates)
```
chadoh: And this?

apps/rewards/app/store/reward.js
```diff
@@ -45,6 +51,22 @@ export async function onRefreshRewards(nextState, { userAddress }) {
...
+    .filter(({ verified }) => verified)
+    .map(({ symbol }) => symbol)
+
+  if (!verifiedSymbols.length) {
+    return
+  }
+
+  const res = await fetch(convertApiUrl(verifiedSymbols))
+  const convertRates = await res.json()
+  console.log(convertRates)
```
chadoh: And the two in this file?

apps/rewards/app/utils/metric-utils.js
```diff
@@ -85,6 +85,7 @@ const calculateUnclaimedRewards = (rewards, balances, convertRates) => {
 }
 
 const calculateAllRewards = (rewards, balances, convertRates) => {
+  console.log('Calcualte All Rewards Info:', rewards, balances, convertRates)
```
chadoh: And this?

Quazia:
Yeah we can remove the logging before it gets merged, I was just working out of here and hadn't finished yet.

PR #1309, opened 2019-10-09
post-review fixes

rkzel:
This fixes most of UI/UX issues. What is not yet addressed: FilterBar searchbox being focused when it should not be and borders of dropdowns.

### DIFF
--- a/apps/projects/app/components/Panel/NewIssueCuration/NewIssueCuration.js
+++ b/apps/projects/app/components/Panel/NewIssueCuration/NewIssueCuration.js
@@ -7,6 +7,7 @@ import {
   FormField,
 } from '../../Form'
 import { useAragonApi } from '../../../api-react'
+import { GU } from '@aragon/ui'
 import { usePanelManagement } from '..'
 import { issueShape } from '../../../utils/shapes.js'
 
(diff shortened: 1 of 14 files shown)


apps/projects/app/components/Panel/NewIssueCuration/NewIssueCuration.js
```diff
@@ -68,34 +69,36 @@ const NewIssueCuration = ({ allIssues, selectedIssues }) => {
...
-            onChange={updateIssues}
-            values={selectedIssues}
-            input={issues}
-            allOptions={allIssues}
-          />
-        }
-      />
-    </Form>
+    <div css={`margin: ${2 * GU}px 0`}>
+      <Form onSubmit={submitCuration} submitText="Submit Curation">
```
chadoh: Let's pass this `css` prop right to `Form`.

apps/projects/app/components/Panel/PanelComponents.js
```diff
@@ -0,0 +1,47 @@
...
+  </React.Fragment>
+)
+IssueTitle.propTypes = issueShape
+
+const IssueLinkRow = styled.div`
+  height: 31px;
+  display: flex;
+  align-items: center;
+  margin-bottom: 10px;
+  cursor: pointer;
```
chadoh: Since this `div` is already wrapped in an `<a>`, I don't think we need to specify `cursor: pointer`

PR #1312, opened 2019-10-09
Allocations: remove Budget Card hover styling

chadoh:
Note from Javi: "Budget cards shouldn’t be clickable at this stage, since doing so doesn’t lead to anything"


chadoh:
Closing in favor of #1313

PR #1316, opened 2019-10-09
Project ui

rkzel:
More UI fixes.


apps/projects/app/components/Panel/FundIssues/FundIssues.js
```diff
@@ -464,6 +447,7 @@ const FundIssues = ({ issues, mode }) => {
     // during development, sometimes this fails with a cryptic "cannot perform action" error
     // in case this happens in QA, let's leave this logging here to at least have some paper trail
     console.log( // eslint-disable-line
+      'ipfs file', issuesArray,
```
chadoh: The other things logged her are the arguments to `addBounties`. It doesn't look like this ipfs file (or is it an issues array?) is passed to it. Do you think it might help us to leave this in?

rkzel: `const ipfsAddresses = await computeIpfsString(issuesArray)`
issuesArray is actually what gets saved to IPFS. I also added bounties array (it was helpful for debugging deadline problems) - there is redundancy there, in what gets logged, but for debugging it is useful (was for me) to have more info logged rather than less. For production we should disable it all.

PR #1318, opened 2019-10-10
Allocations: fix editing budget account

chadoh:
* Stop passing `tokens`; calculate it in NewBudget
* Fix finding `editingBudget`
* Rename `onCreateBudget` to `saveBudget`, make it deal with updating as well as creating
* Listen for `SetBudget` event in Allocations background reducer


PeterMPhillips:
I did a commit with small changes to allocations arapp.json and PlanningKit.sol, should be working now :)

chadoh:
Thanks, @PeterMPhillips! Looks like this was incorporated into the new template (#1319), which has just recently been merged to `dev`! I think we can probably get rid of your commit now? Sorry about that!

PR #1322, opened 2019-10-11
Allocations/Autocomplete: fix visual & code styles

chadoh:
* Remove thick green border on selected item
* Remove extra padding below selected item (believe it or not, that's what the `display: block` is for)
* Fix code style issues identified by linter

### DIFF
--- a/shared/ui/components/form/inputs/LocalIdentitiesAutoComplete.js
+++ b/shared/ui/components/form/inputs/LocalIdentitiesAutoComplete.js
@@ -1,4 +1,4 @@
-import React, { useState, useEffect, useCallback } from 'react'
+import React, { useCallback, useEffect, useState } from 'react'
 import { useAragonApi } from '@aragon/api-react'
 import PropTypes from 'prop-types'
 import styled from 'styled-components'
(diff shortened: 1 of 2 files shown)


e18r:
@chadoh is there any discussion going on in any channel about form input validation for the recipients autocomplete? I think we need to have that conversation before removing the green border around the autocomplete, which is currently the only feedback the user is receiving. I had a private discussion about it with @jayalaves, but I guess it should have been made public.

So where should we have that discussion?

shared/ui/components/form/inputs/LocalIdentitiesAutoComplete.js
```diff
@@ -14,16 +14,16 @@ import {
 const withKey = item => ({ key: item.address, ...item })
 const sortAlphAsc = (a, b) => a.name.localeCompare(b.name)
 
-const LocalIdentitiesAutoComplete = React.memo(
```
ottodevs: Why remove memo here?

chadoh: The linter complained that the `displayName` property for this component was not there. I don't think `memo` actually gets us anything here; see [When to useMemo and useCallback](https://kentcdodds.com/blog/usememo-and-usecallback).

e18r:
Yeah, let's merge this :+1:

PR #1324, opened 2019-10-11
Allocations/New: show "Available in Vault"

chadoh:
* Simplify markup
* Place text in order that makes most sense for screenreaders; use `flex-direction: column-reverse` for sensible visual design

Available Budget: 111 ETH; Available in Vault: 200 ETH


topocount:
This looks great Chad! Thanks for implementing it.

PR #1325, opened 2019-10-11
Allocations: remove red styling from Amounts

chadoh:
Since every amount in this column will always be a debit, we've decided it does not help the user understand anything to shout at them with this bold color.

Amount: 11 ETH


chadoh:
This is not a cross-browser compatibility issue.

PR #1326, opened 2019-10-11
Pulling app image and name from execution target for app badge

PeterMPhillips:
Right now the Dot Voting app has images for each supported app in the assets folder. This change gets the images from api.installedApps() and saves them in executionTargetData for each vote. This way we don't have to have a hardcoded list of accepted apps and we don't need to make new icons when an app's icon gets updated.

Currently, we are creating the app badge using GenerateBadge in vote-styled-components.js, however if we update aragon-ui we could use the AppBadge component like Aragon's Voting app.


apps/dot-voting/app/store/votes.js
```diff
@@ -226,8 +229,76 @@ const marshallVote = ({
...
+      app.currentApp()
+        .pipe(first())
+        .subscribe((currentApp) => {
+          resolve({
+            ...vote,
+            executionTargetData: {
+              ...currentApp,
+              identifier: undefined,
+            }
+          })
```
topocount: I know we have a lot of the Promise-resolve pattern in our event store code, but I think it's important to utilize the rxjs libraries more effectively, like we've started to do in [Allocations](https://github.com/AutarkLabs/open-enterprise/blob/dev/apps/allocations/app/store/allocation.js). I'm more than happy to make this change if you'd prefer.

apps/dot-voting/app/utils/vote-styled-components.js
```diff
@@ -1,32 +1,27 @@
...
-    <img src={badgeDetails[type].icon} alt="" />
-    <span css="margin: 0 10px; padding-top: 3px; height: 24px">{badgeDetails[type].text}</span>
+      display: inline-flex;
+      font-weight: 300;
+      border-radius: 4px;
+      font-size: 16px;
+      line-height: 1.5;
+      white-space: nowrap;
+      color: #30404F;
+      background-color: rgb(12,196,230,0.09);
```
ottodevs: We might want to use named colors from aragon-ui here

apps/dot-voting/app/store/votes.js
```diff
@@ -226,8 +229,76 @@ const marshallVote = ({
...
+              }
+            })
+          } else {
+            const { appAddress, icon, identifier, name } = targetApp
+            resolve({
+              ...vote,
+              executionTargetData: {
+                address: appAddress,
+                name,
+                iconSrc: icon(24),
```
ottodevs: `icon` naming seems confusing, it seemed to be an object but then it is a function, and the param passed to it seems so cryptic, I think at least a comment here would help any future maintainer

apps/dot-voting/app/utils/vote-styled-components.js
PeterMPhillips: Agreed. However, I noticed that the Voting app uses aragon-ui verions 1.0.0-alpha.23 which contains an AppBadge component. If we switch to that, we can just get rid of vote-styled-component. If that's something we don't want to do, I can update the GenerateBadge component with proper theme variables.

apps/dot-voting/app/store/votes.js
PeterMPhillips: Yeah, I'm not sure what the change is you're looking for. It would be great if you could make the update and I'll try to replicate your solution if the issue comes up again in the future.

apps/dot-voting/app/utils/vote-styled-components.js
ottodevs: I general I am more towards following our internal development conventions and then make the needed changes once we update, or in case you feel we need to update aragon-ui do it yourself, but generally avoid merging things that accumulate tech debt to our repo, that is just my preference, do whatever you think is fair @PeterMPhillips

apps/dot-voting/app/store/votes.js
ottodevs: So any idea about why icon is a function? is this the best name we have for that? As a developer, do you honestly think that it will be clear for other developers to understand what is going on with `iconSrc: icon(24)`

ottodevs: Ok thanks for the clarifying comment, the icon() function still seems extremely confusing, but at least one can understand what it does after this long explanation.

ottodevs: I think something like `getIconWithSize(24)` - for example - would be self-explaining in this case, I let you final judgement on that @PeterMPhillips

PeterMPhillips: The function is passed by the Aragon API's installedApps() function. This is an issue we'd have to take up in that repo

apps/dot-voting/app/store/votes.js
topocount: As in the link to the allocations store code in my comment, we want to utilize the rxjs libraries more effectively, since it provides some pretty awesome tooling for reshaping the data as needed, and converting to promises using the `toPromise` method instead of resolving inside a `subscribe` call. 

This requires gaining some familiarity with the [rxjs operators](https://rxjs.dev/api/operators) more generally, but it looks like all you'll need to reshape the data here is to add a [map](https://rxjs.dev/api/operators/map) operator inside the pipe method to reshape the data as you have done inside the subcribe call, and then swap out the `subscribe` for a `toPromise()`

PeterMPhillips: Is this more to your liking?

```const decorateVote = async (vote) => {
  const executionTargets = getVoteExecutionTargets(vote)
  if (!executionTargets.length) {
    // If there's no execution target, consider it targetting this Dot Voting app
    const currentApp = await app.currentApp().pipe(first()).toPromise()
    return {
      ...vote,
      executionTargetData: {
        ...currentApp,
        identifier: undefined,
      }
    }
  }```

File: apps/dot-voting/app/store/votes.js [hunk 4]
```diff
@@ -226,8 +229,69 @@ const marshallVote = ({
...
+    const [targetAddress] = executionTargets
+    return app.installedApps().pipe(
+      first(),
+      mergeMap(installedApps => from(installedApps)),
+      first(
+        app => app.appAddress === targetAddress,
+        {
+          appAddress: targetAddress,
+          name: 'External',
+          icon: currentApp.icon,
```
topocount: We might want to come up with a more informative icon for potential external forwarded actions than just the Dot Voting icon at some point, such as a question mark or something else, but there's not clear use case for this yet

apps/dot-voting/app/store/votes.js
```diff
@@ -226,8 +229,69 @@ const marshallVote = ({
...
+    // If there's multiple targets, make a "multiple" version
+    return {
+      ...vote,
+      executionTargetData: {
+        address: '0x0000000000000000000000000000000000000000',
+        name: 'Multiple',
+        iconSrc: currentApp.icon(24),
+        identifier: undefined,
+      }
+    }
```
topocount: We might want to consider just defaulting to the last app in the target chain here, but this will require more discussion around forwarding design, and if that satisfies all use cases. Otherwise, we should consider a popout that shows the remaining execution chain or something more useful than just saying `Multiple`.

PR #1327, opened 2019-10-12
Added expanding view to allocations history

PeterMPhillips:


apps/allocations/app/components/App/AllocationsHistory.js
```diff
@@ -63,30 +62,36 @@ const AllocationsHistory = ({ allocations }) => {
...
-      //        </RecipientAmount>
-      //      </Recipient>
-      //    )
-      //  })
-      //}}
+      renderEntryExpansion={({ recipients, amount, token }) => {
+        let totalSupports = 0
+        recipients.forEach((recipient) => {
+          totalSupports += Number(recipient.supports)
+        })
```
chadoh: const totalSuppoorts = recipients.reduce

apps/allocations/app/components/App/AllocationsHistory.js
```diff
@@ -63,30 +62,36 @@ const AllocationsHistory = ({ allocations }) => {
...
+        return recipients.map((recipient, index) => {
+          const allocated = BigNumber(recipient.supports).div(totalSupports)
+          return (
+            <Recipient key={index}>
+              <IdentityBadge
+                entity={recipient.candidateAddress}
+                shorten={true}
```
chadoh: Don’t need the ={true}

apps/allocations/app/components/App/App.js
```diff
@@ -134,7 +134,7 @@ const App = () => {
   const PanelContent = panel ? panel.content : null
 
   return (
-    <Main scrollView={false}>
+    <Main scrollView={true}>
```
ottodevs:
```suggestion
    <Main>
```

As the [this prop](https://github.com/aragon/aragon-ui/tree/newstyle/src/components/Main#scrollview) is already set to `true` by default, it seems confusing to add it, I think it would make sense to just get rid of the prop, we usually don't add props when we use the defaults. Do you think of a good reason to keep it @PeterMPhillips ?

apps/allocations/app/components/App/AllocationsHistory.js
```diff
@@ -117,21 +120,31 @@ const Amount = styled.div`
...
-//   font-size: 12px;
-// `
+const Recipient = styled.div`
+   display: flex;
+   flex-direction: row;
+   width: 100%;
+   padding: 10px 10px 4px;
+   border-radius: 4px;
+   background-color: white;
+   box-shadow: 0 2px 4px #dde4e9;
```
ottodevs: Is there an aragon color that uses that `#dde4e9` code? We follow a convention of using named colors from aragon-ui themes always when they exist, so it allows updating any palette change seamlessly.

apps/allocations/app/components/App/AllocationsHistory.js
```diff
@@ -117,21 +120,31 @@ const Amount = styled.div`
...
-//   align-self: flex-end;
-//   font-size: 12px;
-// `
+const Recipient = styled.div`
+   display: flex;
+   flex-direction: row;
+   width: 100%;
+   padding: 10px 10px 4px;
+   border-radius: 4px;
+   background-color: white;
```
ottodevs: We prefer to use `theme.colorDefinition` from `aragon-ui` theme palette, instead of hardcoding the color.
The main reason is to allow, for example, switching to dark or high contrast mode, also better updates on themes.

apps/allocations/app/components/App/AllocationsHistory.js
```diff
@@ -63,30 +62,34 @@ const AllocationsHistory = ({ allocations }) => {
...
-      //      </Recipient>
-      //    )
-      //  })
-      //}}
+      renderEntryExpansion={({ recipients, amount, token }) => {
+        const totalSupports = recipients.reduce((total, recipient) => {
+          return total + Number(recipient.supports)
+        }, 0)
+        return recipients.map((recipient, index) => {
+          const allocated = BigNumber(recipient.supports).div(totalSupports)
```
ottodevs: In this case, let's keep this here, but for the future, we might want to avoid adding any kind of logic inside the return of the render function, and calculate any value outside of that instead, to prevent unnecessary extra re-renders and also to make our render functions more performant.

apps/allocations/app/components/App/AllocationsHistory.js
```diff
@@ -63,30 +62,34 @@ const AllocationsHistory = ({ allocations }) => {
...
+              />
+              <RecipientAllocation>
+                <RecipientProgress>
+                  <ProgressBar
+                    value={allocated.toNumber()}
+                    color={String(theme.accentEnd)}
+                  />
+                </RecipientProgress>
+                <RecipientAmount theme={theme}>
+                  { displayCurrency(BigNumber(amount).times(allocated)) } {' '}
```
ottodevs: We try to avoid calculations in out render returns, they accumulate and become expensive eventually degrading the performance of the components, so it is better as a rule of thumb to define every logic in constants and use here calculated values instead, just take note for the future please.
You don't need to fix it right now, but this is undoubtedly tech debt we are accumulating here.

apps/allocations/app/components/App/AllocationsHistory.js
```diff
@@ -63,30 +62,34 @@ const AllocationsHistory = ({ allocations }) => {
...
+                  {getTokenSymbol(token)} {' • '}
+                  { allocated.times(100).dp(0).toNumber() }{'%'}
```
ottodevs: Same here about logic in the render return

apps/allocations/app/components/App/AllocationsHistory.js
chadoh: @ottodevs I agree with that general sentiment, but I'm not sure it's useful or even possible here. Where do you picture moving this calculation?

apps/allocations/app/components/App/AllocationsHistory.js
chadoh: Same question here. These are not constants; they are values that we need to calculate for each recipient.

apps/allocations/app/components/App/AllocationsHistory.js
chadoh: Same question here 😄 

"Avoid all logic in render function" seems like an idea from Angular or Bootstrap or something. Avoid _side effects_ in the render function? Yes, absolutely. But logic? This seems like the most natural place for it to live to me, and I'm not even sure if or how we could move it elsewhere.

apps/allocations/app/components/App/App.js
PeterMPhillips: It was set to false which was preventing the page for scrolling. If removing it has the same effect than I'll do that.

apps/allocations/app/components/App/AllocationsHistory.js
PeterMPhillips: In the future I will do this. However, in this case I'll just remove the css entirely since it doesn't fit with @jayalaves design.

apps/allocations/app/components/App/AllocationsHistory.js
PeterMPhillips: Understood. As I said above, I'll be removing this css to fit with the original design

apps/allocations/app/components/App/App.js
ottodevs: It seems that aragon-ui docs are wrong then! thanks for testing it!

apps/allocations/app/components/App/AllocationsHistory.js
ottodevs: The idea is to move outside the return, so any change to it does not force a re-render, and is possible to event memoize the values in future optimizations, but as said, is just a general suggestion, no need to change it here for now, thanks!

apps/allocations/app/components/App/AllocationsHistory.js
ottodevs: My idea here was to do something like this:

```js
const MyComponent = ({amount, allocated}) => {
    const myCalculatedValue = displayCurrency(BigNumber(amount).times(allocated).mul(whatever).exp(anotherThing))
    const anotherExpensiveLogic = (...) // some expensive calculations here
    return(
        <Whatever>{myCalculatedValue}</Whatever>
    )
}
```

For me, it keeps returns cleaner and also allows for easier refactors / extracting logic, the general idea is to make the components dumber and eventually move logic to hooks or even background worker, that would make our frontends so light, also it is a good practices since there are some places, likes projects app where adding logic to returns resulted in unmanageable performance problems in the past, specially when you need to render the same component multiple times.

PR #1328, opened 2019-10-13
allow budget and allocation inputs to handle large values (gt 10e21)

topocount:
These changes allow users to create allocations for token amounts where the raw value can be bigger than 10e21, before token decimal division

### DIFF
--- a/apps/allocations/app/components/Panel/NewAllocation.js
+++ b/apps/allocations/app/components/Panel/NewAllocation.js
@@ -133,12 +133,13 @@ class NewAllocation extends React.Component {
       budgetValue,
       descriptionValue,
       amountValue,
-      recipients
+      recipients,
+      tokenValue
     } = this.state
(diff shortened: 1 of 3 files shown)


apps/allocations/app/components/Panel/NewAllocation.js
```diff
@@ -133,12 +133,13 @@ class NewAllocation extends React.Component {
...
       amountValue,
-      recipients
+      recipients,
+      tokenValue
     } = this.state
     const allocation = {
       budgetId: this.props.budgetId,
       budgetName: budgetValue,
-      balance: amountValue * 10e17,
+      balance: BigNumber(amountValue).times(BigNumber(10).pow(tokenValue.decimals)).toString(10),
```
Quazia: This looks good, long term I think this could be a helper function something like `displayBNToString(value, decimals)`

PR #1332, opened 2019-10-14
Create and use dev template with flexible permissions

ottodevs:
It need more polish to be perfect, like inherit from OE Template and just change permissions, but for now it should be enough to be able to quickly test apps for QA


ottodevs:
I will merge this for now, to be able to review other PRs faster, the only changes are related to a new folder, so we can easily track this in the future.

PR #1333, opened 2019-10-14
Web3/rewards

Quazia:
This PR contains the changes from @e18r to add the context menus back in as well as a lot of changes to the app-state-reducer so that the expected data is actually composed and available to the front-end. 

With these changes whether a reward is one-time or recurring is inferred by grouping all sequential rewards with the same description. This will fail in cases where somebody creates a Reward with the same description directly after another Reward with the same description.


apps/rewards/app/app-state-reducer.js
```diff
@@ -2,100 +2,109 @@ import {
...
-        reward.dateReference = new Date()
-      } else if (reward.occurances.toString() === '1'){
-        reward.rewardType = ONE_TIME_DIVIDEND
-        reward.dateReference = new Date(reward.endDate)
+    state.rewards = state.rewards === [] ? [] : state.rewards.reduce((rewards, reward) => {
+      const currentReward = rewards.find(filteredElement => {
+        return filteredElement.description === reward.description && parseInt(filteredElement.rewardId, 10) + filteredElement.occurances === parseInt(reward.rewardId)
+      })
+      if(currentReward !== undefined){
+        currentReward.occurances += 1
```
topocount: "_occurrences_"

PR #1335, opened 2019-10-15
Allocations: improved labels colors in New Allocation side panel

e18r:
Started using AragonUI's `Field` element instead of our own `FormField`.

### DIFF
--- a/apps/allocations/app/components/Panel/NewBudget.js
+++ b/apps/allocations/app/components/Panel/NewBudget.js
@@ -1,10 +1,17 @@
 import PropTypes from 'prop-types'
 import React from 'react'
 import { useAragonApi } from '../../api-react'
-import { DropDown, IconClose, Info, TextInput, theme } from '@aragon/ui'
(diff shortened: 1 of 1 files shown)


apps/allocations/app/components/Panel/NewBudget.js
```diff
@@ -91,7 +99,8 @@ class NewBudget extends React.Component {
       <Form
         onSubmit={this.createBudget}
         submitText={buttonText}
-        disabled={nameError || amountError || amountOverFunds}
+        disabled={name === '' || amount === '' || nameError || amountError
```
Quazia: This change doesn't feel directly related to the overarching PR, the line length feels handled weird, and now there's two conditions that will disable the submit without any sort of user messaging. Still probably better to have this check in here and I'm not sure that building this out falls under the scope of this PR.

Quazia:
Need to wait on #1334 for coverage to pass.

PR #1336, opened 2019-10-15
Change projects `_repoId` to use fromHex

Quazia:
This PR resolves item #1 of the Mainnet UX Audit document:

> Improve radspec (adding project)
> We passed in the repo ID but it seems to be in bytes form vs. the actual ID used by Github.
> 
> 
> Solution (options):
> Remove this from the radspec since it’s not immediately useful
> Show the actual Github ID
> Add a metadata field in the contract that gets populated with the repo name when adding a project. In the future we can have an ‘updateProject’ contract call which allows user to “sync repo name” in the case of when a repo name was updated

This leverages solution #2, showing the actual Github ID (For example MDEwOlJlcG9zaXRvcnkxNzEyMjg2Mjk=)

Not sure that this is informative enough for the user.

### DIFF
--- a/apps/projects/contracts/Projects.sol
+++ b/apps/projects/contracts/Projects.sol
@@ -357,7 +357,7 @@ contract Projects is AragonApp, DepositableStorage {
 // Repository functions
 ///////////////////////
     /**
-     * @notice Add repository `@fromHex(_repoId, "ascii")`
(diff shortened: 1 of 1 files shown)


Quazia:
This has been modified to show nothing instead of the repoID.

PR #1341, opened 2019-10-15
Copy VoteCardGroup from Voting app

Quazia:
In response to Mainnet UX Audit 14

>  “Past votes” and “Open votes” not correct size
> Past votes should say “Closed votes” like Voting app
> The font size should be the same as voting app

Switched to using the exact code used in voting app `VoteCardGroup.js` and switched the copy.

https://github.com/aragon/aragon-apps/blob/master/apps/voting/app/src/components/VoteCard/VoteCardGroup.js


ottodevs:
I resolved the merge conflict on AllocationsHistory.js

PR #1343, opened 2019-10-15
Added displayYouBadge prop to VotingOptions. Now to display the badge…

PeterMPhillips:
… that shows a user's vote, you must set displayYouBadge to true

### DIFF
--- a/apps/dot-voting/app/components/VoteDetails/VotingResults.js
+++ b/apps/dot-voting/app/components/VoteDetails/VotingResults.js
@@ -32,6 +32,7 @@ const VotingResults = ({ vote, voteWeights }) => {
           totalSupport={totalSupport}
           color={`${theme.accent}`}
           voteWeights={voteWeights}
+          displayYouBadge={true}
         />
       </div>
       {!vote.open &&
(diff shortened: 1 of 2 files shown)


apps/dot-voting/app/components/VoteDetails/VotingResults.js
```diff
@@ -32,6 +32,7 @@ const VotingResults = ({ vote, voteWeights }) => {
           totalSupport={totalSupport}
           color={`${theme.accent}`}
           voteWeights={voteWeights}
+          displayYouBadge={true}
```
ottodevs: We don't really need to add true when using boolean props, but I am not really opposed to being explicit about that if that helps you @PeterMPhillips , so I will not really suggest the change this time. Just tell me your opinion about this, to be consistent in the future.

apps/dot-voting/app/components/VotingOptions.js
```diff
@@ -13,6 +13,7 @@ class VotingOptions extends React.Component {
     voteWeights: [],
     // animationDelay can also be a number to disable the random delay
     animationDelay: { min: ANIM_DELAY_MIN, max: ANIM_DELAY_MAX },
+    displayYouBadge: false,
```
ottodevs: I would say that we don't need to add false as default prop, since undefined will be evaluated also as false at line 61

PR #1344, opened 2019-10-15
Projects/Settings: Add Funding Model option

chadoh:
* Can be Hourly or Fixed
* When Fixed, hide Base Rate field
* Store this value by setting Base Rate to 0
* Add fundingModel attribute to bountySettings in appState

See commit messages for full details. You'll see that in my first commit, I improved the styling to make it easier to add this new field. Here's what it looks like:

## Small

Funding Model: Hourly on mobile

## Medium

Screen Shot 2019-10-15 at 17 48 38

## Large

Screen Shot 2019-10-15 at 17 49 05

### DIFF
--- a/apps/projects/app/components/Content/Settings.js
+++ b/apps/projects/app/components/Content/Settings.js
@@ -1,7 +1,6 @@
 import React, { useEffect, useState } from 'react'
 import PropTypes from 'prop-types'
 import styled from 'styled-components'
-import NumberFormat from 'react-number-format'
 import { useAragonApi, useNetwork } from '../../api-react'
 import {
   Box,
@@ -24,6 +23,7 @@ import { REQUESTED_GITHUB_DISCONNECT } from '../../store/eventTypes'
 import useGithubAuth from '../../hooks/useGithubAuth'
 import { LoadingAnimation } from '../Shared'
 import { EmptyWrapper } from '../Shared'
+import NumberInput from '../Shared/NumberInput'
 
 const bountyDeadlines = [ 'Days', 'Weeks', 'Months' ]
 const bountyDeadlinesMul = [ 24, 168, 720 ]
@@ -52,13 +52,14 @@ const GitHubConnect = ({ onLogin, onLogout, status }) => {
   return (
     <Box heading="GitHub" css="height: 100%">
       <Text.Block>{bodyText}</Text.Block>
-      <StyledButton
+      <Button
+        css={`margin-top: ${1 * GU}px`}
         wide
         mode="secondary"
         onClick={buttonAction}
       >
         {buttonText}
-      </StyledButton>
+      </Button>
     </Box>
   )
 }
@@ -111,35 +112,44 @@ const ExperienceLevel = ({
   generateExpLevelHandler,
 }) => {
   let last = expLevels[expLevels.length - 1]
-  let disableAdd = last.mul !== '' && last.name !== '' ? false : true
+  let canAdd = last && last.mul !== '' && last.name !== ''
   return (
-    <div>
+    <React.Fragment>
       <SettingLabel text="Difficulty multipliers" />
 
       {expLevels.map((exp, index) => (
-        <Field key={index} label={'LEVEL ' + index}>
-          <StyledNumberFormat
-            fixedDecimalScale
-            decimalScale={2}
-            value={exp.mul}
-            allowNegative={false}
-            onChange={generateExpLevelHandler(index, 'M')}
-          />
-          <StyledTextInput
-            defaultValue={exp.name}
-            onChange={generateExpLevelHandler(index, 'N')}
-          />
+        <Field
+          css={`margin-bottom: ${1.5 * GU}px`}
+          key={index}
+          label={'LEVEL ' + index}
+        >
+          <div css="display: flex">
+            <NumberInput
+              css="width: 32%; margin-right: 3%"
+              fixedDecimalScale
+              decimalScale={2}
+              value={exp.mul}
+              allowNegative={false}
+              onChange={generateExpLevelHandler(index, 'M')}
+            />
+            <TextInput
+              type="text"
+              css="font-size: 16px; width: 65%"
+              defaultValue={exp.name}
+              onChange={generateExpLevelHandler(index, 'N')}
+            />
+          </div>
         </Field>
       ))}
-      <StyledButton
-        disabled={disableAdd}
+      <Button
+        disabled={!canAdd}
         compact
         mode="secondary"
         onClick={onAddExpLevel}
       >
         + Add Another
-      </StyledButton>
-    </div>
+      </Button>
+    </React.Fragment>
   )
 }
 ExperienceLevel.propTypes = {
@@ -148,16 +158,6 @@ ExperienceLevel.propTypes = {
   generateExpLevelHandler: PropTypes.func.isRequired,
 }
 
-const EmptyBaseRate = () => (
-  <div css="margin-bottom: 12px">
-    <SettingLabel text="Base rate" />
-    <Text.Block>
-      Once you have tokens in your Vault you will be able to set your
-      bounty base rate, which provides you with the ability to allocate bounties to issues.
-    </Text.Block>
-  </div>
-)
-
 const BaseRate = ({
   baseRate,
   onChangeRate,
@@ -165,10 +165,10 @@ const BaseRate = ({
   onChangeCurrency,
   bountyCurrencies,
 }) => (
-  <div css="margin-bottom: 24px">
+  <React.Fragment>
     <SettingLabel text="Base rate" />
-    <StyledInputDropDown>
-      <StyledNumberFormat
+    <StyledInputDropDown css="margin-bottom: 24px">
+      <NumberInput
         fixedDecimalScale
         decimalScale={2}
         value={baseRate}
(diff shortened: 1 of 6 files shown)


chadoh:
@jayalaves See the screenshots. I moved the Info box to below the Funding Model field, as we discussed. Do you like it there, or would you prefer it go below the Base Rate field? And do you want any modifications to the spacing around it?

apps/projects/app/components/Content/Settings.js
```diff
@@ -298,35 +297,39 @@ const Settings = ({ onLogin }) => {
...
+    // we persist `fundingModel: fixed` by setting baseRate to 0
+    const baseRateModified = fundingModels[selectedFundingModelIndex] === 'Fixed'
+      ? 0
+      : baseRate
 
     api.changeBountySettings(
       expLevelsMul,
       expLevelsDesc,
-      toHex(baseRate * 100),
+      toHex(baseRateModified * 100),
```
chadoh: @Quazia is this how you would expect this to look?

apps/projects/app/store/helpers/settings.js
```diff
@@ -20,7 +20,12 @@ export const syncSettings = async state => {
     )
     const baseRateDecimal = baseRate / 100.0
 
-    state.bountySettings = { ...settings, baseRate: baseRateDecimal, expLvls }
+    state.bountySettings = {
+      ...settings,
+      baseRate: baseRateDecimal,
+      expLvls,
+      fundingModel: Number(baseRate) === 0 ? 'Fixed' : 'Hourly',
```
chadoh: @Quazia any feedback on this?

apps/projects/app/components/Content/Settings.js
```diff
@@ -421,28 +437,24 @@ Settings.propTypes = {
   onLogin: PropTypes.func.isRequired,
 }
 
-const StyledInputDropDown = ({ children }) => {
+const StyledInputDropDown = ({ children, className }) => {
```
ottodevs: I think as a convention, we should not prefix with the "Style" word components that are not Styled components. There is a simple reason. only styled components can be extended by other styled components, if we ever try in the future to extend this component, for example, with styled(StyledNumberFormat) that will cause an error and be confusing for the developer, please, avoid using Styled prefixes for non-styled-components...

Also why don't just use a style component for this one? I don't like to see so much inlined CSS in the components.

Another good reason, if the other 2 are not enough, is that inlined css is not easily lintable, making it error prone, while styled-components are currently being automatically linted in our project config (with stylelint)

ottodevs: Could you please elaborate on why we need to use the `className` prop here?

apps/projects/app/components/Content/Settings.js
```diff
@@ -421,28 +437,24 @@ Settings.propTypes = {
...
         border-radius: 3px 0 0 3px;
         border: 1px solid ${theme.border};
         box-shadow: 0 4px 4px 0 rgba(0, 0, 0, 0.03);
-        width: 140px;
+        flex: 1;
         z-index: 1;
-        :focus {
-          outline: 0;
-          border: 1px solid ${theme.infoSurface};
+        &, &:focus-within {
```
ottodevs: I would like to ask what is doing this new rule introduced here? I tried to remove it on my browser and didn't notice any change to the UI.
I also checked https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-within but is still not clear how it applies to our current use case.

jayalaves:
@chadoh 

> I moved the Info box to below the Funding Model field, as we discussed. Do you like it there, or would you prefer it go below the Base Rate field?

This looks exactly as expected. Thank you!

@ottodevs 

> I observed an UX issue where selecting Fixed rate, does not allow to select the token for funding, while selecting Hourly rating allows to also select which token will be used as default for funding. Is this the expected behavior?

Do you think we should limit the token of choice, also if the funding is set to fixed? It was the intention in the design to leave that open and allow for more granular control over the funding.

apps/projects/app/components/Content/Settings.js
```diff
@@ -470,38 +483,48 @@ const SettingsMain = styled.div`
...
-})`
-  height: 40px;
-  width: 185px;
-  font-size: 16px;
+  display: grid;
+  grid-gap: ${3 * GU}px;
+  grid-template-columns: ${({ layoutName }) => layoutName === 'small' ? '1fr' : '1fr 1fr'};
+  margin-bottom: ${3 * GU}px;
 `
+const StyledNumberFormat = ({ className, theme, ...props }) => (
```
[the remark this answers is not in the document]
chadoh: > only styled components can be extended by other styled components

That's not true. The `className` convention allows any component to be wrapped with `styled`. You'll see that I wrapped `StyledNumberFormat` throughout this change, using the `css` prop.

That said, I don't really like the name of this component, but I didn't think it would strengthen this PR to change the name of it. By keeping the name and the signature the same, I was able to keep unrelated code (the places that use the component) unchanged. I value that more than I value a convention around when to use `Styled` as a prefix.

> Also why don't just use a style component for this one? I don't like to see so much inlined CSS in the components.

This is related to keeping the columns a consistent width. On fields like Base Rate, we want the "dropdown part" to be as wide as its content—"ETH" vs "Autarkian Dollars"—and we want the numeric part to take up the rest of the space. This means we want to give the numeric part `flex-grow: 1` (or similar), and give `flex-grow: 0` to the dropdown. This `flex-grow: 1` on the numeric part fights with the `padding: 0 10px`, and the whole number+dropdown ends up extending beyond the right boundary of its column by 20px.

The only fix I could come up with, after a lot of trying other things, is to put the number input inside a div. I liked that I was able to keep the function signature the same as the previous single-element component.

As far as using named components vs inline `css` props: I used to be with you, that named components are preferrable. After reviewing lots of PRs from Radek where he followed A1's lead in using more inline `css` props, I stopped caring as much. Naming each of these sub-components seems at least as confusing as having inline `css` props. What would they be named?

> Another good reason, if the other 2 are not enough, is that inlined css is not easily lintable, making it error prone, while styled-components are currently being automatically linted in our project config (with stylelint)

Ah, it looks like this is, indeed, [still an issue](https://github.com/styled-components/stylelint-processor-styled-components/issues/270). Bummer. 

I do find this compelling.

I'll spend a couple minutes trying to come up with decently explanatory names for the sub-components within this one, so we can get linting on them.

I don't think it's worth spending more than a couple minutes on, though. Eventually, stylelint will probably work around this limitation.

apps/projects/app/components/Content/Settings.js
chadoh: I've addressed most of these comments in the first spot where you made them, but I'll explain some specifics about `StyledInputDropDown`:

* This component was already non-`styled`. I did not change that; I do not think it adds to the clarity of changes to make such unrelated alterations.
* As I already mentioned, passing `className` allows wrapping this component with `styled`, which is what the inline `css` prop does behind the scenes. This allowed me to move the `margin-bottom` to only the occurrence of `StyledInputDropDown` where we needed it.

apps/projects/app/components/Content/Settings.js
chadoh: Make sure you click into the input. We are adding border to the div, not the input itself. `:focus` applies directly to an input; `:focus-within` allows adding border to a div when an element inside it has focus.

apps/projects/app/components/Content/Settings.js
chadoh: I have moved this component to the Shared directory and renamed it to `NumberInput`. I used regular `styled` components rather than the inline `css` prop. I also added a comment explaining why the wrapping div is necessary. Let me know what you think, Otto!

chadoh:
@ottodevs thanks for the thorough review! @e18r thanks for handling the FundIssues side panel! I've rebased with latest `dev` and addressed your feedback, @ottodevs. I've tested the FundIssues behavior both when this setting is Fixed and when it is Hourly, and it works well. I'd love another review! I think it might be read.

apps/projects/app/components/Content/Settings.js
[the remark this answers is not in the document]
ottodevs: Sorry I was confused about extending styles from previous styled-components versions Chad.
My review on this was a bit excessive, I found your comments fair, even if the linting reason was compelling, after revisiting this, I think it is not enough to really spent lot of time on it.

I changed my mind after reading your explanation, so now I find the current code is correct. I agree with you, we can perfectly live with this `stylelint` limitation, just being a bit extra careful when dealing with inlined styles.

apps/projects/app/components/Content/Settings.js
ottodevs: Thank you for that, anyway, since you are coding a lot of the frontend, I guess it is fair to use whatever convention makes you comfortably working with the code, because I don't consider this enough important to ask you refactoring or renaming components anymore. Sorry if my previous review was a bit out of place.
Your reasons were pretty convincing for me.

ottodevs: I spent some hours refreshing my knowledge about styled-components , so now I caught on it. Sorry for the inconvenience Chad.

chadoh:
> So, if the user has Fixed rate, but selects Hourly, then saves setting, it will automatically go back to Fixed Rate, is a bit weird behaviour, probably more UX related than code. I don't know if that could result confusing, anyway is not a big deal and it does not seem to break anything.

This is similar to how it worked already. I don't think this PR is the right place to invest time in discussing & improving this UX, though I agree we can improve this!

PR #1350, opened 2019-10-16
UI fixes: 38b, 39, 43, 51

rkzel:
38b: Labels are not uppercased by default,
43: Projects text search, "clear text" is now standard link color,
51: Fund Issues Panel, Update Form looks more like in Figma.
FilterBar: moved 2 out of 3 Popovers away from main component, so search box doesn't lose focus (it was almost working before, needed small fixes),
Overflow filters (the ones hidden on small screens) are now working much better.


apps/projects/app/components/Card/Issue.js
```diff
@@ -21,6 +21,7 @@ const labelsTags = (labels, theme) =>
       css="margin-right: 10px; width: auto"
       background={'#' + label.node.color + '99'}
       color={`${theme.surfaceContent}`}
+      uppercase={false}
```
ottodevs: Just as a side comment, I think omitting the prop would be perfectly fine on boolean props to denote a false value on them.
```suggestion
```

apps/projects/app/components/Content/IssueDetail/DetailsCard.js
```diff
@@ -173,6 +173,7 @@ const DetailsCard = ({ issue }) => {
               style={{ marginRight: '5px', marginTop: '20px' }}
               background={'#' + label.node.color + '99'}
               color={`${theme.content}`}
+              uppercase={false}
```
ottodevs: We can probably just omit the prop and it will be `false`, not a big deal, just in case you didn't know that @rkzel 
```suggestion
```

apps/projects/app/components/Panel/FundIssues/FundIssues.js
```diff
@@ -24,8 +25,7 @@ import {
   Info,
 } from '@aragon/ui'
 
-import { Form, FormField, FieldTitle } from '../../Form'
-import { DateInput } from '../../../../../../shared/ui'
+import { Form, FormField, FieldTitle, DateInput } from '../../Form'
```
ottodevs: It is interesting to keep using shared components instead of local ones, to avoid duplication, for example, this work will be repeated once we want to also update the `DateInput` component in `Rewards` or other apps

apps/projects/app/components/Form/Input/DateInput.js
```diff
@@ -8,22 +8,22 @@ import DatePicker from './DatePicker'
...
 `
 const TextInputDate = styled(TextInput).attrs({
   readOnly: true
 })`
-  width: 123px;
-  height: 32px;
+  height: 40px;
   display: inline-block;
   padding-top: 3px;
+  width: 100%;
```
ottodevs: Not sure if these changes are also useful for rewards apps, in case they are, I think it would be a good idea to put them into a shared folder so we don't need to duplicate efforts.

apps/projects/app/components/Card/Issue.js
rkzel: Yes, but the default is `true`, so I need to explicitly give it `false` value.

apps/projects/app/components/Panel/FundIssues/FundIssues.js
rkzel: Yes. Created issue #1406 for it.

apps/projects/app/components/Shared/FilterBar/FilterDropDown.js
```diff
@@ -0,0 +1,163 @@
...
+
+  return (
+    <ClickOutHandler onClickOut={handleClickOut}>
+      <Spring
+        config={springs.smooth}
+        to={{ openProgress: Number(opened) }}
+        native
+      >
+        {({ openProgress }) =>
+          type === 'overflow' ? (
```
chadoh: I'd rather we kept `FilterDropDown` and `FilterOverflow` as separate components, rather than passing this `type` property in. We could move the `ClickOutHandler` and `Spring` stuff to some other component, if we wanted. Is that the only reason that they're combined right now?

apps/projects/app/components/Shared/FilterBar/FilterBar.js
```diff
@@ -75,6 +75,46 @@ SearchPopover.propTypes = {
...
 }
 
+const SortPopover = ({ visible, opener, setSortMenuVisible, sortBy, updateSortBy }) => {
+  const theme = useTheme()
+  const sorters = [
+    'Name ascending',
+    'Name descending',
+    'Newest',
+    'Oldest',
+  ]
```
chadoh: No need for this to be defined within the component. This causes it to be garbage-collected and re-allocated over and over.

apps/projects/app/components/Shared/FilterBar/FilterBar.js
```diff
@@ -75,6 +75,46 @@ SearchPopover.propTypes = {
...
+
+  return (
+    <Popover
+      visible={visible}
+      opener={opener}
+      onClose={() => setSortMenuVisible(false)}
+      css="padding: 12px"
```
chadoh: Let's use `GU` instead of `px`

apps/projects/app/components/Shared/FilterBar/FilterBar.js
```diff
@@ -75,6 +75,46 @@ SearchPopover.propTypes = {
...
+      placement="bottom-start"
+    >
+      <Label text="Sort by" />
+      {sorters.map(way => (
+        <FilterMenuItem
+          key={way}
+          onClick={updateSortBy(way)}
+        >
+          <div css="width: 24px">
```
chadoh: `GU` here, too

apps/projects/app/components/Shared/FilterBar/FilterBar.js
```diff
@@ -280,14 +323,14 @@ const FilterBar = ({
...
           </ActionLabel>
         </FilterMenuItem>
-      )),
+      ))}
       <hr css={`
         height: 1px;
         border: 0;
         width: 100%;
         background: ${theme.border};
       `} />,
```
chadoh: I think we need to get rid of this comma, now.

apps/projects/app/components/Shared/FilterBar/FilterBar.js
rkzel: fixed.

apps/projects/app/components/Shared/FilterBar/FilterBar.js
rkzel: fixed.

apps/projects/app/components/Shared/FilterBar/FilterBar.js
rkzel: Moved to the top.

apps/projects/app/components/Shared/FilterBar/FilterDropDown.js
```diff
@@ -0,0 +1,163 @@
...
+  type: PropTypes.string.isRequired,
+  children: PropTypes.node.isRequired,
+}
+
+const DropDownButtonOverflow = ({ disabled, handleBaseButtonClick, openProgress, handleClose, children }) => (
+  <Main>
+    <FilterButton
+      onClick={handleBaseButtonClick}
+      disabled={disabled}
+      width={'40px'}
```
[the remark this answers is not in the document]
rkzel: It used to be a variable... fixed.

apps/projects/app/assets/index.js
```diff
@@ -6,5 +6,6 @@ import IconFilter from './IconFilter'
 import IconBarbell from './IconBarbell'
 import IconOpen from './IconOpen'
 import IconClose from './IconClose'
+import IconDropArrow from './IconDropArrow'
 
-export { IconMore, IconSort, IconGrid, IconCoins, IconFilter, IconBarbell, IconOpen, IconClose }
+export { IconMore, IconSort, IconGrid, IconCoins, IconFilter, IconBarbell, IconOpen, IconClose, IconDropArrow }
```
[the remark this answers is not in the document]
rkzel: I tried and it said:
Support for the experimental syntax 'exportDefaultFrom' isn't currently enabled (1:8):

apps/projects/app/components/Shared/FilterBar/FilterDropDown.js
rkzel: They started as one component and diverged to the current point - which is the point they can be properly separated. Will to a special ticket.

rkzel:
Thank you! One problem with mobile screen vs FilterBar is that the `small` layoutName covers everything from reasonable width to a very small one. We need better granularity - and yes, calculating number of filters shown needs to return. It's not a 5-minutes change, so I'm going to merge current state, and finish the two remaining items as separate PR tonight (calculating width and separation of FilterDropDown elements).

PR #1352, opened 2019-10-16
Projects: default work deadline in weeks + improved unit conversion logic

e18r:
Changed the available units for the default work deadline to days, weeks and months. Left the default as it was, which is two weeks. Changed the unit conversion logic to the following:

1. If there's a unit in which the amount of time is an integer, choose that. If there is more than one such unit, choose the longest. So for instance 180 days will be displayed as 6 months.
2. If the amount of time has to be displayed with decimals, then choose the longest unit that's less or equal to half the amount of time to be displayed. So for instance 10.5 days is displayed in days because a week isn't less or equal to 5.25 days.

### DIFF
--- a/apps/projects/app/components/Content/Settings.js
+++ b/apps/projects/app/components/Content/Settings.js
@@ -338,7 +338,7 @@ const Settings = ({ onLogin }) => {
   }
 
   const baseRateChange = e => setBaseRate(e.target.value)
-  const bountyDeadlineChangeT = e => setBountyDeadlineT(+e.target.value)
+  const bountyDeadlineChangeT = e => setBountyDeadlineT(Number(e.target.value))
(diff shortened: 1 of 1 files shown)


apps/projects/app/components/Content/Settings.js
```diff
@@ -322,7 +338,7 @@ const Settings = ({ onLogin }) => {
   }
 
   const baseRateChange = e => setBaseRate(e.target.value)
-  const bountyDeadlineChangeT = e => setBountyDeadlineT(e.target.value)
+  const bountyDeadlineChangeT = e => setBountyDeadlineT(+e.target.value)
```
ottodevs:
```suggestion
  const bountyDeadlineChangeT = e => setBountyDeadlineT(Number(e.target.value))
```

PR #1358, opened 2019-10-17
Add --files to publish:http scripts

ottodevs:
This small PR adds the option to `publish:http`, that prevents publishing the whole app folder, that will solve 2 main issues:
- The large `apm` package size, that also takes more space in the hard disk
- The extra-time needed to add those files that are not needed to be in the package

Even `aragon-cli` suggests to do so:
image

In the future I think we should move our publish commands to an external script to have better readability/maintainability and deduplicate a bit, but for now this will help a bit, even in publish:http!

---

The second part of this PR just removes `--all` from every `truffle compile` call, since truffle automatically detects which contract changed to compile again, we are basically overriding this smart behavior to force compiling every contract each time! that is really a ton of unnecessarily wasted time, when `--all` should just be used for debugging purposes.

source: https://www.trufflesuite.com/docs/truffle/getting-started/compiling-contracts#command

### DIFF
--- a/apps/address-book/package.json
+++ b/apps/address-book/package.json
@@ -17,7 +17,7 @@
     "prepublishOnly": "truffle compile --all",
     "publish": "aragon apm publish patch && sleep 3",
     "publish:cd": "yes | aragon apm publish major --files dist/ --environment continuous-deployment --apm.ipfs.rpc https://ipfs.autark.xyz:5001 --ipfs-check false --skip-confirmation true",
-    "publish:http": "npm run build:script && yes | aragon apm publish major --http localhost:1111 --http-served-from ./dist --propagate-content false --skip-confirmation true",
+    "publish:http": "npm run build:script && yes | aragon apm publish major --files dist --http localhost:1111 --http-served-from ./dist --propagate-content false --skip-confirmation true",
     "publish:major": "aragon apm publish major --files dist/",
     "publish:minor": "aragon apm publish minor --files dist/",
     "publish:patch": "aragon apm publish patch --files dist/",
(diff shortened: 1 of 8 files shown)


topocount:
> Looks great! Thank you!
> 
> One thing I would love to have you improve before merging, which I think will be really helpful in the future:
> 
> Your description of these changes _in the Pull Request description_ is amazing. In the future, if I am looking at the code, and I want to understand the `prepublishOnly` command, I want to be able to quickly access your description of why `--all` isn't necessary. Having your description _in the commit comment_ would really help! It would let me use tools like `git blame` or `tig` to find that info, without needing to leave my code editor, load the commit in GitHub, and then click through to the PR. Also, if we ever decide to use a different host, rather than GitHub, having your descriptions right in the commit history will bring them along with us.
> 
> More inspiration for writing great commit messages: https://chris.beams.io/posts/git-commit/

I'm in favor of setting on a standard commit message format with the team that can be enforced in all future PRs. Something to discuss in our next dev meeting.

ottodevs:
> Looks great! Thank you!
> 
> One thing I would love to have you improve before merging, which I think will be really helpful in the future:
> 
> Your description of these changes _in the Pull Request description_ is amazing. In the future, if I am looking at the code, and I want to understand the `prepublishOnly` command, I want to be able to quickly access your description of why `--all` isn't necessary. Having your description _in the commit comment_ would really help! It would let me use tools like `git blame` or `tig` to find that info, without needing to leave my code editor, load the commit in GitHub, and then click through to the PR. Also, if we ever decide to use a different host, rather than GitHub, having your descriptions right in the commit history will bring them along with us.
> 
> More inspiration for writing great commit messages: https://chris.beams.io/posts/git-commit/

I am down to start doing this when as soon as the whole team agrees to stick to this practice, so I guess it is something to discuss in our next devs call, as Kevin pointed!

Thank you very much for this interesting feedback Chad!

PR #1360, opened 2019-10-17
Projects: stop making GitHub queries in script.js

chadoh:
Since we can't be confident that the user is logged in at the time events are read from the blockchain,

And since we can't be sure we'll see these events again due to how aragonAPI's store caching works,

We want everything in the appState reducer (aka script.js) to rely only on blockchain and IPFS data.

This means that the other data in `repos` needs to be loaded in on the frontend, once we know that the user has logged into GitHub.

Note that this architecture uses a `globalDecoratedRepos` variable, defined at the top level of the new `hooks/useRepos` file. This ensures that the first frontend component that needs to decorate repos (which happens to be App.js) will update this file-global variable, and subsequent components that need decorated repos can avoid unnecessary API calls to GitHub.

**Recommendation:** view the file changes [with indentation changes ignored](./1360/files?w=1)

### DIFF
--- a/apps/projects/app/App.js
+++ b/apps/projects/app/App.js
@@ -30,6 +30,7 @@ import Unauthorized from './components/Content/Unauthorized'
 import { LoadingAnimation } from './components/Shared'
 import { EmptyWrapper } from './components/Shared'
 import { Error } from './components/Card'
+import { DecoratedReposProvider } from './context/DecoratedRepos'
 
 const App = () => {
   const { api, appState } = useAragonApi()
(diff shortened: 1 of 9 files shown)


chadoh:
@schwartz10 thanks!

I thought about that race condition, and that maybe we don’t need to care. Two reasons:

1. It won’t happen often. Two of the components that use this hook are side panels. They won’t both be open at the same time. And how frequently will a repo be loaded from the blockchain while one of the side panels is open? Probably pretty infrequently. 
2. The operation is idempotent. Even if we do fire an unnecessary call to GitHub, the results of both queries will be identical.

That said, I agree that your proposal avoids any unexpected weirdness, and would be especially helpful as the app grows and we use this hook in new places. Silly as it is, I think I am biased against context because of the need for a bajillion higher order “provider” components in App.js. I will try to get over it and implement one of the solutions that you’ve proposed.

chadoh:
@Schwartz10 

> The general pattern to avoid in the background script is conditionally setting store state. Are we confident that repos was the only place that was happening? I poked around the code a little and I'm wondering if the same pattern exists for issues too:

The `issues` logic that you linked to there is only fetching on-chain data. It's basically doing an _upsert_ – checking if the issue is already in `appState.issues`, updating it if so, inserting it if not. That seems safe to me. Do you agree?

chadoh:
@Schwartz10 this has been updated to use Context. Note that the most straightforward way to do this was to base this PR on my prior work in #1355, so I'd appreciate your review there, too!

apps/projects/app/App.js
```diff
@@ -183,52 +184,53 @@ const App = () => {
...
+                    </React.Fragment>
+                  )
+                  : (
+                    <React.Fragment>
+                      <Tabs
+                        items={tabs.map(t => t.name)}
+                        onChange={handleSelect}
+                        selected={activeIndex.tabIndex}
+                      />
+                      <TabComponent
```
topocount: This component is throwing an error (on line 164) now after:
1. Signing into Github
2. Adding a Repo
3. Signing out of Github
4. Signing back into Github

chadoh: Ah, thanks, I'll look into it!

chadoh: Updated. I got to delete more code!

PR #1361, opened 2019-10-17
simplify data syncing

Schwartz10:


apps/projects/app/app-state-reducer.js
```diff
@@ -1,27 +1,5 @@
...
-      so it knows when its receiving the store cached state, and stick with it
-    */
-    return { ...state }
-  }
-
-  /*
-    otherwise, ignore application state
-  */
-  return INITIAL_STATE
+  return { ...state }
```
chadoh: Why duplicate state? Does this do something useful, or just cause app re-renders for no reason?

```suggestion
  return state
```

PR #1394, opened 2019-10-21
Dot Voting: fix padding for wider layouts

### LINKED ISSUE #1373: Dot Voting: improve spacing between Open & Closed votes

### LINKED ISSUE #1374: Tiles are not properly aligning in medium-width in Dot Voting
Tiles are overlapping with each other
DAO URL: https://beta.autark.xyz/#/oe-demo/0x98996b45bc4c1c848d53de4a1b59c63e4c2fdfaf

This occurs from 1000-1300 width
image

chadoh:
Fixes #1373 
Fixes #1374 

We want to use the same padding values on wider screens that we do on narrow screens.

Also, there's no reason to make the headings unselectable. I see no benefit of doing so, only potential frustration.

### DIFF
--- a/apps/dot-voting/app/components/VoteCardGroup.js
+++ b/apps/dot-voting/app/components/VoteCardGroup.js
@@ -6,16 +6,11 @@ import {
   GU,
   Tag,
   textStyle,
-  unselectable,
-  useLayout,
   useTheme,
 } from '@aragon/ui'
 
 const VoteCardGroup = ({ title, count, children }) => {
   const theme = useTheme()
-  const { layoutName } = useLayout()
-  const compactMode = layoutName === 'small'
-  const rowHeight = compactMode ? null : 294
 
   return (
     <section>
@@ -24,8 +19,7 @@ const VoteCardGroup = ({ title, count, children }) => {
           display: flex;
           align-items: center;
           margin-bottom: ${3 * GU}px;
-          ${compactMode ? `padding: 0 ${2 * GU}px;` : ''}
-          ${unselectable};
+          padding: 0 ${2 * GU}px;
         `}
       >
         <div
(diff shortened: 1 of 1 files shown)


apps/dot-voting/app/components/VoteCardGroup.js
```diff
@@ -6,16 +6,12 @@ import {
...
   Tag,
   textStyle,
-  unselectable,
   useLayout,
   useTheme,
 } from '@aragon/ui'
 
 const VoteCardGroup = ({ title, count, children }) => {
   const theme = useTheme()
-  const { layoutName } = useLayout()
```
ottodevs: We could also remove the `useLayout` import since it is not used anymore here

apps/dot-voting/app/components/VoteCardGroup.js
```diff
@@ -47,7 +42,7 @@ const VoteCardGroup = ({ title, count, children }) => {
           <Tag limitDigits={4} label={count} size="small" />
         </span>
       </h2>
-      <CardLayout columnWidthMin={30 * GU} rowHeight={rowHeight}>
+      <CardLayout columnWidthMin={30 * GU} rowHeight={null}>
```
ottodevs: I would better omit the prop, or pass an ugly `undefined` instead, since it can cause some issue when used in the tagged template literal [here](https://github.com/aragon/aragon-ui/blob/4dac5c74736e5542a626eb19949d9214b504d3dd/src/components/CardLayout/CardLayout.js#L22), because of the resulting `nullpx`

Omitting it will probably be a better idea as the component will use the defaultProp value instead `21 * GU`
```suggestion
      <CardLayout columnWidthMin={30 * GU}>
```

chadoh: We do _not_ want the height to be `21 * GU`! We want the height to be `"auto"` or `"initial"`. Maybe we can use `css="grid-auto-rows: auto"` instead, though.

I've [started a conversation](https://github.com/aragon/aragon-ui/commit/b2d1da7f7cce7ed3dd7cf843296c41ac13f67100#r35598639) about the `21 * GU` default; please chime in there if you have ideas around how to improve this component!

ottodevs: I understand, anyway passing `null` as the prop value does not seem a good solution for that problem, it could also break propTypes in the component probably.

PR #1399, opened 2019-10-22
Projects: truncate issue title in fund issue panel

### LINKED ISSUE #1371: Truncation issues for Fund Issue panel
- [ ] In Firefox, the issue name is not being truncated. This means we wont be able to see the bounty amount

- [ ] The left margin with the icon should remain fixed, so the text doesn't move when truncated. Additionally, we should always reserve empty space for the amount badge such that the truncated amount of characters doesn't change once you type in hours.
fundpanel

e18r:
Fixes #1371 

Screenshot from 2019-10-22 04-24-01

Tested on Chromium and Firefox.

## TODO

- [ ] Wait until #1344 gets merged, then retarget this PR at `dev`

### DIFF
--- a/apps/projects/app/components/Panel/FundIssues/FundIssues.js
+++ b/apps/projects/app/components/Panel/FundIssues/FundIssues.js
@@ -727,13 +727,14 @@ const DetailsArrow = styled.div`
 `
 const IssueTitleBox = styled.div`
   grid-area: title;
-  padding: ${2 * GU}px ${2 * GU}px 0 ${2 * GU}px;
+  padding: ${2 * GU}px;
+  padding-bottom: 0;
   display: flex;
-  justify-content: flex-start;
 `
 const IssueAmountBox = styled.div`
   grid-area: amount;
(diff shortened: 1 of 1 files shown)


apps/projects/app/components/Panel/FundIssues/FundIssues.js
```diff
@@ -728,7 +729,13 @@ const IssueTitleBox = styled.div`
   grid-area: title;
   padding: ${2 * GU}px ${2 * GU}px 0 ${2 * GU}px;
   display: flex;
-  justify-content: space-between;
+  justify-content: flex-start;
+`
+const IssueAmountBox = styled.div`
+  grid-area: amount;
+  padding: ${2 * GU}px ${2 * GU}px 0 ${2 * GU}px;
```
chadoh: This might be easier to read as two rules:

```js
padding: ${2 * GU}px;
padding-bottom: 0;
```

However, I think maybe this logic would be cleaner if we moved it to the wrapping div.

```css
grid-gap: ${2 * GU}px;
padding: ${2 * GU}px;
```

PR #1408, opened 2019-10-22
Added CurrencyBox to NewBudget.js

PeterMPhillips:
When editing a budget, there shouldn't be a dropdown for the currency. Created a new CurrencyBox component that now gets imported by NewAllocations and NewBudget. NewBudget displays CurrencyBox instead of a Dropdown when it is in editing mode.

### DIFF
--- a/apps/allocations/app/components/Form/Field/CurrencyBox.js
+++ b/apps/allocations/app/components/Form/Field/CurrencyBox.js
@@ -1,0 +1,14 @@
+import styled from 'styled-components'
+
+const CurrencyBox = styled.div`
+  border: 1px solid #dde4e9;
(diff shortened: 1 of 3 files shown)


PeterMPhillips:
@chadoh is correct, we should just be using a disabled DropDown instead of creating our own component. There really isn't any point in fixing this issue with on this branch since most of the work will be undone. I'll close this PR

PeterMPhillips:
I'm not sure what happened, but upon looking at the `disabled` DropDown component again it looks totally different than the CurrencyBox component we are using. So I've reopened this PR to make the NewBudget panel to use the CurrencyBox component in 'edit' mode

PR #1410, opened 2019-10-23
Switch to using AppBadge from Aragon-UI

PeterMPhillips:
In order to do this the app's aragon-ui package has to be upgraded. This created issues when trying to use the identity provider from the shared folder. Since the Address Book app also uses an upgraded ui, I copied the LocalIdentityBadge folder from that app to this. Also, I copied the LocalLabelAppBadge from Aragon's Voting app. In the future, we can upgrade the ui's on all the other apps and then move the identity management files into the shared folder. But this works for the time being.


apps/dot-voting/app/components/VoteDetails/index.js
```diff
@@ -26,7 +26,7 @@ const VoteDetails = ({ vote, onVote }) => {
   const { voteWeights, votingPower } = useUserVoteStats(vote)
   const tokenContract = tokenAddress && api.external(tokenAddress, tokenAbi)
 
-  const network = useNetwork()
+  //const network = useNetwork()
```
chadoh: Can e delete this line, and stop importing `useNetwork`?

PeterMPhillips: Oops, meant to delete that after testing!

PR #1418, opened 2019-10-23
Address Book: update type badges colors

### LINKED ISSUE #1412: Organization and Project - font colors on the badges are not correct
See here: https://www.dropbox.com/s/iuf9xm7ug7bf9lo/Screen%20Shot%202019-10-23%20at%2011.04.02%20AM.png?dl=0

e18r:
Fixes #1412 

Uses AragonUI's theme for setting the colors of the type badges
Removes the "Project" type

### DIFF
--- a/apps/address-book/app/components/App/Entities.js
+++ b/apps/address-book/app/components/App/Entities.js
@@ -7,21 +7,20 @@ import {
   DataView,
   Tag,
   Text,
+  useTheme,
 } from '@aragon/ui'
 
 import LocalIdentityBadge from '../LocalIdentityBadge/LocalIdentityBadge'
 import { IconDelete } from '../../../../../shared/ui'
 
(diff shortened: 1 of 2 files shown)


apps/address-book/app/components/App/Entities.js
```diff
@@ -48,9 +47,12 @@ const Entities = ({ entities, onRemoveEntity }) => {
             entity={entryAddress}
           />,
           <Tag
+            style={{
+              fontWeight: 1000
+            }}
```
chadoh: Does the `css` prop also work with `Tag`? I think in general we favor that over `style`, if the component supports it. That would let us make it one line, too: `css="font-weight: 1000"`

Also, does this accomplish the same thing as `font-weight: bold`? Or is it actually bolder than that?

PR #1425, opened 2019-10-23
Rewards: miscellaneous cleanup

e18r:
Removed console.logs, solved a warning and put explicit props in App.

### DIFF
--- a/apps/rewards/app/components/App/App.js
+++ b/apps/rewards/app/components/App/App.js
@@ -333,5 +333,18 @@ const EmptyContainer = styled.div`
 // eslint-disable-next-line react/display-name
 export default () => {
   const { api, appState, connectedAccount, displayMenuButton } = useAragonApi()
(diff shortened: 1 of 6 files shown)


apps/rewards/app/components/App/App.js
```diff
@@ -333,5 +333,18 @@ const EmptyContainer = styled.div`
...
+      api={api}
+      rewards={appState.rewards}
+      myRewards={appState.myRewards}
+      metrics={appState.metrics}
+      myMetrics={appState.myMetrics}
+      amountTokens={appState.amountTokens}
+      claims={appState.claims}
+      {...appState}
+      connectedAccount={connectedAccount}
+      displayMenuButton={displayMenuButton}
```
chadoh: The newest version of the client does not need this, and the latest version of aragonAPI does not provide it. I think we can get rid of it.

PR #1431, opened 2019-10-23
Address Book: minor cleanup

chadoh:
I noticed [some][1] [minor][2] code style issues in Address Book that I wanted to clean up.

* replace tab with spaces
* get rid of superfluous `={true}`
* simplify variable names
* get rid of intermediate variable
* alphabetize props
* fix "invalid prop type: expected string, got object" error thrown by `Tag`
* change keys to simple numbers – it turns out aragonUI does not require keys at all here, but our linter does. In any case, having all the keys be the same value is not what we want.

  [1]: https://github.com/AutarkLabs/open-enterprise/pull/1274#pullrequestreview-305441192
  [2]: https://github.com/AutarkLabs/open-enterprise/pull/1418#pullrequestreview-306026325

### DIFF
--- a/apps/address-book/app/components/App/Entities.js
+++ b/apps/address-book/app/components/App/Entities.js
@@ -18,8 +18,15 @@ const entitiesSort = (a, b) => a.data.name.toUpperCase() > b.data.name.toUpperCa
 const Entities = ({ entities, onRemoveEntity }) => {
   const theme = useTheme()
   const ENTITY_TYPES = [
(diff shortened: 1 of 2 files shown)


chadoh:
> I don't really like how we need to return an array of components here, being an array it what requires the key prop. It seems a kind of anti-pattern for me, but 🤷‍♂️ is aragon-ui so not much we can do

I agree, at first I put this in a `React.Fragment` but that doesn't work with DataView.

rkzel:
Tested, working.

PR #1433, opened 2019-10-23
Updated Dot Voting permission in template

stellarmagnet:
Add candidate options should be assigned to Voting, not Tokens.

### DIFF
--- a/templates/open-enterprise/contracts/BaseOEApps.sol
+++ b/templates/open-enterprise/contracts/BaseOEApps.sol
@@ -116,9 +116,8 @@ contract BaseOEApps is BaseCache, TokenCache {
     )
         internal
     {
-        //TODO: we should pass _tokenManager into ROLE_CREATE_VOTES as 2nd param, not _dotVoting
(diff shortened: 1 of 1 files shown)


templates/open-enterprise/contracts/BaseOEApps.sol
```diff
@@ -116,9 +116,8 @@ contract BaseOEApps is BaseCache, TokenCache {
     )
         internal
     {
-        //TODO: we should pass _tokenManager into ROLE_CREATE_VOTES as 2nd param, not _dotVoting
         _acl.createPermission(_grantee, _dotVoting, _dotVoting.ROLE_CREATE_VOTES(), _manager);
-        _acl.createPermission(_grantee, _dotVoting, _dotVoting.ROLE_ADD_CANDIDATES(), _manager);
+        _acl.createPermission(_manager, _dotVoting, _dotVoting.ROLE_ADD_CANDIDATES(), _manager);
```
stellarmagnet: instead of _manager and _grantee I think it will be cleaner to do...

_manager = _voting
_grantee = _tokenManager

PR #1448, opened 2019-10-24
Updated no data copy and app icons

stellarmagnet:


apps/dot-voting/app/App.js
```diff
@@ -63,7 +62,7 @@ const Empty = () => (
     >
       <EmptyStateCard
         title="You do not have any dot votes."
-        text="Use the Allocations app to get started."
+        text="After you create an allocation or issue curation, you can vote here."
```
ottodevs: I was waiting for this change so bad but always forgot to make it haha ❤️

PR #1452, opened 2019-10-25
Minor FilterBar UI fixes

rkzel:
Split FilterDropDown into FDD and Overflow components. Also, fixed padding in items lists and added very light grey as background color to artificial "labelless" and "milestoneless" items.

### DIFF
--- a/apps/projects/app/components/Shared/FilterBar/FilterDropDown.js
+++ b/apps/projects/app/components/Shared/FilterBar/FilterDropDown.js
@@ -7,49 +7,107 @@ import { springs, GU, theme } from '@aragon/ui'
 import FilterButton from './FilterButton'
 import { IconMore, IconDropArrow } from '../../../assets'
 
-const FilterDropDown = ({ caption, children, enabled, type }) => {
+const borders = {
+  'overflowTop': '3px 3px 0 0',
+  'overflowMiddle': '0',
+  'overflowBottom': '0 0 3px 3px',
+  'filter': '3px',
+}
+
+const SpringWrap = ({ handleClickOut, opened, children }) => (
+  <ClickOutHandler onClickOut={handleClickOut}>
+    <Spring
+      config={springs.smooth}
+      to={{ openProgress: Number(opened) }}
+      native
+    >
+      {({ openProgress }) => children(openProgress)}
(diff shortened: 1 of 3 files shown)


apps/projects/app/components/Shared/FilterBar/FilterDropDown.js
```diff
@@ -7,130 +7,117 @@ import { springs, GU, theme } from '@aragon/ui'
...
+  )
+}
+FilterDropDown.propTypes = {
   caption: PropTypes.string.isRequired,
-  openProgress: PropTypes.object.isRequired,
-  handleClose: PropTypes.func.isRequired,
-  type: PropTypes.string.isRequired,
   children: PropTypes.node.isRequired,
+  enabled: PropTypes.bool.isRequired,
+  type: PropTypes.string.isRequired,
```
chadoh: It no longer uses `type`, right?

apps/projects/app/components/Shared/FilterBar/prepareFilters.js
```diff
@@ -66,6 +67,7 @@ const prepareFilters = (issues, bountyIssues) => {
       } else {
         filters.labels['labelless'] = {
           name: 'Issues without labels',
+          color: 'F2F2F2',
```
chadoh: Should these values come from `theme`?

apps/projects/app/components/Shared/FilterBar/FilterDropDown.js
rkzel: Indeed. It could be in this PR, but next makes it.. different. Next PR has dynamic Overflow.

rkzel: Removed `type` for now.

apps/projects/app/components/Shared/FilterBar/prepareFilters.js
rkzel: I wasn't sure what it should be - this color is for Tags that are artificially created for labelless issues. As suggested on #design I changed that to a) not include color at all, which b) signifies that an outline is needed instead (in FilterBar itself).

PR #1461, opened 2019-10-25
Allocations: set a fixed width for recipients container

### LINKED ISSUE #1419: Allocations history -- the bar width should be decoupled from the identity badge width
Screenshot of the problem:
image

e18r:
Fixes #1419 

Screenshot from 2019-10-25 13-38-28

### DIFF
--- a/apps/allocations/app/components/App/AllocationsHistory.js
+++ b/apps/allocations/app/components/App/AllocationsHistory.js
@@ -6,6 +6,7 @@ import {
   IconCross,
   ProgressBar,
   Text,
+  useLayout,
   useTheme,
 } from '@aragon/ui'
 import LocalIdentityBadge from '../LocalIdentityBadge/LocalIdentityBadge'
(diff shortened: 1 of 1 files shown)


apps/allocations/app/components/App/AllocationsHistory.js
```diff
@@ -74,7 +74,7 @@ const AllocationsHistory = ({ allocations }) => {
         return recipients.map((recipient, index) => {
           const allocated = BigNumber(recipient.supports).div(totalSupports)
           return (
-            <div key={index}>
+            <RecipientContainer key={index}>
```
chadoh: While we're here, I'd love to get rid of the unnecessary `index` variable and use `key={recipient.candidateAddress}` instead.

apps/allocations/app/components/App/AllocationsHistory.js
```diff
@@ -123,6 +123,10 @@ const Amount = styled.div`
   font-weight: 600;
 `
 
+const RecipientContainer = styled.div`
+  width: 50%;
```
chadoh: How does this look on mobile?

topocount: @chadoh this looks okay for me. The location of the detail expansion in this block bothers me a bit, but that's not enought to prevent a merge in my opinion

image

PR #1464, opened 2019-10-25
Switched position of widgets and removed redundant 'Status' title

PeterMPhillips:


rkzel:
I added padding to the Boxes to make it as per design. Otherwise - good! :)

PR #1465, opened 2019-10-25
Added token value per receiver of allocations

PeterMPhillips:
If there is a dot vote for dispersing allocations, after the vote is completed it shows the amount given to each recipient in parentheses after their percentage.


apps/dot-voting/app/components/VotingOption.js
```diff
@@ -60,11 +60,18 @@ const VotingOption = ({ valueSpring, label, percentage, color, threshold, userVo
...
-              {Math.round(percentage)}%
+        <div>
+          {percentage !== -1 &&
+              <Text size="xsmall" color="#98A0A2">
+                {Math.round(percentage)}%
+              </Text>
+          }
+          {allocation && (
+            <Text size="xsmall" color="#98A0A2" css={`margin-left: ${0.25 * GU}px`}>
+                ({allocation})
```
rkzel: Can we use theme for the colors, please?

PeterMPhillips: Done. I also noticed another component that used a hardcoded colour so I changed that as well.

e18r: @PeterMPhillips, that way of importing `theme` is deprecated. The new way of doing it is by importing the `useTheme` hook,

    import { useTheme } from '@aragon/ui'

and then, inside a functional component,

     const theme = useTheme()

apps/dot-voting/app/components/VoteDetails/index.js
```diff
@@ -109,11 +108,12 @@ const VoteDetails = ({ vote, onVote }) => {
                 voteWeights={voteWeights}
                 votingPower={votingPower}
               />
-            ) :(
+            ) : (
               <VotingResults
                 vote={vote}
                 options={vote.data.options}
                 voteWeights={voteWeights}
+                balance={(type === 'allocation') ? BigNumber(vote.data.balance).div(BigNumber(10 ** decimals)).toNumber() : 0}
```
e18r: No need for this prop because `VotingResults` already has access to `vote.data.type` and `vote.data.balance`.

apps/dot-voting/app/components/VoteDetails/VotingResults.js
```diff
@@ -32,6 +32,8 @@ const VotingResults = ({ vote, voteWeights }) => {
           totalSupport={totalSupport}
           color={`${theme.accent}`}
           voteWeights={voteWeights}
+          balance={!vote.open ? balance : 0}
+          symbol={(!vote.open && vote.data.tokenSymbol) ? vote.data.tokenSymbol : ''}
```
e18r: These props seem overloaded with meaning to me. What do I mean by that? Not only you are passing down the balance and the symbol to `VotingOptions`, but also you are encoding in these two props the information of whether to show the allocations or not. I think this is error-prone because for instance what happens if `vote.open` is true but `vote.data.tokenSymbol` is falsey? Does that mean show or not show the allocations?

Maybe it makes more sense to make these checks explicit by including a `showAllocations` props set to `true` if `vote.open` and `vote.data.type === 'allocation'` (and `vote.data.tokenSymbol` is not falsey? I'm not sure). That way we can condense the logic for showing / not showing allocations in a single place and stop overloading these props. You can just pass `balance` and `vote.data.tokenSymbol` without having to check them.

apps/dot-voting/app/components/VotingOptions.js
```diff
@@ -52,16 +57,21 @@ class VotingOptions extends React.Component {
...
-                valueSpring={value}
-                percentage={safeDiv(parseInt(option.value, 10), totalSupport)*100}
-                color={color}
-                userVote={(voteWeights.length && displayYouBadge) ? Math.round(voteWeights[i]) : -1}
-                {...option}
-              />
-            )}
+            {({ value }) => {
+              const percentage = safeDiv(parseInt(option.value, 10), totalSupport)
+              const allocation = symbol ? `${BigNumber(balance).times(percentage).dp(3).toString()} ${symbol}` : ''
```
e18r: I think these two calculations will eventually end up in the app state reducer, which I believe is the best place for them. There, they can be included in `vote.options` reducing the complexity here. But that's for another PR.

Here, you can check `showAllocations` instead of `symbol`, and if it's true, pass it as a string with the parentheses already in place; if it's false, pass an empty string. That way you remove the check you have to make in `VotingOption`.

apps/dot-voting/app/components/VotingOption.js
```diff
@@ -60,11 +60,18 @@ const VotingOption = ({ valueSpring, label, percentage, color, threshold, userVo
...
-        {percentage !== -1 &&
-            <Text size="xsmall" color="#98A0A2">
-              {Math.round(percentage)}%
+        <div>
+          {percentage !== -1 &&
+              <Text size="xsmall" color={theme.textTertiary}>
+                {Math.round(percentage)}%
+              </Text>
+          }
+          {allocation && (
```
e18r: If you already pass either a string with parentheses or an empty string, you can remove this check. See my comment in `VotingOptions.js`.

apps/dot-voting/app/components/VotingOption.js
PeterMPhillips: I tried that. But useTheme() doesn't pass any text values, and I need textTertiary. This is probably why it was hardcoded in the first place.

apps/dot-voting/app/components/VoteDetails/Status.js
```diff
@@ -2,12 +2,12 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import styled from 'styled-components'
 import { format } from 'date-fns'
-import { Box, Countdown, GU } from '@aragon/ui'
+import { Box, Countdown, GU, theme } from '@aragon/ui'
```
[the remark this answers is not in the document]
e18r: As you point out in another comment, `textTertiary` doesn't exist anymore. The new equivalent is `contentSecondary`.

apps/dot-voting/app/components/VotingOption.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react'
 import styled from 'styled-components'
-import { GU, Tag, Text } from '@aragon/ui'
+import { GU, Tag, Text, theme } from '@aragon/ui'
```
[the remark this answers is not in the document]
e18r: ditto, with `contentSecondary`.

apps/dot-voting/app/components/VotingOptions.js
```diff
@@ -16,6 +16,7 @@ class VotingOptions extends React.Component {
     voteOpen: true,
     balance: 0,
     symbol: '',
+    decimals: '18',
```
e18r: Screenshot from 2019-11-11 10-53-30

PR #1472, opened 2019-10-28
Dot Voting: fix stubbedApi logic

chadoh:
* add executionTargetData, required since #1326 
* return string for `voteId`, to satisfy `propTypes` expectations
* add resolveAddressIdentity to shared/api-react
* handle some uses of `call`, the lack of which caused noisy warnings
* update imports from `@aragon/api-react` to local stub, which resolves more warnings

### DIFF
--- a/apps/dot-voting/app/api-react.js
+++ b/apps/dot-voting/app/api-react.js
@@ -1,3 +1,4 @@
+import { Observable } from 'rxjs'
 import buildStubbedApiReact from '../../../shared/api-react'
 import { ETHER_TOKEN_FAKE_ADDRESS } from './utils/token-utils'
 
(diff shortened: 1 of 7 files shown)


ottodevs:
How is it supposed to work?

I did the following steps:
- Fetch, checkout and pull this branch
- Run `npm prune && npm i`(just in case)
- `npm run frontend:dot`

Then I receive this error, and the screen is completely empty:
image

As you can see I tried to debug some variables and yeah as you can see, the issue is that executionTargetData is not present in the vote.data:
image

Also, is it really needed to have these duplications? like metadata and description, for example...

I tried to create some votes but does not change the situation.

After some testing, I found that in brave `api.createVote()` returns an observable, while in firefox it works as it should (creating the vote), I fix in brave by changing it to: `await api.createVote().toPromise()`

I am not really sure if the actual issue is because there is some old cache, although I tried to clean the chache.

PR #1476, opened 2019-10-28
Allocations: Remove "Deactivate" option

### LINKED ISSUE #1474: Remove "deactivate" action from Budget card
Right now, we haven't fully implemented this feature, so we should remove it from the UI

chadoh:
Fixes #1474

This removes the "Deactivate" option from the Budget context menu, as well as everything that's there to support it. This might be more than we need, but if we don't have firm plans for when we will start supporting "deactivate", then I think it's best to err toward "remove more" rather than "remove less".


chadoh:
Closing in favor of #1477

PR #1478, opened 2019-10-28
Rewards: help bubbles

### LINKED ISSUE #1466: Add help icons for Rewards input fields
The current implementation is out of sync with Figma & Design specs.

We should upgrade to the new aragon-ui, if we haven't and develop the help tips.

e18r:
**TODO before merging**
- [x] Wait for #1473 to be merged
- [x] Redirect this PR towards `dev`

Added the help bubbles in the New Rewards side panel according to the design. This requires as a prerrequisite an upgrade in AragonUI created in #1473. This fixes #1466.

 
Screenshot from 2019-10-28 13-42-41
Screenshot from 2019-10-28 13-42-53


apps/rewards/app/components/Form/Field/FormField.js
```diff
@@ -32,12 +32,20 @@ const FormField = ({
...
           </Text>
         )}
+        {hint && (
+          <div
+            style={{
+              marginLeft: '0.3rem',
+              display: 'flex',
+              alignItems: 'center',
+            }}
+          >
```
PeterMPhillips: It doesn't look like `display` and `alignItems` have any effect here

e18r:
@PeterMPhillips I agree with you about the help bubbles being too visually dominant. However, aligning them to the right goes against the design. What does @jayalaves say about it?

apps/rewards/app/components/Form/Field/FormField.js
e18r: Oh, but they do. This is how the help bubbles look without them:
Screenshot from 2019-10-31 14-15-13

And this is how they look with them:
Screenshot from 2019-10-31 14-16-29

Subtle, I agree. Without these CSS props, the bubble is a little too high:
Screenshot from 2019-10-31 14-17-23

Whereas with the props:
Screenshot from 2019-10-31 14-20-13

PeterMPhillips: oh nice! I didn't notice at first, but I completely agree now that you've pointed it out.

PR #1479, opened 2019-10-28
Made work status details arrange dynamically

PeterMPhillips:
Added a component that put work details into a separate `<div>` and removes preceding 'dot' if the 'compact' prop is set to true. This way the details wrap nicely onto a new line by setting 'compact' to true on small screen sizes.

### DIFF
--- a/apps/projects/app/components/Card/Issue.js
+++ b/apps/projects/app/components/Card/Issue.js
@@ -2,7 +2,7 @@ import PropTypes from 'prop-types'
 import React from 'react'
 import styled from 'styled-components'
 
-import { Text, Tag, Checkbox, ContextMenu, useTheme, IconClock, IconConnect, IconCalendar } from '@aragon/ui'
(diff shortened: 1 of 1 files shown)


chadoh:
@PeterMPhillips please add before and after screenshots for UI changes like this

apps/projects/app/components/Card/Issue.js
```diff
@@ -27,8 +27,27 @@ const labelsTags = (labels, theme) =>
     </Tag>
   ))
 
+const FlexibleDiv = ({ compact, children }) => {
+  return compact ? (
+    <div>
+      {children}
+    </div>
+  ) : (
```
chadoh: I find ternaries like this a bit confusing. Why not this?

```js
if (compact) return <div>children</div>

return (
  <React.Fragment>
    {dot}
    {children}
  </React.Fragment>
)
```

I'm also wondering if we could do something like this instead:

```js
return (
  <React.Fragment>
    {compact ? <br /> : dot}
    {children}
  </React.Fragment>
)
```

PR #1480, opened 2019-10-28
Fixed panel form so datepicker attaches correctly

PeterMPhillips:


shared/ui/components/form/inputs/DatePicker.js
```diff
@@ -244,7 +259,11 @@ DatePicker.propTypes = {
   monthFormat: PropTypes.string,
   monthYearFormat: PropTypes.string,
   weekDayFormat: PropTypes.string,
-  yearFormat: PropTypes.string
+  yearFormat: PropTypes.string,
+
+  // Positioning
```
e18r: The fact that you made a comment here could be a sign that the prop names you chose, `horizontal` and `vertical`, might not be the most explanatory. How about `horizontalAlign` and `verticalAlign`? Too long? Maybe `halign` and `valign`? Or `hAlign` and `vAlign`?

apps/rewards/app/components/Panel/NewReward/NewReward.js
```diff
@@ -358,10 +362,12 @@ class NewRewardClass extends React.Component {
       />
       <FormField
         label="End date"
+        width={`calc(50% - ${GU}px)`}
```
e18r: This is a nice improvement. But in what way is it related to the issue at hand?

apps/rewards/app/components/Panel/NewReward/NewReward.js
```diff
@@ -339,10 +340,13 @@ class NewRewardClass extends React.Component {
     <HorizontalContainer>
       <FormField
         label="Start date"
+        width={`calc(50% - ${GU}px)`}
         required
         input={
           <DateInput
             name="dateStart"
+            horizontal="left"
```
e18r: Why define a `vertical` prop if it's not going to be used?

shared/ui/components/form/inputs/DatePicker.js
PeterMPhillips: I made a comment here because all the other props were categorized in such a way and I was keeping it consistent. Happy to change the names to verticalAlign and horizontalAlign if that is the preference.

apps/rewards/app/components/Panel/NewReward/NewReward.js
PeterMPhillips: It was really low hanging fruit for a big visual improvement on mobile screens. Why wouldn't I fix it?

apps/rewards/app/components/Panel/NewReward/NewReward.js
PeterMPhillips: For future use cases. One issue I keep encountering is that the the DatePicker overflows the element engaging the scrollbar. If instead we allow it to go up instead of down, the user wont have to scroll down to use the DatePicker. But at the moment I though it best to not engage in making those design decision and left the current behaviour as is.

shared/ui/components/form/inputs/DatePicker.js
e18r: Oh, that's true, I hadn't noticed that. Well, only if my comment makes sense to you.

topocount: @e18r it looks like the Projects app uses its own DatePicker right now and not the shared one, so testing in Projects is superfluous at the moment. I'll log an issue for that shortly

apps/rewards/app/components/Panel/NewReward/NewReward.js
e18r: Fair enough. It's just confusing for a reviewer to understand where you were coming from, so I think it's a good practice to provide a small explanation in the PR to help review it. Because as a reviewer you're out of context and it's hard to tell the really meaningful changes from other improvements, and the less changes the better.

Also bear in mind that sometimes other people notice the same low-hanging fruit as you and fix it in another PR that gets merged more or less at the same time as yours, which could cause avoidable conflicts. So the best practice is to create a dedicated PR, especially if the change is big, or at least check that there is no other pending PR that solved it already.

apps/rewards/app/components/Panel/NewReward/NewReward.js
e18r: I see what you mean. Good call!

apps/rewards/app/components/Panel/NewReward/NewReward.js
PeterMPhillips: Understood. I'll try to separate those sorts of changes into their own PRs. Its been hard to avoid fixing small issues as they come up in unrelated issues but you make a valid point :)

chadoh: Something that helps me with things like this: [`hub`](https://github.com/github/hub). Here's my basic workflow:

* When I stage changes, I always use `git add -p`
* I stage only things that are relevant to the current effort
* Commit them
* `git checkout dev && git checkout -b new-branch-name-here`
* Then I `git add -p` again, stage & commit the unrelated changes – make the commit message the same thing you want to say in your PR title + description
* `git push`
* `git pull-request`

Then, with a few commands that become pretty automatic once you do them enough, you can quickly push all those good cleanup changes to their own dedicated PR.

PR #1483, opened 2019-10-29
Dot Voting: allow voting with number input

chadoh:
Rather than only showing the vote value in a grey box that looks like an input field, this uses an actual input field. The amounts can be edited from either the numeric input or the slider.

slide or type

## TODO

- [x] Wait for #1472 merge; retarget this one at `dev`


chadoh:
> It won't let you type a number larger than remaining, but it won't let you clear the box either and you're forced to fall back to the slider to make the box usable again

Good find, @topocount. I'm surprised that it doesn't let you clear the box!

PR #1497, opened 2019-10-30
Streamline publish flow

topocount:
Apps now only publish when there are changes to the app, contract, or shared files. Contract changes trigger a major version bump, app file or shared file changes trigger a minor bump, and no changes don't trigger anything!

I've also removed the compile step from the build command since the build command is called only for frontend builds by the CLI. No other processes we maintain depend upon contract compilation when the build command is called, so removing it saves us time during deployments and local dev builds

**NOTE** I have opted not to continue propagating files from CI because it is a time and resource-intensive process, and also because the publish commands in their current form pin files to our VPS. So for CD testing purposes, we just need to redirect the client in the aragon client settings to utilize our VPS and access our dev builds for testing


shared/deployments/check-publish.sh
```diff
@@ -0,0 +1,23 @@
...
+
+if [ $? == "1" ]
+then
+  aragon apm publish major --files dist/ --environment continuous-deployment --apm.ipfs.rpc https://ipfs.autark.xyz:5001 --ipfs-check false --propagate-content false --skip-confirmation --no-prepublish
+  exit $?
+else 
+  echo "no contract changes"
+fi
+
+git diff --quiet HEAD~ HEAD -- $PWD/app $PWD/../../shared/identity $PWD/../../shared/lib $PWD/../../shared/store-utils $PWD/../../shared/ui $PWD/../../shared/utils
```
chadoh: It seems a little bit strange that every single app's `publish:cd` process will check this same set of directories (other than `PWD/app`). But maybe that's a good thing?

shared/deployments/check-publish.sh
```diff
@@ -0,0 +1,23 @@
...
+
+if [ $? == "1" ]
+then
+  aragon apm publish minor --files dist/ --environment continuous-deployment --apm.ipfs.rpc https://ipfs.autark.xyz:5001 --ipfs-check false --propagate-content false --skip-confirmation
+  exit $?
```
chadoh: Why exit with an error status?

shared/deployments/publish-cd.sh
```diff
@@ -1,12 +0,0 @@
-#!/bin/bash
-IFS=$'\n'
-APP_CIDs=($(lerna run publish:cd --no-ci | grep -o "Qm[[:alnum:]]\{44\}$"))
-unset IFS
-echo number of apps published: ${#APP_CIDs[*]}
-
-for cid in "${APP_CIDs[@]}"
-do
-  echo propagating $cid ... $'\n'
-  aragon ipfs propagate $cid > /dev/null &
```
chadoh: Seems like this propagation did not make it into the new root `publish:cd` command. Is that ok?

shared/deployments/check-publish.sh
topocount: Yeah this makes it much easier to maintain this flow. The other directories are all shared, and out of an abundance of caution, and due to the fact that minor and patch publishes cost very little gas and are quick, it makes sense to publish all apps in CD if a shared file is updated.

shared/deployments/check-publish.sh
topocount: This exits with the error status of the publish command. If it's zero, it will exit with zero. Otherwise it'll throw an error and Travis will reflect the result. The current publish script show that the publish was successful even if an app errors on publishing, which isn't very helpful

shared/deployments/publish-cd.sh
topocount: as mentioned above: 
>NOTE I have opted not to continue propagating files from CI because it is a time and resource-intensive process, and also because the publish commands in their current form pin files to our VPS. So for CD testing purposes, we just need to redirect the client in the aragon client settings to utilize our VPS to consume our dev builds for testing

chadoh: Great, thank you for the reminder!

So, this means that anyone who tests these CD builds will need to configure their client to point at our IPFS endpoint, or assets will not load properly. Is that correctL

topocount: yeah. Yalda and Javi already do this, and we also make it a point to pin the latest aragon app files to this node as well

shared/deployments/check-publish.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+git diff --quiet HEAD~ HEAD -- $PWD/contracts
```
ottodevs: ❤️

PR #1499, opened 2019-10-30
Added caching and syncing to apps

PeterMPhillips:
This merge ensures that all apps use caching in their background scripts and display a syncing message while cached events are being loaded. Generally, this means editing 4 files for each app:

app/store/init.js:
- This file sets the initial state with a `isSyncing` variable set to `false`.
- Initializes the state with the `cachedState` if available

app/store/events.js:
- The events file must watch for two events: `SYNC_STATUS_SYNCING` to set `isSyncing = true` and `SYNC_STATUS_SYNCED` to set `isSyncing = false`

app/components/App/App.js:
- The App.js file gets the `isSyncing` variable from the `appState` and passes it to the Empty card. Also, the SyncIndicator component is added to the main screen and is set to visible when `isSyncing == true`

app/components/Card/Empty.js
- This card now takes an `isSyncing` bool prop and displays a syncing message instead of the regular message if `isSyncing == true`

To test that the app correctly displays a syncing message you'll have make some calls to the app contract and then mine 100 blocks by calling `npm run mine 100` in the terminal. Then clear the cache and wait for the app to reload.

### DIFF
--- a/apps/address-book/app/components/App/App.js
+++ b/apps/address-book/app/components/App/App.js
@@ -1,7 +1,7 @@
 import React, { useState } from 'react'
 
 import { useAragonApi } from '../../api-react'
-import { Button, Header, IconPlus, Main, SidePanel } from '@aragon/ui'
+import { Button, Header, IconPlus, Main, SidePanel, SyncIndicator } from '@aragon/ui'
 
 import { IdentityProvider } from '../LocalIdentityBadge/IdentityManager'
 import { ipfsAdd } from '../../../../../shared/utils/ipfs'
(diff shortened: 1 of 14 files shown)


topocount:
> Working, thank you :) One thing I noticed though is that there are differences in behavior between apps:
> 
> AB displays "Syncing..." in Empty Card.
> Allocations displays "Syncing.." and SyncIndicator at the same time.
> DV is the same as AB,
> Projects while syncing is displaying Empty Card which is different from usual Empty Card - different in size too, so it kind of flickers...
> Rewards is the same as DV and AB.
> 
> I'm not sure if those discrepancies are something we need to address right now?

@stellarmagnet this might be a good one for you to test locally if possible. If not possible, we can merge it and adjust it as needed.

PeterMPhillips:
Just a note, as per my conversation with @stellarmagnet  on Keybase, I changed the language of the sign in card for the projects app.

PR #1500, opened 2019-10-31
Rewards: allow past dates after reference asset creation

### LINKED ISSUE #1489: Improve date validation for New Reward
### One-time Merit
* Check with @topocount with whether they can be in the past

### One-time Dividend
* We should allow reference dates that are in the past or the same day.
* We should not allow the reference date to be a date prior to when the reference token was created: `The selected reference date occurs before the reference asset ANT was created. Please choose another date.`
   * ... Where `ANT` is the token symbol.

### Recurring Dividend
* We should allow start & end dates that are in the past or the same day.
* We should not allow either of these dates to be prior to when the reference token was created: `The selected reference date occurs before the reference asset ANT was created. Please choose another date.`

For all use cases, please test that when you do create a new reward for a date in the past (and it is a valid date in the past), that the new reward is:
* Properly created
* Properly claimed

e18r:
Fixes #1489 

Screenshot from 2019-10-30 21-50-21

### DIFF
--- a/apps/rewards/app/components/App/App.js
+++ b/apps/rewards/app/components/App/App.js
@@ -156,6 +156,7 @@ class App extends React.Component {
   }
 
   onNewReward = async reward => {
+    const BLOCK_PADDING = 1 // ensuring the start block is different from the end block
     let currentBlock = await this.props.api.web3Eth('getBlockNumber').toPromise()
     const amountBN = new BigNumber(reward.amount)
     const tenBN =  new BigNumber(10)
@@ -190,8 +191,9 @@ class App extends React.Component {
       startBlock -= reward.duration
     }
     if(reward.rewardType === ONE_TIME_DIVIDEND){
-      startBlock = currentBlock
-      reward.duration = millisecondsToBlocks(Date.now(), reward.dateReference)
+      const rawBlockDuration = millisecondsToBlocks(Date.now(), reward.dateReference)
+      startBlock = reward.dateReference <= new Date() ? currentBlock + rawBlockDuration - BLOCK_PADDING : currentBlock 
+      reward.duration = reward.dateReference <= new Date() ? BLOCK_PADDING : rawBlockDuration
     }
     this.props.api.newReward(
       reward.description, //string _description
@@ -282,7 +284,6 @@ class App extends React.Component {
         </Wrapper>
       )
     }
-
     return (
       <Wrapper>
         <Header
--- a/apps/rewards/app/components/Panel/NewReward/NewReward.js
+++ b/apps/rewards/app/components/Panel/NewReward/NewReward.js
@@ -17,6 +17,7 @@ import {
 import { Form, FormField } from '../../Form'
 import { DateInput } from '../../../../../../shared/ui'
 import moment from 'moment'
+import { isBefore } from 'date-fns'
 import { isAddress } from '../../../utils/web3-utils'
 import { ETHER_TOKEN_VERIFIED_BY_SYMBOL } from '../../../utils/verified-tokens'
 import TokenSelectorInstance from './TokenSelectorInstance'
@@ -76,6 +77,7 @@ const INITIAL_STATE = {
     dateEndPassed: 'End date must take place after today.',
     dateStartAfterEnd: 'Start date must take place before the end date.',
     singleDisbursement: 'While you selected a recurring dividend, based on your parameters, there will only be a single disbursement.',
+    dateBeforeAsset: 'The selected ${dateType} date occurs before the reference asset, ${tokenSymbol}, was created. Please choose another date.',
   },
 }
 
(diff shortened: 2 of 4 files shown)


apps/rewards/app/components/Panel/NewReward/NewReward.js
```diff
@@ -172,31 +174,82 @@ class NewRewardClass extends React.Component {
...
+          && isBefore(state.dateReference, tokenCreationDate)) {
+        semanticErrors.push({
+          name: 'dateBeforeAsset',
+          params: {
+            dateType: 'reference',
+            tokenSymbol: tokenSymbol,
+          }
+        })
+      }
+    }
```
chadoh: We don't necessarily need to change anything for now, but I find all of this deeply-nested logic impossible to follow. 

* if the reference asset is not null and reward type is not null
  * if reference asset is OTHER and customToken is verified
    * use the symbol and creationDate from the customToken
  * else (if reference asset is OTHER but customToken is unverified, or if reference asset is anything else)
    * use the symbol and creationDate from the reference token that has the same address as the reference asset
  * furthermore, if rewardType is recurring dividend or one-item merit
    * if start date is before creationDate, add error for start date
    * if end date is before creationDate, add error for end date
  * finally, if rewardType is one-time dividend and date reference is before creationDate, add error for dateReference

Phew! Did I get that right?

In my experience, these sorts of complicated logic trees almost always have hidden bugs, and generate new bugs frequently. 

Again, I'm not sure how to fix it right now, but I want to note this complexity, so we can start thinking about how we might be able to restructure this component to allow simplified logic here.

apps/rewards/app/components/Panel/NewReward/NewReward.js
```diff
@@ -76,6 +77,7 @@ const INITIAL_STATE = {
     dateEndPassed: 'End date must take place after today.',
     dateStartAfterEnd: 'Start date must take place before the end date.',
     singleDisbursement: 'While you selected a recurring dividend, based on your parameters, there will only be a single disbursement.',
+    dateBeforeAsset: 'The selected ${dateType} date occurs before the reference asset, ${tokenSymbol}, was created. Please choose another date.',
```
chadoh: While we're here, maybe we could move `messages` outside of component state, since it's never being changed. And if we restructure it a little bit, we don't need custom string-parsing logic:

```js
const messages = {
  customTokenInvalid: () => 'Token address must be of a valid ERC20 compatible clonable token.',
  meritTokenTransferable: () => 'Merit rewards must be non-transferable.',
  amountOverBalance: () => 'Amount must be below the available balance.',
  dateReferencePassed: () => 'Reference date must take place after today.',
  dateStartPassed: () => 'Start date must take place after today.',
  dateEndPassed: () => 'End date must take place after today.',
  dateStartAfterEnd: () => 'Start date must take place before the end date.',
  singleDisbursement: () => 'While you selected a recurring dividend, based on your parameters, there will only be a single disbursement.',
  dateBeforeAsset: (dateType, tokenSymbol) => `The selected ${dateType} date occurs before the reference asset, ${tokenSymbol}, was created. Please choose another date.`,
}
```

Then we can use it like this, and get rid of the whole `showMessage` function:

```js
messages[error.name](...error.params)
```

---

Backing up, I sort of wonder if this whole "using keyed message names" is worth it. Once we have i18n, maybe we would want something like this. But until then, why not push the error messages themselves directly into an `errors` array? Like, in `setSemanticErrors` (which I'd rename to `setErrors`), why not something like this:

```js
if (state.rewardType === ONE_TIME_MERIT && state.transferable) {
  errors.push('Merit rewards must be non-transferable.')
}
// etc, etc
```

Then we could get rid of the whole `messages` thing, get rid of the whole `showMessage` thing, and in our `errorBlocks` and `warningBlocks` functions, we would loop over simple arrays of strings.

Am I missing something?

e18r: @chadoh, you're not missing anything. On the contrary, I agree with you that this logic is too complex and I appreciate your suggestions to make it better. The last part, though, I don't quite agree. I think in general strings shouldn't be part of code, and the less entangled with code, the better, so that it's easier to change them and make sure they are consistent with eachother.

apps/rewards/app/components/Panel/NewReward/NewReward.js
e18r: I don't think you got it right, which proves your point :)

Let me rework this with your suggestions. I have a few ideas.

apps/rewards/app/components/Panel/NewReward/NewReward.js
```diff
@@ -168,89 +166,60 @@ class NewRewardClass extends React.Component {
...
         )
       ) &&
-        semanticErrors.length === 0
+        errors.length === 0
     )
     return valid
   }
 
-  setSemanticErrors = (changed) => {
+  // FIXME: What exactly is this doing? What's a good name for it?
```
chadoh: @topocount do you know the answer to this question?

topocount: I think `getReferenceToken` would be a better name

apps/rewards/app/components/App/App.js
```diff
@@ -190,8 +191,9 @@ class App extends React.Component {
       startBlock -= reward.duration
     }
     if(reward.rewardType === ONE_TIME_DIVIDEND){
-      startBlock = currentBlock
-      reward.duration = millisecondsToBlocks(Date.now(), reward.dateReference)
+      const rawBlockDuration = millisecondsToBlocks(Date.now(), reward.dateReference)
+      startBlock = reward.dateReference <= new Date() ? currentBlock + rawBlockDuration - BLOCK_PADDING : currentBlock 
+      reward.duration = reward.dateReference <= new Date() ? BLOCK_PADDING : rawBlockDuration
```
topocount: @chadoh how ugly is this? This feels a bit like bike-shedding but should I just switch to an `if-else` control structure?

apps/rewards/app/components/Panel/NewReward/NewReward.js
e18r: OK, done. Let me know what you think, @chadoh

e18r:
What's the state of this PR? How can I be helpful?

apps/rewards/app/components/App/App.js
chadoh: I think an `if/else` would be easier to follow!

chadoh:
@e18r I think all that remains is that last bit of cleanup that Kevin commented about.

@topocount you tested everything and feel confident all branches of the logic work as expected?

PR #1534, opened 2019-11-04
Projects: request minimal set of permissions

### LINKED ISSUE #1528: Explain what granting GitHub permission to the Projects app entails
When I go to log in to GitHub from the Projects app, it seems to be asking for quite a lot of permissions that I would frankly be entirely uncomfortable granting to any centralized app. The Projects app documentation does not say anything about how this permission is used or what it even really means to grant this access permission.

It would be nice if this were explained so I can understand the risks associated with granting such invasive permissions to the Projects app.

https://autark.gitbook.io/open-enterprise/apps/projects

Screenshot_2019-11-02 Build software better, together

chadoh:
Fixes #1528

* Add help icon explaining why we need GitHub permissions

  initial github auth help icon

* At first sign-in, request "no scopes", which gives only read access to public data. This is enough for viewing repositories + their issues, which is enough for most Projects functionality – creating & fulfilling bounties, reviewing bounty submissions, curating issues

  initial github auth

* When creating new issues, explain [extra permissions required](https://developer.github.com/apps/building-oauth-apps/understanding-scopes-for-oauth-apps/#available-scopes), and allow user to give more permissions at this time. This is the only Projects action that requires increased permissions.

  extra permissions for creating issues

### DIFF
--- a/apps/projects/app/components/Panel/NewIssue/AuthorizeGitHub.js
+++ b/apps/projects/app/components/Panel/NewIssue/AuthorizeGitHub.js
@@ -1,0 +1,89 @@
+import React, { useCallback, useEffect, useState } from 'react'
+import { useAragonApi } from '../../../api-react'
+import { Button, GU, Link, Text } from '@aragon/ui'
+import { LoadingAnimation } from '../../Shared'
+import {
+  REQUESTED_GITHUB_TOKEN_SUCCESS,
+  REQUESTED_GITHUB_TOKEN_FAILURE,
+} from '../../../store/eventTypes'
+import { getToken, githubPopup, STATUS } from '../../../utils/github'
+
+const Loading = () => (
+  <div css={`text-align: center; margin-top: ${15 * GU}px`}>
+    <LoadingAnimation />
+  </div>
+)
+
+const SCOPE = 'public_repo'
+let popupRef = null
+
+export default function AuthorizeGitHub() {
+  const { api } = useAragonApi()
+  const [ githubLoading, setGithubLoading ] = useState(false)
+
+  useEffect(() => {
+    window.addEventListener('message', handlePopupMessage)
+    return () => {
+      window.removeEventListener('message', handlePopupMessage)
+    }
+  })
+
+  const handlePopupMessage = useCallback(async message => {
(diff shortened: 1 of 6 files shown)


chadoh:
@john-light what do you think of this approach?

apps/projects/app/components/Panel/NewIssue/AuthorizeGitHub.js
```diff
@@ -0,0 +1,81 @@
...
+  return (
+    <>
+      <br />
+      <Text.Block size="xlarge">We’re going to need more permissions</Text.Block>
+      <br />
+      <Text.Block>
+        The Projects app can make GitHub issues on your behalf, but to do that we need you to grant us more permissions.
+        <br />
+        <br />
+        We don’t think apps should ask for more permissions than they need. If you’re ready to start using the Projects app to make issues, we’ll request the minimal set of permissions that we can (GitHub call this the “<a href="https://developer.github.com/apps/building-oauth-apps/understanding-scopes-for-oauth-apps/#available-scopes"><code>public_repo</code> scope</a>”).
```
PeterMPhillips: > If you’re ready to start using the Projects app to make issues, we’ll request the minimal set of permissions that we can

Something bothers me about this sentence. I think it's because of the phrase 'that we can' at the end. Maybe it would be better like "If you’re ready to start using the Projects app to make issues, we’ll request the minimal set of permissions that we need" or "If you’re ready to start using the Projects app to make issues, we’ll request the minimal set of permissions necessary to do so"? I'm not sure, maybe @stellarmagnet or @jayalaves  can weigh in.

Also:

> GitHub call this

GitHub calls this

chadoh: What I learned (this might be [wrong](https://www.quora.com/When-referring-to-a-company-name-as-the-subject-of-a-sentence-which-is-grammatically-correct-do-or-does)):

* "GitHub call this", treating companies/brands/bands as a collective, is the European version.
* "GitHub calls this", treating companies/brands/bands as an individual is the North American version.

Anyhow, it has been reworded!

PR #1543, opened 2019-11-06
Address Book: added disclaimer for identity provider

### LINKED ISSUE #1538: Add disclaimer to Address Book
Add a disclaimer that is at the top of the Address Book app. This is temporary until our dependency PR is merged.

See Figma: https://www.figma.com/file/GEilkBpZ6cl10pPSnLMeDuZb/Address-Book?node-id=359%3A1481
* Hyperlink should open new window here: https://github.com/aragon/aragon.js/pull/391
* Display this disclaimer only on the Empty page

e18r:
Fixes #1538

### DIFF
--- a/apps/address-book/app/components/Card/Empty.js
+++ b/apps/address-book/app/components/Card/Empty.js
@@ -1,13 +1,32 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import styled from 'styled-components'
-import { Button, EmptyStateCard, GU, LoadingRing, unselectable } from '@aragon/ui'
+import { Button, EmptyStateCard, GU, Info, Link, LoadingRing } from '@aragon/ui'
 import emptyStatePng from '../../assets/no-contacts.png'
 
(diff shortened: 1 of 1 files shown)


stellarmagnet:
There is also an unnecessary line break after “requires”

apps/address-book/app/components/Card/Empty.js
```diff
@@ -41,11 +56,11 @@ Empty.propTypes = {
 }
 
 const EmptyWrapper = styled.div`
-  ${unselectable};
```
chadoh: Note that I also got rid of `unselectable`. I have tried to select this sort of text multiple times while troubleshooting, and it's always so annoying that I can't. I don't see any benefit to making these cards unselectable.

This change is not related, other than that I was modifying & alphabetizing all the other styles here.

chadoh:
> There is also an unnecessary line break after “requires”

@stellarmagnet this is a limitation of aragonUI. I worked around it, now. Good catch!

PR #1547, opened 2019-11-06
Changes to test-ganache to use aragon devchain

PeterMPhillips:


templates/open-enterprise/temp/scripts/test-ganache.sh
```diff
@@ -33,7 +33,7 @@ setup_testing_variables() {
 
 start_ganache() {
   echo "Starting ganache-cli..."
```
Quazia: Can we get this string changed since it's starting an aragon devchain now?

templates/open-enterprise/temp/scripts/test-ganache.sh
```diff
@@ -33,7 +33,7 @@ setup_testing_variables() {
...
-  npx ganache-cli -i ${NETWORK_ID} -l ${GAS_LIMIT} -a ${ACCOUNTS} -e ${BALANCE} -p ${PORT} > /dev/null &
+  npx aragon devchain --reset -a ${ACCOUNTS} -p ${PORT} > /dev/null &
   pid=$!
   sleep 3
   echo "Running ganache-cli with pid ${pid} in port ${PORT}, gas limit set to: ${GAS_LIMIT}"
```
Quazia: Same here, especially since we arn't setting a gas-limit anymore.

templates/open-enterprise/package.json
```diff
@@ -24,7 +24,7 @@
     "docker:run": "cd ../../node_modules/@aragon/templates-shared/; docker-compose -f docker-compose.yml up -d; RESULT=$?; cd -; $(exit $RESULT)",
     "docker:stop": "cd ../../node_modules/@aragon/templates-shared/; docker-compose down; cd -",
     "docker:wait-gas": "truffle exec ../../node_modules/@aragon/templates-shared/scripts/sleep-until-gaslimit.js --network devnet 6900000",
-    "deploy:rpc": "npm run compile && truffle exec ./scripts/deploy.js --network rpc",
+    "deploy:rpc": "BOUNTIES=$(npm run deploy:bounties | tail -n 1) && npm run compile && truffle exec ./scripts/deploy.js --network rpc --ens 0x5f6f7e8cc7346a11ca2def8f827b7a0b612c56a1 --dao-factory 0x5d94e3e7aec542ab0f9129b9a7badeb5b3ca0f77 --mini-me-factory 0xd526b7aba39cccf76422835e7fd5327b98ad73c9 --r true --s $BOUNTIES",
```
Quazia: I like the approach to the bounties contract here there's probably a few other places we could use this code snippet.

Quazia: Oh one more thing, can we expand the aliases?

templates/open-enterprise/temp/scripts/test-ganache.sh
ottodevs: I am sorry, but I don't understand what is the purpose of this change.

I would like to know if there are clear benefits for this, I initially see some downsides, for example, forcing to reset the local blockchain every time a test is run.

We probably shouldn't rely into `aragon devchain` for tests, one possible reason is the pre-deployed snapshot that is uses, it defeats truffle clean room features for contracts. In the other hand we are forced to reset the local devchain, I think we should not enforce the developer to do this, because there exist different workflows that might be altered. In my case, sometimes I find it convenient to reuse template or other contracts deployments during development, so I would not like to be forced to reset the data when there are better solutions.

I asked @PeterMPhillips the other day to fix template test that was not working because a problem with gas during standard bounties deployment, I am not sure if this was  the kind of solution I expected, I would like to know what specific problem is this PR fixing.

I fixed the template test issue in 987b57b7e4468d46b1efa9ab0d69c8e370bb6f2d

ottodevs: Maybe using `aragon devchain` eventually saves us from running the deployments scripts before tests, but in this case this PR would be incomplete until that solution is tested, and those deployment scripts removed. But I don't really think we should merge this as it is, because we are not using any `aragon devchain` pre-deployed asset in out tests, so this PR as it is does not make lot of sense, changing the command alone does not improve or solve anything, unless I am missing something, in that case, please correct me.

Issue #1526, opened 2019-11-03
Unable to cast a vote to create a new reward


john-light:
**Describe the bug**
I have created a proposal to create a new reward but am unable to cast a vote for this proposal, despite the voting app showing clearly that I have the ability to cast a vote. Get an error "Action impossible".

**Mainnet or testnet?**
testnet

**Organization**
openorgtest.aragonid.eth

**To Reproduce**
1. Create a reward -> create a proposal to create a reward
2. Go to voting app and try to vote on the proposal
3. Notice that you cannot vote to create the reward

**Expected behavior**
I expect to be able to cast my vote to approve or reject creating the reward.

**Screenshots**

![Screenshot_2019-11-02 Aragon(2)](https://user-images.githubusercontent.com/9424721/68078690-c05efc80-fdb1-11e9-8fb3-a6d6817276d2.png)

![rewarderror](https://user-images.githubusercontent.com/9424721/68078696-e4224280-fdb1-11e9-8e3a-1aea313db86f.png)

stellarmagnet:
Does ETH exist in your vault?

Link to your DAO?

stellarmagnet:
Can you share more details with what your selections were in the New Reward panel and what is displayed in the confirmation screen?

john-light:
Reference asset: Other 
Token address: ORG token contract address
Type: Recurring dividend
Amount: 0.01 ETH
Duration: From today til Feb 4, 2020, 1x per week

What is shown on the confirmation screen:

![Screenshot_2019-11-02 Aragon(4)](https://user-images.githubusercontent.com/9424721/68079031-d91ee080-fdb8-11e9-8719-1d61dfced4df.png)

![Screenshot_2019-11-02 Aragon(5)](https://user-images.githubusercontent.com/9424721/68079036-dcb26780-fdb8-11e9-9a56-de237bb250f9.png)

stellarmagnet:
Good find @john-light!

This seems to be related to how we pass in the date for recurring dividends, with respect to when it should actually begin.

For recurring dividends, the first one will occur at the end of first interval, e.g. in your case -- it should have actually attempted to start them on November 10th instead of November 3rd. 

We will make an update to the UI (and radspec) so this is all more clear that the first disbursement is *at the end* if the first interval, which is start date + interval length.

Dividends should still work properly for one-time.

[...]

PR #1548, opened 2019-11-06
Rewards: add block numbers to New Reward summary

e18r:
**TO DO before merging**
- [x] ~~wait for #1500 to get merged~~
- [x] ~~redirect this PR to `dev`~~

Related to #1526

Screenshot from 2019-11-06 14-37-55

### DIFF
--- a/apps/rewards/app/components/App/App.js
+++ b/apps/rewards/app/components/App/App.js
@@ -6,24 +6,6 @@ import { MyRewards, Overview } from '../Content'
 import PanelManager, { PANELS } from '../Panel'
 import styled from 'styled-components'
 import { Empty } from '../Card'
-import {
-  MILLISECONDS_IN_A_DAY,
-  MILLISECONDS_IN_A_WEEK,
-  MILLISECONDS_IN_A_MONTH,
-  MILLISECONDS_IN_A_YEAR,
-  millisecondsToBlocks,
-  millisecondsToDays,
-  millisecondsToWeeks,
-  millisecondsToMonths,
-  millisecondsToYears,
-} from '../../../../../shared/ui/utils'
-
-import { BigNumber } from 'bignumber.js'
-import {
-  ONE_TIME_DIVIDEND,
-  ONE_TIME_MERIT,
-  RECURRING_DIVIDEND,
-} from '../../utils/constants'
 import { networkContextType } from '../../../../../shared/ui'
 import { useAragonApi } from '../../api-react'
 import { IdentityProvider } from '../../../../../shared/identity'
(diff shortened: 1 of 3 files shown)


apps/rewards/app/components/App/App.js
```diff
@@ -156,54 +138,15 @@ class App extends React.Component {
...
       reward.isMerit, //reward.isMerit, //bool _isMerit,
       reward.referenceAsset.key || reward.customToken.address, //address _referenceToken,
       reward.amountToken.address, //address _rewardToken,
-      reward.amount.toString(10), //uint _amount,
-      startBlock, // uint _startBlock
+      reward.amountWei.toString(10), //uint _amount,
+      reward.startBlock, // uint _startBlock
       reward.duration, //uint _duration, (number of blocks until reward will be available)
-      reward.occurances, //uint _occurances,
+      reward.occurrences, //uint _occurrences,
```
chadoh: thanks! 😃

apps/rewards/app/components/Panel/NewReward/NewReward.js
```diff
@@ -641,6 +651,86 @@ class NewRewardClass extends React.Component {
...
+      startBlock,
+      duration,
+      occurrences,
+    ] = this.getBlockProps(currentBlockNumber)
+    const disbursementBlocks = []
+    for (let i = 1; i <= occurrences; i ++) {
+      const block = startBlock + duration * i
+      disbursementBlocks.push(block)
+    }
+    console.log(disbursementBlocks)
```
chadoh: please remove

apps/rewards/app/components/Panel/NewReward/NewReward.js
```diff
@@ -641,6 +651,86 @@ class NewRewardClass extends React.Component {
...
+      startBlock = dateReference <= new Date() ? currentBlock + rawBlockDuration - BLOCK_PADDING : currentBlock 
+      duration = dateReference <= new Date() ? BLOCK_PADDING : rawBlockDuration
+    }
+    this.setState({
+      isMerit,
+      amountWei,
+      startBlock,
+      duration,
+      occurrences,
+    })
```
chadoh: I find it unexpected that `getBlockProps` has a side effect (setting state). Could we set this state in `setDisbursementBlocks`, with the other `setState`?

chadoh: Also, if we remove the call to `this.setState`, it becomes pretty easy to move this function outside of the component.

apps/rewards/app/components/Panel/NewReward/NewReward.js
```diff
@@ -653,7 +743,10 @@ class NewRewardClass extends React.Component {
       dateStart,
       dateEnd,
       disbursements,
+      disbursementBlocks,
     } = this.state
+    this.props.app.web3Eth('getBlockNumber')
+      .subscribe(this.setDisbursementBlocks)
```
chadoh: `showSummary` is run during `render`. `render` should not have side effects, such as setting state.

Let's move this calculation to `componentDidMount` (if we were using a functional component, we could `useEffect`). Then we can fetch the value and set the state once, on initial load of the NewReward component, rather than on every render pass.

topocount: It also doesn't looks like this is getting retriggered on each modification to the reward. For example if a user is bouncing back between the configuration and confirmation panes, the block numbers, and number of blocks needed to be calculated should be updating. It doesn't look like this is happening: 
image

~5700 blocks between a weekly reward is not correct, and an additional block was not calculated for the additional disbursement I generated upon revision

e18r:
@chadoh @topocount thank you for your review comments. I'm addressing them now.

apps/rewards/app/components/Panel/NewReward/NewReward.js
```diff
@@ -89,6 +89,59 @@ const INITIAL_STATE = {
...
+      duration = millisecondsToBlocks(Date.now(), disbursement * MILLISECONDS_IN_A_MONTH + Date.now())
+    }
+    startBlock -= duration
+  }
+  if(rewardType === ONE_TIME_DIVIDEND){
+    const rawBlockDuration = millisecondsToBlocks(Date.now(), dateReference)
+    startBlock = dateReference <= new Date() ? currentBlock + rawBlockDuration - BLOCK_PADDING : currentBlock 
+    duration = dateReference <= new Date() ? BLOCK_PADDING : rawBlockDuration
+  }
+  return [ isMerit, amountWei, startBlock, duration, occurrences ]
```
chadoh: With this many things returned, I'd prefer we used an object instead of an array. Names are easier for humans to work with than position within an array.

apps/rewards/app/components/Panel/NewReward/NewReward.js
```diff
@@ -89,6 +89,59 @@ const INITIAL_STATE = {
...
+    amount,
+    amountToken,
+    rewardType,
+    dateReference,
+    dateStart,
+    dateEnd,
+    disbursement,
+    disbursementUnit,
+    disbursements,
+  } = state
```
chadoh: Rather than passing these all in as a single argument, can we pass them in as named arguments, aka a destructured object?

```js
const getBlockProps = ({
  currentBlock,
  amount,
  amountToken,
  rewardType,
  dateReference,
  dateStart,
  dateEnd,
  disbursement,
  disbursementUnit,
  disbursements,
}) => {
  ...
}
```

apps/rewards/app/components/Panel/NewReward/NewReward.js
e18r: These two issues are indeed related and by solving one I solved the other. Instead of putting `setDisbursementBlocks` in `render`, I put it in `onSubmitDraft`, where it gets executed right when appropriate. Thanks, guys, for these very appropriate and illustrative remarks!

e18r:
@chadoh @topocount Ready for a new review!

Issue #1507, opened 2019-11-01
Any time I access the Settings tab, I see a broken avatar for a second


stellarmagnet:
If logged in with Github, any time I access the Settings tab, my avatar is shown up as broken

![image](https://user-images.githubusercontent.com/2584493/68008333-ce8b0c80-fc3b-11e9-9da6-a160fc7a0383.png)

PeterMPhillips:
![Peek 2019-11-06 15-21](https://user-images.githubusercontent.com/19808076/68346566-4e383180-00a9-11ea-91eb-37057e46a153.gif)

Is it better if the text isn't displayed until the image is available? Or is this change in size too jarring upon loading?

ottodevs:
This was reviewed and merged in #1551 
It seems we are good closing it! (Feel free to re-open if needed, but comment the reason then!)
I tested locally and couldn't reproduce.

PR #1551, opened 2019-11-07
Projects: Fix brief display of broken avatar on Settings page

PeterMPhillips:
Issue #1507 
- When you navigate to the Settings page you briefly see a broken image for the avatar
image

- This fixes the problem by waiting to get the avatar url before displaying the page contents, while waiting it displays the loading message
Peek 2019-11-06 21-48

### DIFF
--- a/apps/projects/app/components/Content/Settings.js
+++ b/apps/projects/app/components/Content/Settings.js
@@ -33,28 +33,21 @@ const fundingModels = [
   'Hourly',
 ]
 
-const GitHubConnect = ({ onLogin, onLogout, status }) => {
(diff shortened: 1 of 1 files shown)


topocount:
I think it might be better to just render the settings panel without the avatar component, and then re-render with the avatar component once it loads. thoughts, @chadoh ?

PeterMPhillips:
That was my first instinct as well but there was a bit of visual jumpiness. @stellarmagnet preferred it this way.

Here is what it looked like they way @topocount wants it:
Peek 2019-11-06 15-32

PR #1564, opened 2019-11-07
Projects/FundIssues: check all vault funds

### LINKED ISSUE #1552: "No funds found" displayed, even though funds exist
1. Create a new DAO
2. Navigate to Organization and mint some test tokens
3. Navigate to Projects app. Connect a repo. Select an issue and "fund issue"

Will get a UI that says No funds found, even though I have funds in my finance app.

## Projects app
image

## Finance app
image

chadoh:
Fixes #1552

* When funding mode is "Fixed": check all vault funds before displaying "No Funds" message
* When funding mode is "Hourly": check only configured funding token

### DIFF
--- a/apps/projects/app/components/Panel/FundIssues/FundIssues.js
+++ b/apps/projects/app/components/Panel/FundIssues/FundIssues.js
@@ -1,11 +1,10 @@
 /* eslint-disable react/prop-types */
 // issues are validated using correct shape - eslint problem?
-import React, { useEffect, useState } from 'react'
+import React, { useEffect, useMemo, useState } from 'react'
 import PropTypes from 'prop-types'
 import styled from 'styled-components'
 import { addHours } from 'date-fns'
 import BigNumber from 'bignumber.js'
-import Icon from '../../Shared/assets/components/IconEmptyVault'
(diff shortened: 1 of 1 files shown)


apps/projects/app/components/Panel/FundIssues/FundIssues.js
```diff
@@ -615,23 +615,16 @@ const FundIssues = ({ issues, mode }) => {
...
-            Your base rate has not been set and you do not have
-            any tokens in your Vault.
-            <br /> <br />
-            Once you have tokens in your Vault, you will be
-            able to begin funding issues.
-            <br /> <br />
-            <Button wide onClick={closePanel} mode="strong">Cancel</Button>
-          </div>
-        </Text>
-      </div>
```
chadoh: This was an alternative "no funds available" message that never actually showed up, because the logic doesn't do what it looks like it does. `tokens.length` is only `0` when `appState` hasn't yet loaded, and `tokenDetails.balance` will never be `0`, it will be `"0"`. So in fact, this showed up briefly before our `useEffect` hooks get called. As soon as those get called, this empty message would never show up. The time between this first & second render was so small that I doubt anything ever got painted to the DOM.

PR #1635, opened 2019-11-12
Rewards/My Rewards: add icons to "status"

### LINKED ISSUE #1627: Rewards/My Rewards: add icons to "status"
Current:
Screenshot from 2019-11-12 10-37-43

Expected:
Screenshot from 2019-11-12 10-37-27

e18r:
Fixes #1627

Screenshot from 2019-11-12 14-39-49

**Note**: I couldn't find the double check icon in AragonUI. I used
the circled check instead.

### DIFF
--- a/apps/rewards/app/components/Content/MyRewards.js
+++ b/apps/rewards/app/components/Content/MyRewards.js
@@ -5,6 +5,9 @@ import {
   ContextMenu,
   ContextMenuItem,
   DataView,
+  IconCheck,
+  IconCircleCheck,
+  IconClock,
   IconCoin,
   IconView,
   Text,
@@ -20,6 +23,39 @@ import Metrics from './Metrics'
 import { useAppState } from '@aragon/api-react'
 import BigNumber from 'bignumber.js'
 import { displayCurrency } from '../../utils/helpers'
+
+const PendingStatus = (theme) => (
+  <div style={{ display: 'flex' }}>
+    <IconClock style={{
+      marginRight: '4px',
+      marginTop: '-2px',
+      color: theme.warning,
+    }}/>
+    Pending
+  </div>
+)
+
+const ReadyStatus = (theme) => (
+  <div style={{ display: 'flex' }}>
+    <IconCheck style={{
+      marginRight: '4px',
+      marginTop: '-2px',
+      color: theme.positive,
+    }}/>
+    Ready to claim
+  </div>
+)
+
+const ClaimedStatus = (theme) => (
+  <div style={{ display: 'flex' }}>
+    <IconCircleCheck style={{
+      marginRight: '4px',
+      marginTop: '-2px',
+      color: theme.positive,
+    }}/>
+    Claimed
+  </div>
+)
 
 const MyRewards = ({
   myRewards,
@@ -107,7 +143,9 @@ const renderOneTimeDividend = (reward, amountTokens) => {
     </Text>
   )
   const disbursementDate = dateReference.toDateString()
-  const status = timeClaimed > 0 ? 'Claimed' : (Date.now() > endDate ? 'Ready to claim' : 'Pending')
+  const status = timeClaimed > 0 ? ClaimedStatus(theme) : (
+    Date.now() > endDate ? ReadyStatus(theme) : PendingStatus(theme)
+  )
   return [ description, disbursementDate, status, displayAmount ]
 }
 
@@ -117,8 +155,6 @@ const renderRecurringDividend = (reward, amountTokens) => {
     description,
     userRewardAmount,
     amountToken,
-    endDate,
-    timeClaimed,
     claims,
     disbursements
   } = reward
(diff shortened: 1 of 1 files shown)


apps/rewards/app/components/Content/MyRewards.js
```diff
@@ -21,6 +24,39 @@ import { useAppState } from '@aragon/api-react'
 import BigNumber from 'bignumber.js'
 import { displayCurrency } from '../../utils/helpers'
 
+const PendingStatus = (theme) => (
+  <div style={{ display: 'flex' }}>
+    <IconClock style={{
```
chadoh: Can we use `css=` in all of these spots instead of `style=`? That will use `styled-components` behind the scenes, which injects `style` tags into the `head` of the document, rather than using inline styles. This results in better browser performance and is easier to work with.

apps/rewards/app/components/Content/MyRewards.js
```diff
@@ -21,6 +24,39 @@ import { useAppState } from '@aragon/api-react'
...
+    }}/>
+    Ready to claim
+  </div>
+)
+
+const ClaimedStatus = (theme) => (
+  <div style={{ display: 'flex' }}>
+    <IconCircleCheck style={{
+      marginRight: '4px',
+      marginTop: '-2px',
```
chadoh: Do you think it's worth pulling these common styles into a separate component?

```jsx
const Status = ({ color, Icon, text }) => (
  <div css="display: flex; align-items: center">
    <Icon css={`margin-right: ${0.5 * GU}px; color: ${color}`} />
    {text}
  </div>
)
```

Now that I wrote this out, I want to note two things and ask another.

1. Having a component that receives _an argument_ (`theme`) rather than _props_ (`{ theme }`) seems unexpected and error-prone to me
2. We can re-define `theme` within this component. `const theme = useTheme()`. Even though it means we need to switch from an implicit to an explicit return, I think the simplification of the calling code makes it worth it.
2. Does the `align-items: center` mean we can get rid of the `margin-top`?

Moving on, the above can be used as:

```jsx
<Status color={theme.positive} Icon={IconCircleCheck} text="Claimed" />
```

Now that I've written this out, I'm not sure if it's even worth having intermediate components for each kind of status, although it does keep the consuming code concise. What do you think?

chadoh: Thinking through this and how we'll pass `color` in but keep `theme` within the `Status` component, maybe something like this makes sense:

```jsx
const types = {
  claimed: { icon: IconCircleCheck, text: 'Claimed', color: 'positive' },
  ready: { icon: IconCheck, text: 'Ready to claim', color: 'positive' },
  pending: { icon: IconClock, text: 'Pending', color: 'warning' },
}

const Status = ({ type }) => {
  const theme = useTheme()
  const status = types[type]
  const text = status.text
  const Icon = status.icon
  const color = theme[status.color]
  return (
    <div css="display: flex; align-items: center">
      <Icon css={`margin-right: ${0.5 * GU}px; color: ${color}`} />
      {text}
    </div>
  )
}

Status.propTypes = {
  type: PropTypes.oneOf(Object.keys(types)).isRequired,
}

Comment (e18r) [reply apps/rewards/app/components/Content/MyRewards.js]: @chadoh, thank you so much for your review. As always, very instructive. Regarding your specific points:

1. Thank you for explaining this to me. It wasn't clear why that component wasn't really working as a component, but rather as a function.
2. Is it the same in terms of performance? Aren't we creating multiple `theme` instances? Are hooks supposed to be used like that?
3. Nice try, but no. The icon needs to be a little up in order to match the text:
Screenshot from 2019-11-15 14-29-47
Otherwise, if it's in the center, the text will be higher:
Screenshot from 2019-11-15 14-30-23
But my takeaway from this is that you make this sort of reviews by just writing the text on GitHub without any kind of testing, which I find quite impressive.

File: apps/rewards/app/components/Content/MyRewards.js (hunk 1 above)
Comment (e18r) [reply apps/rewards/app/components/Content/MyRewards.js]: Whenever I write the `css` property in the custom `<Icon>` element, it fails with
Screenshot from 2019-11-15 14-31-37
So in that case I had to leave the `style` prop. Unless you know of a way to solve it, that is.

File: apps/rewards/app/components/Content/MyRewards.js (hunk 2 above)
Comment (e18r) [reply apps/rewards/app/components/Content/MyRewards.js]: I think your solution for having a single `Status` component with a single `type` prop is brilliant.

File: apps/rewards/app/components/Content/MyRewards.js [hunk 3]
```diff
@@ -21,6 +25,36 @@ import { useAppState } from '@aragon/api-react'
...
+  const status = types[type]
+  const text = status.text
+  const Icon = status.icon
+  const color = theme[status.color]
+  return (
+    <div css='display: flex;'>
+      <Icon style={{
+        marginRight: 0.5 * GU + 'px',
+        marginTop: -0.25 * GU + 'px',
+        color: color,
```
chadoh: You could shorten this to just `color,`, since that expands to `color: color,`

PR #1640, opened 2019-11-14
Rewards/ViewReward: add status info

### LINKED ISSUE #1629: Rewards/ViewReward: add status information
When clicked from My Rewards, add status information to the ViewReward side panel. Also change the title from "View Reward" to "View my reward".

Current:
Screenshot from 2019-11-12 10-50-47

Expected:
Screenshot from 2019-11-12 10-51-11

e18r:
**TO DO before merging**
- [x] Wait for PR #1635 to get merged
- [x] Redirect this PR to `dev`

Fixes #1629

Screenshot from 2019-11-14 09-04-48

### DIFF
--- a/apps/rewards/app/components/App/App.js
+++ b/apps/rewards/app/components/App/App.js
@@ -183,10 +183,14 @@ class App extends React.Component {
     })
   }
 
-  viewReward = (reward, panelTitle) => {
+  viewReward = (reward, isMyReward) => {
+    const panelTitle = isMyReward ?  'View my reward' : 'View Reward'
     this.setState({
       panel: PANELS.ViewReward,
-      panelProps: reward,
+      panelProps: {
+        reward,
+        isMyReward,
+      },
       panelTitle,
     })
   }
@@ -265,14 +269,14 @@ class App extends React.Component {
           <MyRewards
             myRewards={this.props.myRewards}
             myMetrics={this.props.myMetrics}
-            viewReward={reward => this.viewReward(reward, 'View my reward')}
+            viewReward={reward => this.viewReward(reward, true)}
             claimReward={this.claimReward}
           />
         ) : (
           <Overview
             rewards={this.props.rewards === undefined ? [] : this.props.rewards}
             newReward={this.newReward}
-            viewReward={reward => this.viewReward(reward, 'View Reward')}
+            viewReward={reward => this.viewReward(reward, false)}
             metrics={this.props.metrics}
           />
         )}
--- a/apps/rewards/app/components/App/App.js
+++ b/apps/rewards/app/components/App/App.js
@@ -118,7 +118,11 @@ class App extends React.Component {
   }
 
   closePanel = () => {
-    this.setState({ panel: undefined, panelProps: undefined })
+    this.setState({
+      panel: undefined,
+      panelProps: undefined,
+      panelTitle: undefined,
+    })
   }
 
   selectTab = idx => {
@@ -179,10 +183,11 @@ class App extends React.Component {
     })
   }
 
-  viewReward = reward => {
+  viewReward = (reward, panelTitle) => {
     this.setState({
       panel: PANELS.ViewReward,
       panelProps: reward,
+      panelTitle,
     })
   }
 
@@ -218,6 +223,7 @@ class App extends React.Component {
           <PanelManager
             onClose={this.closePanel}
             activePanel={this.state.panel}
+            title={this.state.panelTitle}
             {...this.state.panelProps}
           />
         </IdentityProvider>
@@ -259,14 +265,14 @@ class App extends React.Component {
           <MyRewards
             myRewards={this.props.myRewards}
             myMetrics={this.props.myMetrics}
-            viewReward={this.viewReward}
+            viewReward={reward => this.viewReward(reward, 'View my reward')}
             claimReward={this.claimReward}
           />
         ) : (
           <Overview
             rewards={this.props.rewards === undefined ? [] : this.props.rewards}
             newReward={this.newReward}
-            viewReward={this.viewReward}
+            viewReward={reward => this.viewReward(reward, 'View Reward')}
             metrics={this.props.metrics}
           />
         )}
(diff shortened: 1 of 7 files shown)


apps/rewards/app/components/App/App.js
```diff
@@ -259,14 +269,14 @@ class App extends React.Component {
...
-            viewReward={this.viewReward}
+            viewReward={reward => this.viewReward(reward, true)}
             claimReward={this.claimReward}
           />
         ) : (
           <Overview
             rewards={this.props.rewards === undefined ? [] : this.props.rewards}
             newReward={this.newReward}
-            viewReward={this.viewReward}
+            viewReward={reward => this.viewReward(reward, false)}
```
chadoh: When looking at how a function is called, I like to have some idea of what all the arguments mean, without needing to reference the actual function definition. Passing booleans makes that really hard. What is true or false?

Fortunately, I think this is a case where using object destructuring is pretty straightforward:

```js
viewReward({ reward, isMyReward: false })

File: apps/rewards/app/components/Panel/PanelManager.js [hunk 2]
```diff
@@ -18,12 +18,14 @@ const PANELS = Object.keys(dynamicImport).reduce((obj, item) => {
...
 
-const PanelManager = ({ activePanel = null, onClose, ...panelProps }) => {
-  let panelTitle = activePanel && camel2title(activePanel)
+const PanelManager = ({
+  activePanel = null,
+  onClose,
+  title,
+  ...panelProps
+}) => {
+  let panelTitle = activePanel && (title || camel2title(activePanel))
```
chadoh: This doesn't seem quite right. If `title` was provided, we don't really care if `activePanel` is falsey, right? So maybe more like `title || (activePanel && camel2title(activePanel))`

apps/rewards/app/components/Panel/PanelManager.js
```diff
@@ -18,12 +18,14 @@ const PANELS = Object.keys(dynamicImport).reduce((obj, item) => {
...
   const PanelComponent = activePanel && React.lazy(dynamicImport[activePanel])
-  if (panelProps.reward) {
-    panelTitle += ' #' + panelProps.reward.rewardId
```
chadoh: Since we got rid of this, I think we can use `const` instead of `let` when defining `panelTitle` above.

apps/rewards/app/components/Styles/Header.js
```diff
@@ -0,0 +1,12 @@
+import styled from 'styled-components'
+import { Text } from '@aragon/ui'
+
+const Header = styled(Text.Block).attrs({
+  smallcaps: true,
+})`
+  color: ${({ theme }) => theme.contentSecondary};
```
chadoh: Sort of a separate discussion, but I wonder if it makes more sense to use inline `css=` props for components like this, which would let us easily stop passing in `theme` and allowing the component to define it itself:

```js
import React from 'react'
import { GU, Text, theme } from '@aragon/ui'

export default function Header({ children }) {
  const theme = useTheme()
  return (
    <Text.Block
      smallcaps
      css={`color: ${theme.contentSecondary}; margin-top: ${2 * GU}px; margin-bottom: ${GU}px`}
    >
      {children}
    </Text.Block>
  )
}
```

A trade-off is that stylelist doesn't currently inspect these inline `css=` props correctly.

What do you think?

(Also, rewriting this made me think that we should probably use GU here.)

apps/rewards/app/utils/helpers.js
```diff
@@ -1,9 +1,71 @@
...
+  disbursements
+}) => {
+  if (rewardType === RECURRING_DIVIDEND)
+    return claims === disbursements.length ? <Status type="claimed" /> : (
+      Date.now() > disbursements[claims].getTime() ? <Status type="ready" /> :
+        <Status type="pending" />
+    )
+  else return timeClaimed > 0 ? <Status type="claimed" /> : (
+    Date.now() > endDate ? <Status type="ready" /> : <Status type="pending" />
+  )
```
chadoh: I don't know how I feel about the lack of curly brackets with multiline `if/else` statements. Also, the mix of `if/else` and ternaries seems strange. What do you think of redoing this along these lines?

```js
  if (rewardType === RECURRING_DIVIDEND) {
    if (claims === disbursements.length) return <Status type="claimed" />
    if (Date.now() > disbursements[claims].getTime()) return <Status type="ready" />
    return <Status type="pending" />
  } else {
    if (timeClaimed > 0) return <Status type="claimed" />
    if (Date.now() > endDate) <Status type="ready" />
    return <Status type="pending" />
  }
```

apps/rewards/app/components/Content/MyRewards.js
```diff
@@ -163,33 +120,22 @@ const renderRecurringDividend = (reward, amountTokens) => {
     </Text>
   )
   const disbursementDate = (disbursements[claims] || disbursements[claims-1]).toDateString()
-  const status = claims === disbursements.length ? <Status type="claimed" /> : (
-    Date.now() > disbursements[claims].getTime() ? <Status type="ready" /> :
-      <Status type="pending" />
-  )
-  return [ description, disbursementDate, status, displayAmount ]
+  const Status = getStatus(reward)
+  return [ description, disbursementDate, Status, displayAmount ]
```
chadoh: Did `status` need to be capitalized in these places?

e18r: Not really

apps/rewards/app/components/Panel/PanelManager.js
e18r: Yes, but no need for parentheses because `&&` takes precedence over `||`

apps/rewards/app/components/Styles/Header.js
e18r: I like your suggestion of making this a fully-fledged component. However, I had to use `style` instead of `css`.

chadoh: I find it very strange that wrapping it in `styled()` works, but using inline `css=` doesn't. I thought they did the same thing!!

e18r: :woman_shrugging:

PR #1653, opened 2019-11-15
Rewards/MyRewards: add etherscan link

### LINKED ISSUE #1631: Rewards/MyRewards: add etherscan link
For claimed rewards, add context menu item with etherscan link for the tx hash of the payout.

Current:
Screenshot from 2019-11-12 11-04-55

Expected:
Screenshot from 2019-11-12 11-03-07

e18r:
**TO DO before merging**
- [ ] wait for PR #1640 to be merged
- [ ] redirect this PR to `dev`

Fixes #1631

Screenshot from 2019-11-15 07-18-34

### DIFF
--- a/apps/rewards/app/components/App/App.js
+++ b/apps/rewards/app/components/App/App.js
@@ -40,6 +40,7 @@ class App extends React.Component {
     this.state = {
       selected: 0,
       tabs: [ 'Overview', 'My Rewards' ],
+      claimHashes: {},
     }
     this.updateRewards()
   }
@@ -196,8 +197,12 @@ class App extends React.Component {
   }
 
   claimReward = reward => {
-    // TODO
-    this.props.api.claimReward(reward.rewardId + reward.claims).toPromise()
+    this.props.api.claimReward(reward.rewardId + reward.claims)
+      .subscribe(claimHash => {
+        const { claimHashes } = this.state
+        claimHashes[reward.rewardId] = claimHash
+        this.setState({ claimHashes })
+      })
   }
 
   openDetailsView = reward => {
@@ -271,6 +276,7 @@ class App extends React.Component {
             myMetrics={this.props.myMetrics}
             viewReward={reward => this.viewReward(reward, true)}
             claimReward={this.claimReward}
+            claimHashes={this.state.claimHashes}
           />
         ) : (
           <Overview
--- a/apps/rewards/app/components/Content/MyRewards.js
+++ b/apps/rewards/app/components/Content/MyRewards.js
@@ -6,6 +6,7 @@ import {
   ContextMenuItem,
   DataView,
   IconCoin,
+  IconHash,
   IconView,
   Text,
   useTheme,
@@ -14,48 +15,73 @@ import {
   ONE_TIME_DIVIDEND,
   RECURRING_DIVIDEND,
   ONE_TIME_MERIT,
+  READY,
+  CLAIMED,
 } from '../../utils/constants'
 import { Empty } from '../Card'
 import Metrics from './Metrics'
-import { useAppState } from '@aragon/api-react'
+import { useAppState, useNetwork } from '@aragon/api-react'
 import BigNumber from 'bignumber.js'
-import { displayCurrency, getStatus } from '../../utils/helpers'
+import { displayCurrency, getStatus, getStatusId } from '../../utils/helpers'
+
 
 const MyRewards = ({
   myRewards,
   myMetrics,
   viewReward,
   claimReward,
+  claimHashes,
 }) => {
+  const network = useNetwork()
   const rewardsEmpty = myRewards.length === 0
+
+  const getEtherscanLink = reward => {
+    const networkName = network.id === 1 ? '' : `${network.type}.`
+    const claimHash = claimHashes[reward.rewardId]
+    const link = `https://${networkName}etherscan.io/tx/${claimHash}`
+    return link
+  }
 
   if (rewardsEmpty) {
     return <Empty noButton />
   }
 
-  const renderMenu = (reward) => (
-    <ContextMenu>
-      <StyledContextMenuItem
-        onClick={() => viewReward(reward)}
-      >
-        <IconView css={{
-          marginRight: '11px',
-          marginBottom: '2px',
-        }}/>
-        View
-      </StyledContextMenuItem>
-      {(reward.claims < reward.disbursements.length && (reward.disbursements[reward.claims] < Date.now())) && (
+  const renderMenu = (reward) => {
+    const statusId = getStatusId(reward)
+    return (
+      <ContextMenu>
         <StyledContextMenuItem
-          onClick={() => claimReward(reward)}
+          onClick={() => viewReward(reward)}
         >
-          <IconCoin css={{
+          <IconView css={{
             marginRight: '11px',
             marginBottom: '2px',
           }}/>
-        Claim
-        </StyledContextMenuItem>)}
-    </ContextMenu>
-  )
+          View reward
+        </StyledContextMenuItem>
+        {statusId === READY && (
+          <StyledContextMenuItem
+            onClick={() => claimReward(reward)}
+          >
+            <IconCoin css={{
+              marginRight: '11px',
+              marginBottom: '2px',
+            }}/>
+            Claim
+          </StyledContextMenuItem>)}
+        {statusId === CLAIMED && (
+          <StyledContextMenuItem
+            href={getEtherscanLink(reward)}
+          >
+            <IconHash css={{
+              marginRight: '11px',
+              marginBottom: '2px',
+            }}/>
+            View on Etherscan
+          </StyledContextMenuItem>)}
+      </ContextMenu>
+    )
+  }
 
   return (
     <OverviewMain>
--- a/apps/rewards/app/utils/constants.js
+++ b/apps/rewards/app/utils/constants.js
@@ -26,3 +26,6 @@ export const DISBURSEMENT_UNITS = [
   YEARS,
 ]
 export const OTHER = 'Other…'
+export const PENDING = Symbol('PENDING')
+export const READY = Symbol('READY')
+export const CLAIMED = Symbol('CLAIMED')
--- a/apps/rewards/app/utils/helpers.js
+++ b/apps/rewards/app/utils/helpers.js
@@ -2,6 +2,9 @@ import {
   ETH_DECIMALS,
   ETH_DECIMALS_NUMBER,
   RECURRING_DIVIDEND,
+  PENDING,
+  READY,
+  CLAIMED,
 } from './constants'
 import BigNumber from 'bignumber.js'
 import { IconCheck, IconCircleCheck, IconClock } from '@aragon/ui'
@@ -48,16 +51,30 @@ const ClaimedStatus = (theme) => (
   </div>
 )
 
-export const getStatus = (
-  { rewardType, timeClaimed, endDate, claims, disbursements },
-  theme
-) => {
+export const getStatusId = ({
+  rewardType,
+  timeClaimed,
+  endDate,
+  claims,
+  disbursements,
+}) => {
   if (rewardType === RECURRING_DIVIDEND)
-    return claims === disbursements.length ? ClaimedStatus(theme) : (
-      Date.now() > disbursements[claims].getTime() ? ReadyStatus(theme) :
-        PendingStatus(theme)
+    return claims === disbursements.length ? CLAIMED : (
+      Date.now() > disbursements[claims].getTime() ? READY : PENDING
     )
-  else return timeClaimed > 0 ? ClaimedStatus(theme) : (
-    Date.now() > endDate ? ReadyStatus(theme) : PendingStatus(theme)
+  else return timeClaimed > 0 ? CLAIMED : (
+    Date.now() > endDate ? READY : PENDING
   )
+}
+
+export const getStatus = (reward, theme) => {
+  const statusId = getStatusId(reward)
+  switch(statusId) {
+  case PENDING:
+    return PendingStatus(theme)
+  case READY:
+    return ReadyStatus(theme)
+  case CLAIMED:
+    return ClaimedStatus(theme)
+  }
 }


apps/rewards/app/components/App/App.js
```diff
@@ -196,8 +197,12 @@ class App extends React.Component {
   }
 
   claimReward = reward => {
-    // TODO
-    this.props.api.claimReward(reward.rewardId + reward.claims).toPromise()
+    this.props.api.claimReward(reward.rewardId + reward.claims)
+      .subscribe(claimHash => {
+        const { claimHashes } = this.state
```
e18r: Is it worth declaring a `const` here? I feel like it adds clarity to what I'm doing.

apps/rewards/app/components/Content/MyRewards.js
```diff
@@ -14,48 +16,59 @@ import {
...
   myMetrics,
   viewReward,
   claimReward,
+  claimHashes,
 }) => {
+  const network = useNetwork()
   const rewardsEmpty = myRewards.length === 0
 
+  const getEtherscanLink = reward => {
+    const networkChunk = network.id === 1 ? '' : `${network.type}.`
```
e18r: I could have used `network.type`, but I'm not sure about it's return value for mainnet. If it's `mainnet`, it's not useful because `mainnet.etherscan.io` doesn't work. The documentation is rather laconic about it, and I couldn't find code examples. In any case, checking the `id` seems more robust. Obviously, this hasn't been tested.

apps/rewards/app/components/Content/MyRewards.js
```diff
@@ -14,48 +16,59 @@ import {
...
+    const claimHash = claimHashes[reward.rewardId]
+    const link = `https://${networkChunk}etherscan.io/tx/${claimHash}`
+    return link
```
e18r: I like returning variables instead of complex expressions. It's easier to debug.

apps/rewards/app/components/Content/MyRewards.js
```diff
@@ -143,6 +156,13 @@ MyRewards.propTypes = {
   myMetrics: PropTypes.arrayOf(PropTypes.object).isRequired,
 }
 
+const StyledIcon = ({ Icon }) => (
+  <Icon style={{
+    marginRight: GU * 1.5 + 'px',
+    marginBottom: GU / 4 + 'px',
```
e18r: Originally, this was 11 and 2 pixels respectively. I made this change half-jokingly because it looks pretty awful and I'm not sure it makes sense to include `GU` in this case.

apps/rewards/app/utils/constants.js
```diff
@@ -26,3 +26,6 @@ export const DISBURSEMENT_UNITS = [
   YEARS,
 ]
 export const OTHER = 'Other…'
+export const PENDING = Symbol('PENDING')
+export const READY = Symbol('READY')
+export const CLAIMED = Symbol('CLAIMED')
```
e18r: Although we haven't used `Symbol` before, I think it's the way to go whenever possible.

apps/rewards/app/components/Content/MyRewards.js
chadoh: I'm not sure either! I guess a way to think about it is: if `GU` changed to, say, `2`, would we want this to change accordingly? If `GU` changed to `10` or even `20`, would we want that? If so, let's use GU. If we want this to always be 2px, then let's skip GU here.

apps/rewards/app/components/Content/MyRewards.js
chadoh: Can you test this and check? Does it use the user-configured setting from MetaMask?

apps/rewards/app/components/App/App.js
chadoh: I guess an alternative would be:

```js
this.setState({ claimHashes: {
  ...this.state.claimHashes,
  [reward.rewardId]: claimHash,
}})
```

I'm not sure how I feel about my formatting; hopefully eslint has opinions that we can defer to.

I don't think my suggestion or the way you wrote this read more clearly. I have questions either way!

* Do you think `claimedHashes` is a more accurate name? These are past-tense _claimed_, right? 
* `claimHashes` is initialized to an empty object, and it's only updated when a reward is claimed. If the user refreshes the page, this link will go away. Am I understanding correctly? Is that what we want?

apps/rewards/app/components/Content/MyRewards.js
```diff
@@ -14,48 +16,59 @@ import {
...
   myRewards,
   myMetrics,
   viewReward,
   claimReward,
+  claimHashes,
 }) => {
+  const network = useNetwork()
   const rewardsEmpty = myRewards.length === 0
 
+  const getEtherscanLink = reward => {
```
chadoh: Does this use any component state? If not, let's move it outside the definition of `MyRewards`.

apps/rewards/app/components/Content/MyRewards.js
```diff
@@ -14,48 +16,59 @@ import {
...
-        >
-          <IconCoin css={{
-            marginRight: '11px',
-            marginBottom: '2px',
-          }}/>
-        Claim
-        </StyledContextMenuItem>)}
-    </ContextMenu>
-  )
+  const renderMenu = (reward) => {
```
chadoh: We could also move this outside of the component, and make it a component. But that's probably out-of-scope for now.

apps/rewards/app/components/App/App.js
e18r: > `claimHashes` is initialized to an empty object, and it's only updated when a reward is claimed. If the user refreshes the page, this link will go away.
:scream:

e18r: > Do you think `claimedHashes` is a more accurate name? These are past-tense claimed, right?

You're the English native speaker, but the hashes themselves aren't the thing being claimed. Instead, the hashes refer to claims. So the `claim` in `claimHashes` is a noun rather than a verb. Still, if it doesn't sound right to you, let's consider other options: `claimsHashes` seems more accurate to me because there are many claims, but it sounds bad. Any other ideas?

e18r: > I don't think my suggestion or the way you wrote this read more clearly. I have questions either way!

To me, my way reads more clearly, although maybe I'm not being objective.

apps/rewards/app/components/Content/MyRewards.js
e18r: What I meant by this not having been tested is that it can only be tested on mainnet. Or am I missing something? What setting from Metamask? If you're wondering about the network id, there's no doubt that mainnet is 1.

apps/rewards/app/components/Content/MyRewards.js
e18r: OK, nice rule of thumb. I'm gonna skip it, then.

e18r:
After what was discussed in today's meeting, I decided to stop my current work and ask around if I'm in the right direction. I'm updating this PR's description to reflect the current state.

stellarmagnet:
I don't think we should modify the contract at this stage for this feature. Additionally, we need to figure out how important it is to keep churning on this. Nobody user requested it (iirc), so my thoughts are we close this PR and keep it as a future backlog item for Rewards.

Any time a contract change is being considered, buy-in should be received from the original contract authors and myself before commencing development work.

e18r:
@stellarmagnet There's no need to modify the contract for this, as @ottodevs kindly explained to me yesterday. Thank you for the clarification.

chadoh:
> There's a way to retrieve the transaction hash from the blockchain each time without the need to storing anything. This would be the preferred approach, I think, if it was possible. But is it?

Yes, I believe there is! I think a quick pairing session might be the best way to get you started with this; check your calendar invites 😄

e18r:
Hey @chadoh, I think I got it. Thank you so much though!

apps/rewards/app/components/Content/MyRewards.js
e18r: It does use `claimHashes` which is a prop. And also `network`, which is a hook. So not possible.

apps/rewards/app/components/Content/MyRewards.js
```diff
@@ -14,48 +16,62 @@ import {
   ONE_TIME_DIVIDEND,
   RECURRING_DIVIDEND,
   ONE_TIME_MERIT,
+  READY,
+  CLAIMED,
 } from '../../utils/constants'
 import { Empty } from '../Card'
 import Metrics from './Metrics'
-import { useAppState } from '@aragon/api-react'
+import { useAppState, useAragonApi, useNetwork } from '@aragon/api-react'
```
e18r: Can I remove the `useAppState` hook and import `appState` from `useAragonApi`? I've seen it elsewhere but I'm not 100% sure it's the same thing.

apps/rewards/app/components/Content/MyRewards.js
e18r: So `renderMenu` is used as `DataView`'s `renderEntryActions` property. According to AragonUI's [docs](https://github.com/aragon/aragon-ui/tree/master/src/components/DataView#renderentryactionsentry-index--selected-mode-), `renderEntryActions` gets a function, not a component.
Screenshot from 2019-11-27 12-19-46

Does it still make sense for it to become a component?

apps/rewards/app/components/Content/MyRewards.js
chadoh: I'd prefer we use `useCallback`, in that case.

```js
const getEtherescanLink = useCallback(reward => {
  // existing function here
}, [ claimHashes, network ])
```

Benefits:

1. The function will only be redefined if `claimHashes` or `network` change, rather than being redefined on every render of this function
2. Code clarity/communication: this helps anyone else looking at this code quickly understand what bits of component state this function relies on (`claimHashes` and `network`)

apps/rewards/app/components/Content/MyRewards.js
chadoh: I guess I'm wondering if `useNetwork()` responds to the user setting in MetaMask. If it does, then you can set MM to mainnet and then check `network.type` to see what it looks like. 

How does `useNetwork()` check what network its on, other than using MetaMask? This might be a question for @topocount or @Quazia.

apps/rewards/app/components/Content/MyRewards.js
chadoh: I still think it makes sense to move it outside. I guess we can't make it a component right now, though it might be worth opening an issue in the aragonUI repo.

apps/rewards/app/components/Content/MyRewards.js
chadoh: [Yes](https://hack.aragon.org/docs/api-js-ref-react#useappstate) 😄

They are the same, and I think I prefer always using `useAragonApi`

apps/rewards/app/components/Content/MyRewards.js
```diff
@@ -14,48 +15,62 @@ import {
...
   myMetrics,
   viewReward,
   claimReward,
+  claimHashes,
 }) => {
+  const network = useNetwork()
   const rewardsEmpty = myRewards.length === 0
+  const { api } = useAragonApi()
+  const [ user, setUser ] = useState()
+  api.accounts().subscribe(accounts => setUser(accounts[0]))
```
chadoh: Is this the same as `const { connectedAccount } = useAragonApi()`? See [docs here](https://hack.aragon.org/docs/api-js-ref-react#connectedaccount) – I think it might be.

Why include the user's address in the `claimHashes` structure? Why not add `transactionHash` to the `reward` directly, so we can get rid of this `user` logic, and simplify `getEtherescanLink` with something like:

```js
const link = `https://${networkChunk}etherscan.io/tx/${reward.transactionHash}`
```

apps/rewards/app/store/events.js
```diff
@@ -4,7 +4,12 @@ import { addressesEqual } from '../utils/web3-utils'
 import { INITIALIZATION_TRIGGER } from './'
 
 export const handleEvent = async (state, event, settings) => {
-  const { event: eventName, returnValues, address: eventAddress, } = event
+  const {
+    event: eventName,
+    returnValues,
+    address: eventAddress,
+    transactionHash,
+  } = event
```
chadoh: Can we access the network here, rather than in the frontend? We don't want the link to change, depending on which network the user is currently connected to. We want to link them to the URL where they can view the actual transaction, whichever network that happened to be on!

apps/rewards/app/components/App/App.js
chadoh: I'm leaning toward "get rid of `claimHashes` altogether," as mentioned [here](https://github.com/AutarkLabs/open-enterprise/pull/1653#discussion_r354511112) and [here](https://github.com/AutarkLabs/open-enterprise/pull/1653#discussion_r354512239). We don't have to stick with the current structure of `rewards` that we have now; we can add `transactionHash` or, better yet, `etherscanUrl` to each reward, and simplify all the logic we have here.

apps/rewards/app/components/Content/MyRewards.js
e18r: Because of the way rewards work. You create a reward, and then any token holder can claim it. Since multiple users should be able to claim a reward at any time, we need to track who claimed it. Your approach would only store the last claim, so every user would see the last claim made by any user, instead of their own claim.

apps/rewards/app/store/events.js
e18r: If you're on one network you won't see any rewards made on another network in the first place, so I think it's safe to access the network from the frontend.

apps/rewards/app/components/Content/MyRewards.js
ottodevs: That's just an interface that gets an updated value for the Aragon's api `network` observable: 
https://github.com/aragon/aragon.js/blob/e4a155ef82420147d4926756861a5bd6cd46f84f/packages/aragon-api/src/index.js#L79

It gets the value from the current provider, it can be MM or other tool injecting the provider.

e18r:
After discussing this approach with @chadoh, we realized that there's no need to create a data structure to hold the transaction IDs of every claimed disbursement for every reward, since we're doing this on a specific user's front end and we only care about this user's last claim for each reward. Instead, at the store level, we can create a Etherscan link by looking at claim events where the claimant corresponds to the current user. Since there's only one etherscan link per reward, it can be directly stored as a field in each reward object in the `rewards` data structure.

If it's not possible to obtain the current user at the store level, it can be obtain via a front end trigger. I'm going to update this PR's description with this information as well.

PR #1657, opened 2019-11-18
[WIP] Reputation template

PeterMPhillips:
These changes allow for multiple token options when creating a new DAO

First token options:
- Shareholder
- Membership
- Reputation

Second token option:
- Reputation

Also,
- Can set either the first or second token to be in control of Voting and/or Dot Voting
- The Reputation tokens use WhitelistOracle to control who can transfer tokens

### DIFF
--- a/shared/integrations/token-manager-custom/contracts/TokenManager.sol
+++ b/shared/integrations/token-manager-custom/contracts/TokenManager.sol
@@ -14,7 +14,7 @@ import "@aragon/os/contracts/lib/math/SafeMath.sol";
 import "@aragon/apps-shared-minime/contracts/ITokenController.sol";
 import "@aragon/apps-shared-minime/contracts/MiniMeToken.sol";
 
-import { ITransferOracle } from "@autark/apps-whitelist-oracle/contracts/WhitelistOracle.sol";
+import { ITransferOracle } from "../../whitelist-oracle/contracts/WhitelistOracle.sol";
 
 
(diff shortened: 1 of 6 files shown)


stellarmagnet:
In the way this is currently written, does it require two tokens, or is two tokens optional?

PeterMPhillips:
@stellarmagnet the second token is optional

stellarmagnet:
@PeterMPhillips okay, so it seems we aren't allowing membership + shareholder? I know that isn't exactly "reputation", but we probably want to make this more flexible so we don't have to deal with any more enhancements in the future. Can we update this to a draft PR as the product requirements get finalized, as it seems we have to probably add the agent as well.

PeterMPhillips:
@stellarmagnet Yes, when I discussed this with @Quazia he said to just make the second token reputation, but I think we could give all the options to the second token as well. However, I know @ottodevs is working on template stuff as well and I believe his plan was to selectively merge work from this branch into the branch he was working on, so I'm hesitant to add any more to this branch. On that point, @ottodevs do you even need this PR to remain open?

stellarmagnet:
@PeterMPhillips I wasn't sure what was going on with the two dev streams on the reputation work. Maybe lets keep both PRs as is for now and circle back once the requirements are more clearly documented. cc @ottodevs FYI

ottodevs:
Rebased base branch and merged this PR just for symbolization, the changes are already integrated into the target and properly attributed co-authoring from @PeterMPhillips

PR #1660, opened 2019-11-18
Dot Voting: Fix badge color

PeterMPhillips:
Screenshot from 2019-11-18 13-37-24

### DIFF
--- a/apps/dot-voting/app/components/VotingCard.js
+++ b/apps/dot-voting/app/components/VotingCard.js
@@ -103,7 +103,6 @@ const VotingCard = ({ vote, onSelectVote }) => {
                 cursor: pointer;
                 padding: 2px 8px;
                 pointer-events: auto;
-                mix-blend-mode: normal;
               `}
             >
               {' + ' + (options.length - 2) + ' more'}
--- a/apps/dot-voting/app/components/VotingCard.js
+++ b/apps/dot-voting/app/components/VotingCard.js
@@ -18,6 +18,7 @@ function noop() {}
 
 const VotingCard = ({ vote, onSelectVote }) => {
   const theme = useTheme()
+  console.log(theme)
   const { voteWeights } = useUserVoteStats(vote)
   const { description, endDate, open, totalSupport, voteId, support } = vote
   const { executionTargetData, options, totalVoters } = vote.data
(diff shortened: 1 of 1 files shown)


apps/dot-voting/app/components/VotingCard.js
```diff
@@ -96,13 +96,14 @@ const VotingCard = ({ vote, onSelectVote }) => {
         {options.length > 2 && (
           <div css="text-align: center; width: 100%; margin-top: 10px">
             <Badge
-              shape="compact"
-              foreground={`${theme.surfaceOpened}`}
-              background={`${theme.surfaceUnder}`}
+              foreground={`${theme.surfaceContentSecondary}bf`}
+              background={`${theme.border}7f`}
```
chadoh: What do these `bf` and `7f` do? I haven't seen this pattern before.

PeterMPhillips: Yeah, this is ugly and I was really unhappy about doing it. Its just adding alpha values to the colours in hex. In the figma, these values use the `opacity` value in the css but that doesn't work in the component as the opacity on the background is multiplied by the opacity on the foreground making the foreground more transparent than it should be. However, now that you point it out, I could take the rgb values from the theme color and then have the alpha as a percentage rather than the hex. e.g. `foreground = 'rgba(${theme.surfaceContentSecondary.r}, ${theme.surfaceContentSecondary.g}, ${theme.surfaceContentSecondary.b}, 0.7);'` But I feel like that may be a worse alternative.

stellarmagnet: We don't have to exactly copy what is done in Figma if it breaks the conventions of the design system. @jayalaves can you please provide recommendations to what the hex values should be so they are in line with the design system?

jayalaves: @PeterMPhillips The closest value as far as hex goes without applying opacity for the shadow is #D9D9D9. How does Aragon One style their vote cards on the Voting app?

stellarmagnet: Please hold @PeterMPhillips there is a miscommunication with the problem here :)

jayalaves: @PeterMPhillips Simply use 'surfaceUnder' for the background, and 'contentSecondary' for the text. Those should be nearly if not fully accurate with what we have on Figma.

PeterMPhillips: Thank you @jayalaves!
Screenshot from 2019-11-26 10-19-02

apps/dot-voting/app/components/VotingCard.js
```diff
@@ -96,13 +96,14 @@ const VotingCard = ({ vote, onSelectVote }) => {
...
-              foreground={`${theme.surfaceOpened}`}
-              background={`${theme.surfaceUnder}`}
+              foreground={theme.surfaceContentSecondary}
+              background={theme.surfaceUnder}
+              uppercase={false}
               css={`
                 cursor: pointer;
                 padding: 2px 8px;
                 pointer-events: auto;
+                mix-blend-mode: normal;
```
chadoh: Do we still need this, or is it vestigial?

PeterMPhillips: Good catch! Removed

PR #1661, opened 2019-11-19
Projects: New layout on small screens

PeterMPhillips:
Large:
Screenshot from 2019-11-18 16-17-04

Medium:
Screenshot from 2019-11-18 16-17-23

Small:
Screenshot from 2019-11-18 16-17-47

Extra Small:
Screenshot from 2019-11-18 16-18-45

### DIFF
--- a/apps/projects/app/components/Card/Issue.js
+++ b/apps/projects/app/components/Card/Issue.js
@@ -2,7 +2,7 @@ import PropTypes from 'prop-types'
 import React from 'react'
 import styled from 'styled-components'
 
-import { Text, Tag, Checkbox, ContextMenu, useLayout, useTheme, IconClock, IconConnect, IconCalendar } from '@aragon/ui'
+import { Text, Tag, Checkbox, ContextMenu, GU, useLayout, useTheme, IconClock, IconConnect, IconCalendar } from '@aragon/ui'
 
 import { formatDistance } from 'date-fns'
 import { BountyContextMenu } from '../Shared'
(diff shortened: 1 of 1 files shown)


chadoh:
@jayalaves looking at the screenshots here has me wondering why we both saying difficulty level or due date when a bounty is fulfilled. Can we get rid of those?

@PeterMPhillips if we decide to get rid of those, would you want to do that as part of this PR, or in a follow-up?

chadoh:
So when going from small to medium, we swap the order of labels and the repo/date info. Why? @jayalaves can we keep these consistent? It will result in cleaner code.

stellarmagnet:
> @jayalaves looking at the screenshots here has me wondering why we both saying difficulty level or due date when a bounty is fulfilled. Can we get rid of those?

@chadoh I don't think we should remove difficulty level. For deadline, we can keep the design as is, and say "Done 7 days ago", where the time would be when the bounty was awarded. I just checked and this is Gitcoin's approach.

stellarmagnet:
> So when going from small to medium, we swap the order of labels and the repo/date info. Why? @jayalaves can we keep these consistent? It will result in cleaner code.

@chadoh this was something I previously had the same concerns with, but was on the fence about it. I'm leaning toward keeping it consistent as you are proposing.

jayalaves:
> So when going from small to medium, we swap the order of labels and the repo/date info. Why? @jayalaves can we keep these consistent? It will result in cleaner code.

@chadoh This is most definitely a design inconsistency on my end that was not meant to be there. Please keep it consistent and let me know once deployed so that I can align the result with the designs on Figma.

PeterMPhillips:
@stellarmagnet @jayalaves @chadoh Could I get a description of what actually needs changing? Because if we are no longer rearranging the elements based on screen size, I think we can just close this PR.

stellarmagnet:
I think closing this makes sense. @jayalaves if you agree, please close it. I have assigned this ticket to you.

jayalaves:
@stellarmagnet @chadoh @PeterMPhillips Yes, we can close this and continue the conversation on the [Enhanced bounty experience](https://app.zenhub.com/workspaces/product-management-5ddbeeb7052f2e0001d9c8ac/issues/autarklabs/product/4) epic when needed. These new specs are already ready for Yalda's review and I've asked Chad for a first look on the issues tab in particular in a follow up comment on that thread to make sure we're all on the same page.

PR #1668, opened 2019-11-19
Projects: Removed difficulty modifier from Fixed funding model

PeterMPhillips:
Removed difficulty modifiers from the fixed funding model and added remove buttons to both fixed and hourly.

Screenshot from 2019-11-19 09-44-07
Screenshot from 2019-11-19 09-44-20
Screenshot from 2019-11-19 09-50-53

### DIFF
--- a/apps/projects/app/components/Content/Settings.js
+++ b/apps/projects/app/components/Content/Settings.js
@@ -145,7 +145,7 @@ const ExperienceLevel = ({
                 <TextInput
                   type="text"
                   css="font-size: 16px; width: 65%"
-                  defaultValue={exp.name}
+                  value={exp.name}
                   onChange={generateExpLevelHandler(index, 'N')}
                 />
                 <IconContainer
@@ -166,7 +166,7 @@ const ExperienceLevel = ({
               <TextInput
                 type="text"
                 css="font-size: 16px; flex-grow: 1"
-                defaultValue={exp.name}
+                value={exp.name}
                 onChange={generateExpLevelHandler(index, 'N')}
               />
               <IconContainer
--- a/apps/projects/app/components/Content/Settings.js
+++ b/apps/projects/app/components/Content/Settings.js
@@ -9,11 +9,13 @@ import {
   Field,
   GU,
   Info,
+  IconRemove,
   Link,
   Text,
   TextInput,
   useLayout,
   useTheme,
+  unselectable,
 } from '@aragon/ui'
 
 import { LocalIdentityBadge } from '../../../../../shared/identity'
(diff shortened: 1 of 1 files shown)


stellarmagnet:
When it has no multiplier, the label should say just “difficulty”

PeterMPhillips:
New label:
Screenshot from 2019-11-19 12-08-40

apps/projects/app/components/Content/Settings.js
```diff
@@ -502,4 +548,25 @@ const Column = styled.div`
...
+  all: unset;
+  color: ${({ disabled, theme }) => (disabled ? theme.disabled : theme.contentSecondary)};
+  cursor: ${({ disabled }) => (disabled ? 'not-allowed' : 'pointer')};
+  display: flex;
+  justify-content: center;
+  :hover {
+    color: ${({ disabled, theme }) => disabled ? theme.disabled : theme.contentBorderActive};
+  }
+  :active {
+    color: ${({ disabled, theme }) => (disabled ? theme.disabled : theme.accent)};
```
chadoh: Let's wrap the whole set of hover & active styles in a conditional. We don't need them if `disabled`. So:

```js
${p => !p.disabled && css`
  :hover { ... }
  :active { ... }
`}

File: apps/projects/app/components/Content/Settings.js [hunk 2]
```diff
@@ -114,37 +116,71 @@ SettingLabel.propTypes = {
...
+                css="font-size: 16px; flex-grow: 1"
+                defaultValue={exp.name}
+                onChange={generateExpLevelHandler(index, 'N')}
+              />
+              <IconContainer
+                theme={theme}
+                disabled={expLevels.length <= 1}
+                style={{ transform: 'scale(.8)' }}
+                onClick={() => onRemoveExpLevel(index)}
+                title="Remove this difficulty multiplier"
```
chadoh: This should say "Remove this difficulty" or maybe "Remove this difficulty setting," but should not say "multiplier."

apps/projects/app/components/Content/Settings.js
```diff
@@ -502,4 +548,25 @@ const Column = styled.div`
...
+    color: ${({ disabled, theme }) => disabled ? theme.disabled : theme.contentBorderActive};
+  }
+  :active {
+    color: ${({ disabled, theme }) => (disabled ? theme.disabled : theme.accent)};
+  }
+  > svg {
+    color: inherit;
+    height: 40px;
+    width: 40px;
+    transition: all 0.6s cubic-bezier(0.165, 0.84, 0.44, 1);
```
chadoh: Fancy!

apps/projects/app/components/Content/Settings.js
PeterMPhillips: Does this work for you?
```
${({ disabled, theme }) => (disabled ? `
      color: ${theme.disabled};
      cursor: not-allowed;
    ` : `
      color: ${theme.contentSecondary};
      cursor: pointer;
      :hover {
        color: ${theme.surfaceOpened};
      }
      :active {
        color: ${theme.accent};
      }
    `)
}
```

PeterMPhillips:
Good catch @chadoh. It turns out the fields values were being set with `defaultValue` instead of `value` so they weren't getting updated on changes

javieralaves:
@PeterMPhillips I'm not able to reproduce these changes [here](https://rinkeby.aragon.org/#/neudune/0xe3fc72a7f3dda98a0a76863c9aad73968c6c0c42/) when switching between Fixed and Hourly, am I doing something wrong?

Issue #1584, opened 2019-11-08
Data retained after going from Edit Budget to New Budget


stellarmagnet:
1. Navigate to allocations app that has budgets
2. Click on a budget card and click "edit budget"
3. Close/cancel
4. Click on "New budget"

You'll see the data from the budget you edited there. 

Expected: when I hit new budget, the fields should be empty

javieralaves:
Tested this in the continuous deploy from December 17th, 2019.

✅ This issue no longer exists.

PR #1670, opened 2019-11-19
Allocations: New panel gets fresh state

PeterMPhillips:
Fix for issue #1584:

Before the panel state would update the INITIAL_STATE const on updates, so anytime the panel state was initialized it would just get the last state instead of a new state.

Also, it should be noted that the 'Edit budget' option has been removed from the context menu. I'm not sure why this was done so I left it that way. Whenever we decide to have an 'Edit budget' again this state problem will be fixed.

### DIFF
--- a/apps/allocations/app/components/Panel/NewAllocation.js
+++ b/apps/allocations/app/components/Panel/NewAllocation.js
@@ -54,7 +54,7 @@ class NewAllocation extends React.Component {
   constructor(props) {
     super(props)
     const { budgets, budgetId } = props
-    this.state = INITIAL_STATE
+    this.state = { ...INITIAL_STATE }
     this.state.recipients = {}
     this.state.recipientsValid = {}
     const recipientId = Date.now()
--- a/apps/allocations/app/components/Panel/NewBudget.js
+++ b/apps/allocations/app/components/Panel/NewBudget.js
@@ -39,7 +39,7 @@ class NewBudget extends React.Component {
 
   constructor(props) {
     super(props)
-    this.state =  INITIAL_STATE
+    this.state = { ...INITIAL_STATE }
     if (props.editingBudget.id) {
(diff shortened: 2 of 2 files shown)


PeterMPhillips:
@ottodevs 
First, to test this you have to remove some commented out line of code in components/Card/Budget.js.
Before:
Screenshot from 2019-11-21 09-57-38

After:
Screenshot from 2019-11-21 10-05-46

You can see the behaviour described in the issue here:
NewBudgetBefore

And after the fix:
NewBudgetAfter

Regarding the comment about the issue happening in edit budget, not new budget, there is only one script that handles new budget and edit budget -- NewBudget.js. But regardless, clearly the issue is due to the interaction of both.

Regarding the comment about new allocation, yes this issue doesn't affect the NewAllocation.js script and so it didn't need to be changed, I was attempting to keep things consistent across the scripts and because there is clearly an issue with assumptions that were made about the consistency of `INITIAL_STATE`. However, it does bring to attention that we aren't handling state particularly well in that script. Just a quick walkthrough:
1) At the constructor we set `this.state` with this line: `this.state = INITIAL_STATE`. However, since this is just a reference, any changes to `this.state` causes changes to `INITITAL_STATE`. 
2) So to reset `this.state`/`INITIAL_STATE` we also have to directly set each relevant variables e.g. `this.state.recipients = {}`
3) When a user types a recipient into the field, instead of calling `setState` we directly change the recipients value like so `recipients[e.target.id] = value` which causes `INITIAL_STATE` to be updated to that value, thus requiring us to reset that value in step 2.

Either we should not use `INITIAL_STATE` and just set the state directly inside the constructor or we should take steps to prevent `INITIAL_STATE` from getting updated on state changes so that it can be reliably used to reset the state. Right now we're using `INITIAL_STATE` and resetting values in the constructor and there's no reason why. However, such changes are outside of the scope of this issue.

PR #1672, opened 2019-11-19
Allocations: Fix edit budget symbol

PeterMPhillips:
Third symbol in the dropdown:
Screenshot from 2019-11-19 15-29-17

Displays correctly when editing budget:
Screenshot from 2019-11-19 15-29-44

### DIFF
--- a/apps/allocations/app/components/Panel/NewBudget.js
+++ b/apps/allocations/app/components/Panel/NewBudget.js
@@ -46,7 +46,6 @@ class NewBudget extends React.Component {
       this.state.amount = BigNumber(props.editingBudget.amount)
         .div(ETH_DECIMALS)
       this.state.amountError = false
-      this.state.selectedToken = 0
       this.state.buttonText = 'Submit'
     }
   }
--- a/apps/allocations/app/components/Panel/NewBudget.js
+++ b/apps/allocations/app/components/Panel/NewBudget.js
@@ -41,12 +41,13 @@ class NewBudget extends React.Component {
     super(props)
     this.state =  INITIAL_STATE
     if (props.editingBudget.id) {
+      const tokenIndex = props.tokens.findIndex(token => token.symbol === props.editingBudget.token.symbol)
       this.state.name = props.editingBudget.name
       this.state.nameError = false
       this.state.amount = BigNumber(props.editingBudget.amount)
         .div(ETH_DECIMALS)
       this.state.amountError = false
-      this.state.selectedToken = props.editingBudget.token.symbol === 'ETH' ? 0 : 1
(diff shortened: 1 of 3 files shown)


PeterMPhillips:
Updated radspec to use the ETH address but not include the symbol. This way we display the correct amount as long as the token is 18 decimals but do not display the symbol. In the future we can update this pass the token address to the function, but that will require an audit and a major update.

Screenshot from 2019-11-20 13-37-00

apps/allocations/app/components/Panel/NewBudget.js
```diff
@@ -41,12 +41,13 @@ class NewBudget extends React.Component {
...
     this.state =  INITIAL_STATE
     if (props.editingBudget.id) {
+      const tokenIndex = props.tokens.findIndex(token => token.symbol === props.editingBudget.token.symbol)
       this.state.name = props.editingBudget.name
       this.state.nameError = false
       this.state.amount = BigNumber(props.editingBudget.amount)
         .div(ETH_DECIMALS)
       this.state.amountError = false
-      this.state.selectedToken = props.editingBudget.token.symbol === 'ETH' ? 0 : 1
+      this.state.selectedToken = tokenIndex === -1 ? 0 : tokenIndex //Default to ETH if index not found?
```
chadoh: Why the question mark?

PeterMPhillips: Oops! I wrote that with the plan to follow up on what we wanted for the default behaviour. So if we cannot find the token symbol among the props.tokens, what behaviour do we want? Should we just default to ETH or should we have some error message?

chadoh: Seems like an errors state to me!

PeterMPhillips: I decided to take another route. Currently, for editing the budget we have a bit of a roundabout way of displaying the symbol, e.g. `symbols[selectedToken]`. But we already know the token symbol -- `props.editingBudget.token.symbol` so I decided to just display that value in the panel and set `this.state.selectedToken = 0` since we don't need to know the token to update the budget. This way we don't need to do any error handling.

apps/allocations/app/components/Panel/NewBudget.js
```diff
@@ -46,7 +46,7 @@ class NewBudget extends React.Component {
       this.state.amount = BigNumber(props.editingBudget.amount)
         .div(ETH_DECIMALS)
       this.state.amountError = false
-      this.state.selectedToken = props.editingBudget.token.symbol === 'ETH' ? 0 : 1
+      this.state.selectedToken = 0
```
chadoh: It looks like this is what it was initialize to on line 42 above. I think we can safely delete this line.

Issue #1591, opened 2019-11-08
Funding several issues including one with funding behaves poorly


Quazia:
**Describe the bug**

When selecting multiple issues, one with funding several things occur.
Below, only the issue with funding should be displayed in the warning message.
![paste (3)](https://user-images.githubusercontent.com/6549550/68444582-4493f400-01a5-11ea-86d9-9172c10864f3.png)
Additionally, the sentence "To update an individual..." should be removed.

Eventually, when submitting the transaction, it fails with the following error.
![paste (2)](https://user-images.githubusercontent.com/6549550/68444534-2201db00-01a5-11ea-9b7d-05f91e19ebe2.png)

**To Reproduce**

1. Fund one issue with a bounty
2. Select multiple issues, 2 without a bounty and 1 with a bounty.
3. Select "fund issues"
4. This will bring you to the panel above, showing it the issues

javieralaves:
Tested this in the continuous deploy from December 17th, 2019. Works as expected!

PR #1676, opened 2019-11-20
Projects: Switch from issues to bountylessIssues

PeterMPhillips:
Fixes issues in #1591

Accidentally branched off of `projects-contextmenu` and PR #1669

### DIFF
--- a/apps/projects/app/components/Panel/FundIssues/FundIssues.js
+++ b/apps/projects/app/components/Panel/FundIssues/FundIssues.js
@@ -267,7 +267,7 @@ const FundForm = ({
     const zeroErrArray = []
     const dateErrArray = []
     Object.values(bounties).forEach(bounty => {
-      if (!bounty.payout) zeroErrArray.push(bounty.issueId)
+      if (!bounty.payout || bounty.payout === '0') zeroErrArray.push(bounty.issueId)
       if (today > bounty.deadline) dateErrArray.push(bounty.issueId)
     })
     setZeroError(zeroErrArray)
@@ -278,7 +278,7 @@ const FundForm = ({
       !!zeroErrArray.length ||
       !!dateErrArray.length
     )
-  }, [ validate, bounties, description, maxErrors ])
+  }, [ bounties, description, maxErrors ])
 
   return (
     <>
@@ -318,7 +318,7 @@ const FundForm = ({
                   issue={issue}
                   bounty={bounties[issue.id]}
                   tokens={tokens}
-                  onBlur={() => setValidate(true)}
+                  onFocus={() => setValidate(true)}
                   updateBounty={updateBounty(issue.id)}
                 />
               ))}
@@ -364,7 +364,19 @@ const FundIssues = ({ issues, mode }) => {
     if (bountySettings.fundingModel === 'Fixed') return appState.tokens
     return [appState.tokens.find(t => t.addr === bountySettings.bountyCurrency)]
   }, [bountySettings])
-  const [ bounties, setBounties ] = useState(bountiesFor({ bountySettings, issues, tokens }))
+  const bountylessIssues = useMemo(
+    () => issues.filter(i => !i.hasBounty),
+    [issues]
+  )
+  const alreadyAdded = useMemo(
+    () => issues.filter(i => i.hasBounty),
+    [issues]
+  )
+  const [ bounties, setBounties ] = useState(bountiesFor({
+    bountySettings,
+    issues: bountylessIssues,
+    tokens
+  }))
 
   const fundsAvailable = useMemo(() => tokens.reduce(
     (sum, t) => sum.plus(BigNumber(t.balance)),
@@ -382,7 +394,7 @@ const FundIssues = ({ issues, mode }) => {
       }
     }
 
-    if (update.hours || update.exp) {
+    if (update.hours !== undefined || update.exp) {
       const { exp, hours } = newBounties[issueId]
       const { baseRate, expLvls } = bountySettings
       newBounties[issueId].payout = hours * baseRate * expLvls[exp].mul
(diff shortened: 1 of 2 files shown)


apps/projects/app/components/Panel/FundIssues/FundIssues.js
```diff
@@ -246,9 +246,9 @@ const FundForm = ({
     const today = new Date()
     const zeroErrArray = []
     const dateErrArray = []
-    Object.values(bounties).forEach(bounty => {
-      if (bounty.payout === 0) zeroErrArray.push(bounty.issueId)
-      if (today > bounty.deadline) dateErrArray.push(bounty.issueId)
+    issues.forEach(issue => {
+      if (Number(bounties[issue.id].payout) === 0) zeroErrArray.push(issue.id)
+      if (today > bounties[issue.id].deadline) dateErrArray.push(issue.id)
```
chadoh: This looks like it may have changed because of a merge conflict. I think we want the `bounties` version.

apps/projects/app/components/Panel/FundIssues/FundIssues.js
```diff
@@ -381,16 +388,16 @@ const FundIssues = ({ issues, mode }) => {
     // computes an array of issues and denests the actual issue object for smart contract
     const issuesArray = []
 
-    for (let key in issues) {
+    for (let key in bountylessIssues) {
```
chadoh: While we're here, let's change this to a `map`:

```js
const issuesArray = bountylessIssues.map(issue => {
  const {
    exp,
    fundingHistory,
    deadline,
    hours,
    payout,
    token,
  } = bounties[issue.id]
  return {
    ...issue,
    key: issue.id, // what is this? I think we can probably get rid of it
    exp,
    fundingHistory,
    deadline,
    hours,
    payout,
    token,
  }
})
```

Note that this new logic conflicts with my under-review work at #1659. If #1659 ends up getting merged before this one, I can help you resolve the conflicts. If this one gets merged first, I can handle it!

apps/projects/app/components/Panel/FundIssues/FundIssues.js
```diff
@@ -258,9 +258,9 @@ const FundForm = ({
     const today = new Date()
     const zeroErrArray = []
     const dateErrArray = []
-    Object.values(bounties).forEach(bounty => {
-      if (!bounty.payout) zeroErrArray.push(bounty.issueId)
-      if (today > bounty.deadline) dateErrArray.push(bounty.issueId)
+    issues.forEach(issue => {
+      if (!bounties[issue.id].payout || bounties[issue.id].payout === '0') zeroErrArray.push(issue.id)
+      if (today > bounties[issue.id].deadline) dateErrArray.push(issue.id)
```
chadoh: I see this change made it back in. Do we need it now?

apps/projects/app/components/Panel/FundIssues/FundIssues.js
```diff
@@ -352,6 +352,13 @@ const FundIssues = ({ issues, mode }) => {
     return [appState.tokens.find(t => t.addr === bountySettings.bountyCurrency)]
   }, [bountySettings])
   const [ bounties, setBounties ] = useState(bountiesFor({ bountySettings, issues, tokens }))
```
chadoh: We want this below the setup of `bountylessIssues`, right?

```js
const [ bounties, setBounties ] = useState(bountiesFor({ bountySettings, issues: bountylessIssues, tokens }))

**Comment by PeterMPhillips** [comment]
@chadoh did the fix where we are passing `bountylessIssues` to the `bountiesFor` function. Also, I made some changes to improve error validation (sometime error messages wouldn't trigger when they should, and sometimes the button would be disabled when it shouldn't)
e.g. the submit button wouldn't enable until after you clicked out of the field
before:
BrokenValidation

after:
FixedValidation

**Comment by chadoh** [comment]
> until after you clicked out of the field

This was intentional, so that it wouldn't say "amount must be greater than 0" when the panel first loads, as [discussed here](https://github.com/AutarkLabs/open-enterprise/issues/1565#issuecomment-555686708).

Also, now that #1659 has been merged, there are some big conflicts here. I can help resolve these!

File: apps/projects/app/components/Panel/FundIssues/FundIssues.js [hunk 5]
```diff
@@ -364,7 +364,19 @@ const FundIssues = ({ issues, mode }) => {
...
   }, [bountySettings])
-  const [ bounties, setBounties ] = useState(bountiesFor({ bountySettings, issues, tokens }))
+  const bountylessIssues = useMemo(
+    () => issues.filter(i => !i.hasBounty),
+    [issues]
+  )
+  const alreadyAdded = useMemo(
+    () => issues.filter(i => i.hasBounty),
+    [issues]
+  )
```
chadoh: @PeterMPhillips while resolving conflicts, I changed this logic. Rather than loop through `issues` once, on every render pass, it will now loop through `issues` twice on initial render, and then never again, unless the `issues` prop changes (which should only happen if you close the panel and re-open it).

apps/projects/app/components/Panel/FundIssues/FundIssues.js
```diff
@@ -318,7 +318,7 @@ const FundForm = ({
                   issue={issue}
                   bounty={bounties[issue.id]}
                   tokens={tokens}
-                  onBlur={() => setValidate(true)}
+                  onFocus={() => setValidate(true)}
```
chadoh: What problem did this fix? The behavior seems the same to me.

PeterMPhillips:
The onBlur to onFocus change does effect functionality. If you look here:
BrokenValidation

You can see that the `Fund Issue` button does not become enabled until after a user a clicked out of the `Amount` field.

chadoh:
Thank you for the clarification!
  
  
  

  
  

  
  
>

PR #1677, opened 2019-11-20
chore: create basic E2E tests for all apps

topocount:
This can be tested by:
1. `npm i` in this branch
2. `npm run cypress`

ensure you don't have anything else that's especially resource-hungry running because the fully-automated process of setting up an environment for testing is intense

Apps with completed basic flows:
- [x] Address Book
- [x] Allocations
- [x] Dot Voting
- [x] Rewards
- [x] Projects

Note: these tests should work with any dev dao deployed by run npm start:dev with @aragon/cli@6.3.3

Remaining TODOs:
- ~~acceptance-test @aragon/cli@6.4.x to see if the included client is updated with the event trigger and our heavily-used pathways are solid~~ not needed now that the client is side-loaded
- [x] update the `npm run cypress` command to cleanly set up a full-stack environment, run the tests, and then cleanly exit all services and scripts
- [ ] document the `npm run cypress` command as well as the development process for E2E tests


chadoh:
@topocount how do I actually run these tests? I tried running all the `cypress`-related commands from package.json. `cypress:open` got me the closest, but none of the tests passed; "could not connect to localhost:3000," they complained.

topocount:
> @topocount how do I actually run these tests? I tried running all the `cypress`-related commands from package.json. `cypress:open` got me the closest, but none of the tests passed; "could not connect to localhost:3000," they complained.

The process right now is still manual. I wasn't going to update the "single-command" test flow until I do more testing with the latest version of the CLI to ensure we're good to use it regularly. We can't have single command testing without this because we need the event trigger in the client that it utilizes.

Right now my process for getting this going:
- `ipfs daemon --offline` in any console
- `npm run start:dev` in our repo
- `npm run start:local` in the latest master of @aragon/aragon
- `npm run cypress:open` in our repo followed by clicking on the `aragon.spec.js` test. (soon to be renamed openEnterprise.spec.js)

chadoh:
I'm getting lots of errors when trying to run the tests:

Screen Shot 2019-11-22 at 14 34 42

---

Screen Shot 2019-11-22 at 14 35 53

---

Screen Shot 2019-11-22 at 14 36 29

---

Screen Shot 2019-11-22 at 14 36 42

cypress/integration/aragon.spec.js
```diff
@@ -11,144 +18,368 @@ context('Aragon', () => {
...
-    // https://on.cypress.io/nextall
-    cy.get('.traversal-next-all')
-      .contains('oranges')
-      .nextAll().should('have.length', 3)
-  })
+  context('Projects', () => {
+    let unfetch
+    before('logs into Projects', () => {
+      cy.contains('button','Projects',{ timeout: 150000 }).click().wait(10000)
+      cy.getWithinIframe('button:contains("Sign In")').click()
```
chadoh: We don't need to click this button to spoof the sign in flow. Since we can't interact with the popup anyhow, I don't think it's useful.

cypress/integration/aragon.spec.js
```diff
@@ -11,144 +18,368 @@ context('Aragon', () => {
...
-      .should('contain', 'Data')
-  })
 
-  it('.closest() - get closest ancestor DOM element', () => {
-    // https://on.cypress.io/closest
-    cy.get('.traversal-badge')
-      .closest('ul')
-      .should('have.class', 'list-group')
-  })
+  context('Allocations', () => {
```
chadoh: I would like to see many test files, rather than one giant one. Can we put test files near the code where they're used, the way people do with Jest tests? That would be my favorite way to organize them. Failing that, I'd at least like to see a directory per app. 

* `cypress/integration/projects/add-new-github-project.js`
* `cypress/integration/projects/fund-issue.js`
* `cypress/integration/projects/submit-application.js`

Do you think this pattern would work?

chadoh:
Have you figured out why it keeps POSTing even after the test runs have finished?

cypress/integration/aragon.spec.js
topocount: We click the button to create the `message` event listener in our app code. Once that listener is active. we post our own message to the app window that gets us logged in.

cypress/integration/aragon.spec.js
topocount: This pattern would work, but it would be slower, since each new file will require a fresh page load and syncing. I'd prefer to just keep one file so that automated testing can run as quickly as possible, and when we write tests for syncing, we can go this route.

topocount:
> Have you figured out why it keeps POSTing even after the test runs have finished?

This is just a convention of the interactive test environment. When I browse to the app outside of the test environment and log network activity, the level of noise we're seeing compared to the console and browser environment controlled by cypress isn't even close

topocount:
> I'm getting lots of errors when trying to run the tests:

I replicated the root of many of these errors is likely that you did not `npm i` upon switching into this branch

cypress/integration/aragon.spec.js
chadoh: Ah, this will change once we finally get #1300 merged!

cypress/integration/aragon.spec.js
chadoh: Hmm, ok. Maybe we could extract individual bits of this into functions, to keep this file manageable. 

Can we run separate tests in parallel? Test Rewards in one process while we test Projects in another, for example?

ottodevs: > This pattern would work, but it would be slower, since each new file will require a fresh page load and syncing. I'd prefer to just keep one file so that automated testing can run as quickly as possible, and when we write tests for syncing, we can go this route.


What is preventing us from using one test file in the root while importing tests from other files there? I think it would be the same result, in terms of quickness, with the added benefit of having a manageable suite, so I would take that route.

templates/dev/scripts/create-dao.js
```diff
@@ -1,10 +1,9 @@
 /* global artifacts, web3 */
-const { randomId } = require('@aragon/templates-shared/helpers/aragonId')
 const newDAO = require('./new-dao')
 
 
-const VOTE_DURATION = 60 // seconds
 // TODO: this is min quorum in dot-voting? it does not accept zero, but 1e-16
+const VOTE_DURATION = 60 // seconds
```
ottodevs: Please, note you are messing up the TODO note there by reordering this line

templates/dev/scripts/create-dao.js
```diff
@@ -15,13 +14,13 @@ const settings = {
   // The order is important
   dotVotingSettings: [ SUPPORT_REQUIRED, MIN_ACCEPTANCE_QUORUM, VOTE_DURATION ],
   financePeriod: 0,
-  id: randomId(),
+  id: '',
```
ottodevs, line 17: Is this required for the e2e testing? it seems to be messing our current development flow

package.json
```diff
@@ -7,7 +7,8 @@
     "husky": "^3.0.0",
     "lerna": "^3.10.8",
     "truffle-privatekey-provider": "1.1.0",
-    "wait-on": "^3.2.0"
+    "wait-on": "^3.2.0",
+    "web3": "^1.2.4"
```
ottodevs, line 13: Maybe we should consider moving the cypress suite to a sub-package to avoid polluting the mono-repo root dependencies.

ottodevs:
> Right now my process for getting this going:
> 
> * `ipfs daemon --offline` in any console
> * `npm run start:dev` in our repo
> * `npm run start:local` in the latest master of @aragon/aragon
> * `npm run cypress:open` in our repo followed by clicking on the `aragon.spec.js` test. (soon to be renamed openEnterprise.spec.js)

We may want to automate those steps in a single script / command that assumes the `aragon` repo path is accesible at `../aragon` (even A1 does this for some of their apps), and throw a warning msg in the terminal in case that folder is not accessible.

I was thinking today in how easy is to do that and the huge amount of time it would save us... 

It makes sense instead of waiting for aragon-cli to integrate it... because now it was the event triggers, but next will be the reputation-template UI, the gas indicators... we are constantly adding new updates to the client, so it is worth to add this simple script that `cd`s into the client folder and runs `npm run start:local`  or whatever needed.

topocount:
> > Right now my process for getting this going:
> > 
> > * `ipfs daemon --offline` in any console
> > * `npm run start:dev` in our repo
> > * `npm run start:local` in the latest master of @aragon/aragon
> > * `npm run cypress:open` in our repo followed by clicking on the `aragon.spec.js` test. (soon to be renamed openEnterprise.spec.js)
> 
> We may want to automate those steps in a single script / command that assumes the `aragon` repo path is accesible at `../aragon` (even A1 does this for some of their repos), and throw a warning msg in the terminal in case that folder is not accessible.
> 
> I was thinking today in how easy is to do that and the huge amount of time it would save us...
> 
> It makes sense instead of waiting for aragon-cli to integrate it... because now it was the event triggers, but next will be the reputation-template UI, the gas indicators... we are constantly adding new updates to the client, so it is worth to add this simple script that `cd`s into the client folder and runs `npm run start:local` or whatever needed.

ah great idea! I'll look into this further

templates/dev/scripts/create-dao.js
topocount: I wrote a script that sequentially generates dao id's here: https://github.com/AutarkLabs/open-enterprise/blob/3b65f63b1d400d5cebf553fb21b93b573803809f/templates/dev/scripts/getId.js

The param in the settings above will serve as an override if a dev wants to set their own ID. Leaving it empty or falsy allows the getID script to generate a sequential ID.

ottodevs:
I was having a hard time trying to test this locally, it seems my cypress version was stuck into `0.3.7` even after cleaning `node_modules`. I finally managed to update by manually calling `npm i cypress` from the root folder, and it was upgraded to `0.3.8` but the error is still there.

image

I did `npm i` from clean state and followed all the steps you mention Kevin. The DAO is loading outside cypress chrome (in my regular browser) but it seems to have errors in cypress...

Any idea about this?
image

(You might realize I even tried to change the dao address to that sequentially generated aragon-id, without success, because I have the same error)


BTW I triggered a new CI build and this time it passed green :)

ottodevs:
> Remaining TODOs:
> 
> * [ ]  acceptance-test @aragon/cli@6.4.x to see if the included client is updated with the event trigger and our heavily-used pathways are solid ❌ waiting for a release with the latest aragen updates
> * [ ]  update the `npm run cypress` command to cleanly set up a full-stack environment, run the tests, and then cleanly exit all services and scripts
> * [ ]  document the `npm run cypress` command as well as the development process for E2E tests

Are these TODOs going to be addressed in this PR, or should we create a separate ticket to track those? It would be nice if we clarify that 👍

package.json
topocount: After further thought I'm not sure this makes sense. I think if this causes any issues we can reevaluate.

cypress/integration/aragon.spec.js
topocount: I've decided to split the files up by app. I already appreciate how much cleaner the devX is with it this way. each test start only adds a handful of seconds while the browser environment initializes, so the cost is negligible, especially if we can get parallelization working.
> Can we run separate tests in parallel? Test Rewards in one process while we test Projects in another, for example?

We could possibly do parallelization now. I'll have to play around it with it more because I'm not sure how the web3 provider we're using willl handle nonces if multiple instances of it are committing transactions simultaneously. We apparently have a partner building a really smart web3 provider for this kind of thing but I haven't played with it at all yet.

cypress/integration/aragon.spec.js
topocount: clicking the button is still the easiest way to generate a popup handle and inject the message so I've kept this in

Quazia:
> My main concern here is that we are testing dev builds instead of staging/production builds.
> 
> My suggestion is to first define a staging/production build, that is deploying every piece that needs to be tested to a devchain (testing in rinkeby or production would complicate things as you depend on external factors as having ethers or setup permissions to update daos)
> 
> That way we only need to launch ipfs and devchain processes, removing the burden of launching this additional massive amount of services for every app+client that is not even going to be tested in every test)
> 
> This also improves and makes our build more consistent, as the tests will directly unable to launch if the build is failing.

Yeah, I think we can look at this as a potential enhancement going forward. For now I'd really like to get this code in so unless there's something strictly wrong with how it's being done here (which I don't really see anything as strictly wrong) I think we should approve and merge.

ottodevs:
> > My main concern here is that we are testing dev builds instead of staging/production builds.
> > My suggestion is to first define a staging/production build, that is deploying every piece that needs to be tested to a devchain (testing in rinkeby or production would complicate things as you depend on external factors as having ethers or setup permissions to update daos)
> > That way we only need to launch ipfs and devchain processes, removing the burden of launching this additional massive amount of services for every app+client that is not even going to be tested in every test)
> > This also improves and makes our build more consistent, as the tests will directly unable to launch if the build is failing.
> 
> Yeah, I think we can look at this as a potential enhancement going forward. For now I'd really like to get this code in so unless there's something strictly wrong with how it's being done here (which I don't really see anything as strictly wrong) I think we should approve and merge.

Yeah, that sounds good, let's merge this and work progressively into any improvement we find appropriate

PR #1678, opened 2019-11-20
Rewards/store: Fix claimed rewards' disappearance

### LINKED ISSUE #1518: Improve state management after a reward is claimed
- In My rewards: i claimed a reward, and the row for the reward disappeared after i claimed it
- I left the tab and returned to it, and then I see the reward in "Claimed" status

Expect: I expect to see the reward as "Claimed" as soon as I claim it

e18r:
*This PR is the product of some awesome team work. @chadoh helped me try to reproduce the issue locally; @stellarmagnet taught me how to inspect the state of a DAO on Rinkeby; @ottodevs generously spent many of his sleeping hours going with me through the steps to publish my
solution to Rinkeby and make it work there; and @topocount debugged an issue with claiming rewards. Thank you all so much!*

On Rinkeby, when a reward is claimed it suddenly disappears from My Rewards, and reappears after switching tabs. Since I was unable to reproduce this on my local environment, I investigated the root cause of this problem and found the following facts:

**F1**. The recently-claimed reward disappears because it gets filtered out from the `myRewards` object by the app state reducer.
Screenshot from 2019-11-20 13-46-36

**F2**. The app state reducer filters out the reward because its `userRewardAmount` property suddenly becomes zero.
Screenshot from 2019-11-19 10-26-05

**F3**. The reward with `userRewardAmount` set to zero comes from the store's `onRewardClaimed` function, which gets the reward's properties from `getRewardById`.

**F4**. The `getRewardById` function gets the `userRewardAmount` property from the `rewardAmount` return value of the contract's `getReward` function.
Screenshot from 2019-11-19 16-52-40

**F5**. The `rewardAmount` return value of the contract's `getReward` function is calculated dynamically by the contract's `_calculateRewardAmount` function.
Screenshot from 2019-11-19 16-54-50

**F6**. `rewardAmount` is the reward's amount multiplied by the user's balance at disbursement date, divided by the reference token's supply at disbursement date.
Screenshot from 2019-11-19 17-03-16
Screenshot from 2019-11-19 17-03-02

**F7**. In order for `rewardAmount` to be zero, either the reward's amount, the user's balance or the token's supply have to be zero.

After determining this, I formulated the following hypothesis:

**H1**. `rewardAmount` is zero because the user's balance is zero, because `_calculateRewardAmount` is receiving the wrong user; a user who doesn't possess any of that token. This happens when the store's `getRewardById` function doesn't receive a `userAddress` param.

My hypothesis H1 is supported by the following facts:

**F8**. On the UI, the reward reappears whenever tabs are switched. What happens underneath is that the store's `onRefreshRewards` function gets called. This function calls `getRewardById` just like `onRewardClaimed` (see F3), but unlike it, it specifies a user address.

**F9**. Another store function, `onRewardAdded`, also calls `getRewardById` without specifying a user address, and the same behaviour on the UI can be observed as in `onRewardClaimed`: the
reward's `rewardAmount` field is set to zero. This is less noticeable, but can be observed by creating a reward while on the My Rewards tab.

**F10**. The `userRewardAmount` property of the recently-claimed reward is not the only property that becomes zero. The same happens to `timeClaimed`. On the contract, `timeClaimed` is a map from addresses to integers. The contract's `getReward` function uses the calling user's address to get the value. If the user's address didn't exist on the map, that could explain why it ends up being zero.
Screenshot from 2019-11-19 10-26-43
Screenshot from 2019-11-19 19-45-07

**F11**. @topocount [solved a similar bug](https://github.com/AutarkLabs/open-enterprise/commit/11c9b3ff0fcbd105671a8d362562dab5904e3085?diff=split#diff-44e6c6b68470709dfa38518ab1f9c198R53) a while ago. The description he gave me of the bug matched perfectly H1.

In order to test my hypothesis, @ottodevs and I published my locally modified version of the Rewards app to Rinkeby, created a Rinkeby DAO and installed it there. The bug went away. You can test it [here](https://rinkeby.aragon.org/#/rewards45/0x632b710170835d55946cc717a50586267c2b6832/).
Peek 2019-11-20 12-17

So I can say with confidence that this PR fixes #1518 :)

The only remaining mistery is why can't this be reproduced locally. I don't know enough about the differences between Ganache and Rinkeby (or between the local environment and other blockchain
environments in general) in order to answer that question. But I do believe it's a relevant question that needs further inquiry in order to determine if the best path forward with similar issues is to
somehow make the local environment more closely resemble the blockchain, or to acknowledge the need to sometimes deploy custom Rinkeby DAOs, just like in this case.

Finally, it's important to note that **this PR modifies slightly the audited contract code**. That's because this is the best way to solve the bug at hand. But if modifying the contract code is not a possibility right now, I can think of some inelegant workarounds.

### DIFF
--- a/apps/rewards/contracts/Rewards.sol
+++ b/apps/rewards/contracts/Rewards.sol
@@ -62,7 +62,7 @@ contract Rewards is AragonApp {
     Vault public vault;
 
     /// Events
-    event RewardAdded(uint256 rewardId, address adder);
+    event RewardAdded(uint256 rewardId); /// Emitted when a new reward is created
     event RewardClaimed(uint256 rewardId, address claimant);
 
     /**
@@ -97,7 +97,7 @@ contract Rewards is AragonApp {
 
         _transferReward(reward, rewardAmount);
 
-        emit RewardClaimed(_rewardID, tx.origin);
+        emit RewardClaimed(_rewardID, msg.sender);
         return rewardAmount;
     }
 
@@ -216,7 +216,7 @@ contract Rewards is AragonApp {
         reward.delay = _delay;
         reward.blockStart = _startBlock;
         reward.creator = msg.sender;
-        emit RewardAdded(rewardId, tx.origin);
+        emit RewardAdded(rewardId);
         if (_occurrences > 1) {
             newReward(
                 _description,
--- a/apps/rewards/contracts/Rewards.sol
+++ b/apps/rewards/contracts/Rewards.sol
@@ -62,8 +62,8 @@ contract Rewards is AragonApp {
     Vault public vault;
 
     /// Events
-    event RewardAdded(uint256 rewardId); /// Emitted when a new reward is created
-    event RewardClaimed(uint256 rewardId); /// Emitted when a reward is claimed
+    event RewardAdded(uint256 rewardId, address adder);
+    event RewardClaimed(uint256 rewardId, address claimant);
 
     /**
      * @notice Initialize Rewards app for Vault at `_vault`
@@ -97,7 +97,7 @@ contract Rewards is AragonApp {
 
         _transferReward(reward, rewardAmount);
 
-        emit RewardClaimed(_rewardID);
+        emit RewardClaimed(_rewardID, tx.origin);
         return rewardAmount;
     }
 
@@ -216,7 +216,7 @@ contract Rewards is AragonApp {
         reward.delay = _delay;
         reward.blockStart = _startBlock;
         reward.creator = msg.sender;
-        emit RewardAdded(rewardId);
+        emit RewardAdded(rewardId, tx.origin);
         if (_occurrences > 1) {
             newReward(
                 _description,
--- a/apps/rewards/app/store/reward.js
+++ b/apps/rewards/app/store/reward.js
@@ -10,9 +10,13 @@ const CONVERT_API_BASE = 'https://min-api.cryptocompare.com/data'
 const convertApiUrl = symbols =>
   `${CONVERT_API_BASE}/price?fsym=USD&tsyms=${symbols.join(',')}`
 
-export async function onRewardAdded({ rewards = [], refTokens = [], balances = [] }, { rewardId }, settings) {
+export async function onRewardAdded(
+  { rewards = [], refTokens = [], balances = [] },
+  { rewardId, adder },
+  settings
+) {
   if (!rewards[rewardId]) {
-    rewards[rewardId] = await getRewardById(rewardId)
+    rewards[rewardId] = await getRewardById(rewardId, adder)
     const { referenceToken } = rewards[rewardId]
     const response = await updateBalancesAndRefTokens({ balances, refTokens }, referenceToken, settings)
     return { rewards, refTokens: response.refTokens }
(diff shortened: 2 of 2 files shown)


apps/rewards/contracts/Rewards.sol
```diff
@@ -62,8 +62,8 @@ contract Rewards is AragonApp {
     Vault public vault;
 
     /// Events
-    event RewardAdded(uint256 rewardId); /// Emitted when a new reward is created
-    event RewardClaimed(uint256 rewardId); /// Emitted when a reward is claimed
+    event RewardAdded(uint256 rewardId, address adder);
```
ottodevs: I don't think keeping the comments would hurt, but on the other hand the events names seem quite intuitive, so I am "ok" with the change.

apps/rewards/contracts/Rewards.sol
```diff
@@ -97,7 +97,7 @@ contract Rewards is AragonApp {
 
         _transferReward(reward, rewardAmount);
 
-        emit RewardClaimed(_rewardID);
+        emit RewardClaimed(_rewardID, tx.origin);
```
ottodevs: I am not really sure if this should be `msg.sender` instead... I invoke @Quazia for this or try to read this to ensure `tx.origin` is what we need: https://ethereum.stackexchange.com/questions/76929/tx-origin-alternative

apps/rewards/contracts/Rewards.sol
```diff
@@ -216,7 +216,7 @@ contract Rewards is AragonApp {
         reward.delay = _delay;
         reward.blockStart = _startBlock;
         reward.creator = msg.sender;
-        emit RewardAdded(rewardId);
+        emit RewardAdded(rewardId, tx.origin);
```
ottodevs: Same here: I am not really sure if this should be `msg.sender` instead... I invoke @Quazia for this or try to read this to ensure `tx.origin` is what we need: https://ethereum.stackexchange.com/questions/76929/tx-origin-alternative

apps/rewards/contracts/Rewards.sol
e18r: @ottodevs thank you for pointing this out. I decided to use `tx.origin` because in some cases the proxy address was showing up instead of the user's. But I'd need to test this more. Also, I'm gonna read about the security implications of `tx.origin`, which I was totally unaware of: https://github.com/ethereum/solidity/issues/683

apps/rewards/contracts/Rewards.sol
e18r: Yeah, a bit redundant and unnecessary. But if they're part of some convention or standard practice, it might make sense to leave them.

apps/rewards/contracts/Rewards.sol
e18r: Yes. After reading about it, `tx.origin` is insecure and [not future-proof](https://ethereum.stackexchange.com/a/200/7318). I'm going to test if `msg.sender` works as expected and let you know.

e18r:
@ottodevs I changed `tx.origin` to `msg.sender` for the case of claiming rewards, and I reverted the changes for the case of creating rewards, because `msg.sender` in that case issues the proxy address.

This means that the bug will still be present for the case of creating rewards, but in practice it's barely noticeable at all. It will only manifest itself if all of the following is true:

a) A user creates a reward while on the "My Rewards" tab
b) The created reward can be claimed by the user
c) The user can create rewards directly without going through a vote

This is a pretty rare situation and it isn't covered on the issue this PR fixes, so I consider it safe to leave it for an eventual future PR.

apps/rewards/contracts/Rewards.sol
Quazia: Yeah, neither one of these will behave as we want.

e18r: @Quazia after testing `msg.sender` it does work as expected for the claim transaction event. Or am I missing something?

PR #1680, opened 2019-11-21
Show SubmitWork option to issue owner only

rkzel:
Achieved by comparing issue.assignee to connectedAccount from useAragonApi()
Apart from that:

1. Icons used theme.* colors, but theme itself wasn't available. Adding
it required updating all colors.

2. Issues in in-progress state should have had 'View Applications'
option - it got added. "View", because another ticket requests all
"Review"s to be changed to "View"s.

3. Accept/Reject buttons in Review Application panel were broken, screens before/after:
Screenshot from 2019-11-21 16-53-11
Screenshot from 2019-11-21 17-14-31

### DIFF
--- a/apps/projects/app/components/Panel/ReviewApplication/ReviewApplication.js
+++ b/apps/projects/app/components/Panel/ReviewApplication/ReviewApplication.js
@@ -92,7 +92,7 @@ const ReviewApplication = ({ issue, requestIndex }) => {
         name="Applicant"
         items={issue.requestsData.map(request => request.user.login)}
         onChange={changeRequest}
-        selected={requestIndex}
+        selected={index}
         wide
       />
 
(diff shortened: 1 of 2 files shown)


apps/projects/app/components/Panel/ReviewApplication/ReviewApplication.js
```diff
@@ -173,16 +173,14 @@ const ReviewApplication = ({ issue, requestIndex }) => {
...
+              label="Reject"
+            />
             <ReviewButton
-              icon={<IconCheck />}
               mode="positive"
               onClick={onAccept}
-            >
-                Accept
-            </ReviewButton>
+              icon={<IconCheck />}
```
chadoh: Why re-order these props? Looks like they were in alphabetical order before. What order are they in now?

apps/projects/app/components/Shared/BountyContextMenu.js
```diff
@@ -63,20 +75,20 @@ const BountyContextMenu = ({ issue }) => {
...
             <ActionLabel>
               Submit Application
             </ActionLabel>
           </Item>
           <Item onClick={() => reviewApplication(issue)}>
-            <IconView color={`${theme.surfaceIcon}`} />
+            <IconView color={`${theme.surfaceContent}`} />
             <ActionLabel>
-              Review Application {issue.requestsData ? `(${issue.requestsData.length})` : ''}
+              Review Applications ({issue.requestsData.length})
```
chadoh: Why were we able to get rid of this check?

PR #1692, opened 2019-11-26
Activity log updated

rkzel:
Events are: opening of the issue, placing/updating associated bounty,
applying for work, review (and approval/rejection), submitting work
(and approval/rejection).

Changes: work/application Panels are opened by Links instead of buttons,
descriptions and comments are not shown, user login does not link
to GitHub, descriptions of events are as per design (slightly changed).

Before:

Screenshot from 2019-11-26 15-52-01

After:

Screenshot from 2019-11-26 18-02-10


apps/projects/app/components/Content/IssueDetail/EventsCard.js
```diff
@@ -28,13 +25,6 @@ const IssueEvent = ({ user, ...props }) => {
         </IssueEventAvatar>
         <IssueEventDetails>
           <Text.Block size="small">
-            <Link
-              href={user.url}
-              target="_blank"
-              style={{ textDecoration: 'none', color: `${theme.link}` }}
-            >
-              {user.login}
```
chadoh: I want to double-check that we want to remove the link to the user's profile on GitHub. That seems useful to me.

stellarmagnet: The avatar should be clickable. Is that not the case?

rkzel: It wasn't the case before, it is now. One additional question then:

Screenshot from 2019-11-27 14-54-59

In panels it's the old way - avatar photo and username, usename being the link. Adjust? Username as plain text, avatart photo as link?

rkzel: I realized there is more to be fixed in the panels. Needs separate ticket and PR. Created #1696 for it.

PR #1698, opened 2019-11-28
Review Panels, styling fixes and cleanup

rkzel:
- Improved displaying avatar in activity log and panels: only avatar photo links to GH, username is plaintext.
- Spacing in panels has been fixed, between field titles and texts.
- Many components doubled in Application/Work View panels got moved to PanelComponents and imported from there.

Before:

Screenshot from 2019-11-27 14-54-59
Screenshot from 2019-11-22 21-10-54

After:

Screenshot from 2019-11-28 16-09-52
Screenshot from 2019-11-28 16-17-15


apps/projects/app/components/Content/IssueDetail/EventsCard.js
```diff
@@ -21,7 +21,13 @@ const IssueEvent = ({ user, ...props }) => {
     <IssueEventMain>
       <div css="display: flex">
         <IssueEventAvatar>
-          <img src={user.avatarUrl} alt="user avatar" css="width: 40px; border-radius: 50%" />
+          <Link
```
chadoh: Why not use the new `Avatar` component here?

apps/projects/app/components/Panel/PanelComponents.js
```diff
@@ -40,14 +42,111 @@ export const IssueTitle = ({ issue }) => (
...
-  overflow: hidden;
-  text-overflow: ellipsis;
+export const Avatar = ({ user }) => {
+  const theme = useTheme()
+
+  return (
+    <Link
+      href={user.url}
+      target="_blank"
+      style={{ textDecoration: 'none', color: `${theme.link}`, marginRight: 6 }}
```
chadoh: Why not this?

```jsx
css={`text-decoration: none; color: ${theme.link}; margin-right: ${GU}px`}

File: apps/projects/app/components/Panel/PanelComponents.js [hunk 3]
```diff
@@ -40,14 +42,111 @@ export const IssueTitle = ({ issue }) => (
...
+        alt=""
+        src={user.avatarUrl}
+        css="width: 40px; height: 40px; margin-right: 10px; border-radius: 50%;"
+      />
+    </Link>
+  )
+}
+Avatar.propTypes = {
+  user: PropTypes.shape({
+    url: PropTypes.string.isRequired,
```
chadoh: I think we only use `avatarUrl`

apps/projects/app/components/Panel/PanelComponents.js
```diff
@@ -40,14 +42,111 @@ export const IssueTitle = ({ issue }) => (
...
+  const theme = useTheme()
+
+  return (
+    <Link
+      href={user.url}
+      target="_blank"
+      style={{ textDecoration: 'none', color: `${theme.link}`, marginRight: 6 }}
+    >
+      <img
+        alt=""
```
chadoh: I'm curious about the impact to usability this had. We moved the link to the Avatar, rather than the username, and I think we are hiding the avatar from screen readers by setting `alt=""`. Maybe we could set the alt text to something like `{user.username} from GitHub`. Though that makes the text weird. "chadoh from GitHub chadoh applied 1 day ago" 😕

@javieralaves this might be a question for you. I actually think the way we had it before makes it easier to make the experience good for both sighted and blind users. We could easily hide the avatar from blind users, setting `alt=""`, and then make both the avatar and the username link to GitHub. Can we do that? What was the thinking behind removing the link from the username?

apps/projects/app/components/Panel/PanelComponents.js
```diff
@@ -40,14 +42,111 @@ export const IssueTitle = ({ issue }) => (
...
+        {formatDate(new Date(review.reviewDate), 'd MMM yy, h:MM a (z)')}
+      </Text>
+    </div>
+  )
+}
+Status.propTypes = {
+  review: PropTypes.shape({
+    approved: PropTypes.bool.isRequired,
+    reviewDate: PropTypes.string.isRequired,
+  }),
```
chadoh: `isRequired`?

apps/projects/app/components/Panel/ReviewApplication/ReviewApplication.js
```diff
@@ -69,16 +73,16 @@ const ReviewApplication = ({ issue, requestIndex }) => {
...
+      avatarUrl: request.user.avatarUrl,
       url: request.user.url
     },
     workplan: request.workplan,
     hours: request.hours,
     eta: (request.eta === '-') ? request.eta : (new Date(request.eta)).toLocaleDateString(),
     applicationDate: request.applicationDate
   }
-
+
```
chadoh: Please configure your editor to strip trailing whitespace when you save a file 🙂

apps/projects/app/components/Panel/PanelComponents.js
rkzel: `url` is for the user. `avatarUrl` is for user's avatar pic.

apps/projects/app/components/Panel/PanelComponents.js
rkzel: Yes. Both are used, both are required...

apps/projects/app/components/Panel/ReviewApplication/ReviewApplication.js
rkzel: It actually is configured that way, although it got confused and stopped removing whitespaces. It needed restarting to get that functionality back. Thank you, VSCode...

apps/projects/app/components/Panel/PanelComponents.js
javieralaves: @chadoh The objective was to focus the use of hyperlinks around applications and work submissions, but at the same time I don't have a strong opinion on bringing back the profile link as well. Note that as we move onto the 'GitHub decoupling' territory, some additional changes will be taking place very soon to this panel, in specific making use of some form of AddressBadge instead of what we have at the moment pointing at a GitHub profile; so we may want to wait until then to decide more in detail how to enhance accessibility.

apps/projects/app/components/Panel/PanelComponents.js
rkzel: Probably historicaly reasons. Fixed, both places.

apps/projects/app/components/Content/IssueDetail/EventsCard.js
rkzel: Fixed.

apps/projects/app/components/Panel/PanelComponents.js
chadoh: ok, I like keeping as-is for now. Let's think about this when we decouple from GitHub, as you suggest.

apps/projects/app/components/Panel/PanelComponents.js
chadoh: But is `review` itself required? If so, we can add `isRequired` at the end here, too.

```js
  review: PropTypes.shape({
    approved: PropTypes.bool.isRequired,
    reviewDate: PropTypes.string.isRequired,
  }).isRequired,
```

topocount:
@rkzel please fix the merge conflicts

apps/projects/app/components/Panel/PanelComponents.js
rkzel: Fixeth in both places.

PR #1720, opened 2019-12-04
add storage to oe template (not working)

Schwartz10:


Schwartz10:
https://hack.aragon.org/docs/cli-dao-commands#dao-install

PR #1722, opened 2019-12-04
AragonUI 1.0.0 upgrade

### LINKED ISSUE #1684: Update all apps to aragonUI 1.0
This basically involves updating the package.json for each app, and going through the test script and looking very closely at all UI components to:
- Make sure nothing looks broken or wonky
- Make sure you can click on all clickable elements
- Make sure everything fits properly in the page in all resolutions

e18r:
Fixes #1684 

Upgraded all apps and followed (and improved) the test script on each of them. The test script was run twice on each app, before and after the upgrade, to make sure to detect only those bugs that were specific to the upgrade. Only found one bug caused by the upgrade, which is solved in this PR, and also found three more bugs that aren't related to this upgrade and weren't reported before. Created tickets for them.


Quazia:
LGTM, I'm just re-starting the testing but once CI passes this is ready to merge. I'm hoping restarting the build will resolve the issue.

e18r:
CI seems to have passed :)

apps/address-book/package.json
```diff
@@ -32,7 +32,7 @@
   "dependencies": {
     "@aragon/api": "2.0.0-beta.8",
     "@aragon/api-react": "2.0.0-beta.6",
-    "@aragon/ui": "^1.0.0-alpha.26",
+    "@aragon/ui": "^1.0.0",
```
ottodevs: I think it is important to use the same version, as the `^` character allows for every version in the range between `0.9.9` and `2.0.0` (not included), so in this case, `lerna` decides not to hoist the `ui` package for this app:

```
lerna WARN EHOIST_PKG_VERSION "@tps/apps-address-book" package depends on @aragon/ui@^1.0.0, which differs from the hoisted @aragon/ui@1.0.0.
lerna WARN EHOIST_PKG_VERSION "@tps/apps-discussions" package depends on @aragon/ui@1.0.0-alpha.11, which differs from the hoisted @aragon/ui@1.0.0.
```

I would also update discussions `ui`

ottodevs: Basically: pin it in all apps, or not pin it all

e18r:
@ottodevs Thank you for your review. I removed the ^ sign from the AragonUI's version number in Address Book and made sure the warning disappeared in the console when installing.

As for the Discussions app, @Schwartz10 upgraded it in another PR.

PR #1724, opened 2019-12-04
Dot Voting: Implement deep linking

PeterMPhillips:
Deep linking for Dot Voting. Uses app-logic.js to handle the path and parse the the voteId.
DotDeepLinking


ottodevs:
I was experimenting some issues trying to launch dot voting for review, I am still researching into it, but I temporarily skipped this to advance on the rest of reviews.

Revisiting today ASAP

PR #1740, opened 2019-12-06
Rewards: Deep linking

PeterMPhillips:
Rewards Deep Linking


javieralaves:
Is this supposed to be part of an existing epic?

PeterMPhillips:
@javieralaves I'm unsure. Best ask @ottodevs

ottodevs:
> Is this supposed to be part of an existing epic?

Yes it is @javieralaves it is part of https://github.com/AutarkLabs/product/issues/13

apps/rewards/app/components/App/App.js
```diff
@@ -32,16 +33,19 @@ class App extends React.Component {
     refTokens: PropTypes.array.isRequired,
     amountTokens: PropTypes.array.isRequired,
     claims: PropTypes.array.isRequired,
+    tabs: PropTypes.array.isRequired,
```
ottodevs: We need to take type checking seriously, From now on, I personally will not consider any PR completed unless their PropTypes are correctly shaped.

apps/rewards/app/app-logic.js
```diff
@@ -0,0 +1,65 @@
...
+    str = str.replace(' ', '')
+  }
+  return str
+}
+
+export const usePathSegments = () => {
+  const [ path, requestPath ] = usePath()
+
+  const fromPath = useMemo(() => {
+    const [ , urltab, rewardId ] = path.match(PATH_REGEX) || []
```
ottodevs: We use to follow conventions and define variables with `camelCase` names
```suggestion
    const [ , urlTab, rewardId ] = path.match(PATH_REGEX) || []
```

apps/rewards/app/app-logic.js
```diff
@@ -0,0 +1,65 @@
...
+      rewardId: rewardId ? String(rewardId) : null
+    }
+    return pathSegments
+  }, [path])
+
+  const selectTab = useCallback(
+    tabIndex => {
+      const tab = urlify(TABS[tabIndex])
+      // For consistency "/overview/"" path will be transformed to root "/" path
+      const newpath = (!tabIndex || tabIndex < 1) ? '' : `/${tab}`
```
ottodevs: camelCase please

apps/rewards/app/components/App/App.js
```diff
@@ -111,6 +125,25 @@ class App extends React.Component {
     })
   }
 
+  updatePath = (props) => {
+    const { selected, rewardId } = props.fromPath
+    this.state.selected = selected
```
ottodevs: The state should never be directly assigned as this, this is forbidden by react rules. Apart from that, I don't think we need to use state, when we can just rely on the prop directly.

ottodevs: We can instead use defaultProps if we need a default value? I don't really understand if that is what you were trying here.

apps/rewards/app/components/App/App.js
```diff
@@ -193,20 +228,21 @@ class App extends React.Component {
...
     })
+    this.props.selectReward(this.state.selected, String(reward.rewardId))
   }
 
   claimReward = reward => {
     // TODO
     this.props.api.claimReward(reward.rewardId + reward.claims).toPromise()
   }
-
+  /*
```
ottodevs: This is not needed, for the PR to be considered complete we should clean these kind of comments and dead code

apps/rewards/package.json
```diff
@@ -31,7 +31,7 @@
   },
   "dependencies": {
     "@aragon/api": "2.0.0-beta.8",
-    "@aragon/api-react": "2.0.0-beta.6",
+    "@aragon/api-react": "2.0.0-beta.8",
```
ottodevs: It is needed to update the rest of apps as well, otherwise the custom api-react logic from the stubbed api will use the hoisted version (the old one in this case), causing error in the frontend and preventing the app to launch.

apps/rewards/app/components/App/App.js
```diff
@@ -32,16 +33,19 @@ class App extends React.Component {
     refTokens: PropTypes.array.isRequired,
     amountTokens: PropTypes.array.isRequired,
     claims: PropTypes.array.isRequired,
+    tabs: PropTypes.array.isRequired,
+    fromPath: PropTypes.object.isRequired,
```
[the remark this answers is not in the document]
ottodevs: I tried to fix this one in 0271a6b44fb9deea7b2831014e6c19f698be447c

apps/rewards/app/app-logic.js
```diff
@@ -0,0 +1,65 @@
+import PropTypes from 'prop-types'
```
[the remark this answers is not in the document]
ottodevs: I fixed this in b99b2aa2e43941637c74582a4a71ade4d8635147

apps/rewards/app/app-logic.js
ottodevs: Fixed in b99b2aa2e43941637c74582a4a71ade4d8635147

apps/rewards/app/app-logic.js
ottodevs: Fixed in b99b2aa2e43941637c74582a4a71ade4d8635147

apps/rewards/package.json
ottodevs: Fixed in 5b4e86af98d96ab1aee064fd9341a66d7f03276f

apps/rewards/app/app-logic.js
```diff
@@ -0,0 +1,65 @@
...
+import React, { useCallback, useMemo, } from 'react'
+import { AragonApi, usePath } from '@aragon/api-react'
+
+const PATH_REGEX = /^\/(rewards|myrewards)(?:\/([0-9])?)?(?:\/)?$/
+
+const TABS = [ 'Overview', 'My Rewards' ]
+
+const urlify = (str) => {
```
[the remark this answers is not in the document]
ottodevs: Fixed b9b6631

PeterMPhillips:
I made the requested fixes but I had to change some of the propTypes that you had set. I returned `tabs` to an `array` instead of `oneOf(TABS)` and I changed `requestId` to a `string` since that is the type that is received from the path and the contract. I can change it to `number` but that would mean converting several other variables as well.

apps/rewards/app/components/App/App.js
```diff
@@ -27,7 +27,7 @@ class App extends React.Component {
     displayMenuButton: PropTypes.bool.isRequired,
     fromPath: PropTypes.exact({
       selected: PropTypes.number,
-      rewardId: PropTypes.number
+      rewardId: PropTypes.string
```
ottodevs: I see what you mean, we need to really change this type, but it falls out of scope since it is coming from the store this way.

apps/rewards/app/components/App/App.js
```diff
@@ -38,7 +38,7 @@ class App extends React.Component {
     rewards: PropTypes.arrayOf(PropTypes.object).isRequired,
     selectReward: PropTypes.func.isRequired,
     selectTab: PropTypes.func.isRequired,
-    tabs: PropTypes.oneOf(TABS).isRequired,
+    tabs: PropTypes.array.isRequired,
```
ottodevs: It was working for me but 🤷‍♂️

PeterMPhillips:
> Ok, I still see some minor changes that you didn't address (like removing the commented code block), but I guess it is fair enough and it will be harmless for now.
> 
> The main goal was achieved and the code it is overall clear and succinct
> 
> Thanks for the work!

Weird. I for sure deleted that commented block, I must have pressed ctrl-z a bit too many times while working on something else.

PR #1742, opened 2019-12-08
Create Dao Recovery Script

topocount:
This script allows anyone who can run `npm run recoverTemplate` the
ability to generate a custom data string which can be given to users to
recover their DAO if they're unable to complete the second transaction
when deploying a dao from the Open Enterprise Template


package.json
```diff
@@ -62,6 +62,7 @@
     "publish:cd": "shared/deployments/publish-cd.sh",
     "publish:apps": "lerna run publish",
     "publish:http": "lerna run --scope='@tps/apps-*' publish:http --stream",
+    "recoverTemplate": "cd templates/open-enterprise && npm run recover:tx",
```
ottodevs: Did you consider putting this script into a javascript file and adding it to a `bin` section here?
This way the potential users coming installing our packages from `npm` would have an easier way to call this, as a command, for example, the same as `aragon` is command runs for `@aragon/cli`...

I think for this use case might be more appropriate, since lot of users would not know the inner-workings of running `package.json` scripts or figure out if they should use `npm`, `yarn` or whatever... and it doesn't take a lot of effort to make that change...

templates/open-enterprise/scripts/recover-dao.js
```diff
@@ -0,0 +1,95 @@
...
+}
+
+const waitASecond = () => {
+  return new Promise((resolve) => {
+    setTimeout(resolve, 1000)
+  })
+}
+const getVoteDuration = () => {
+  return new Promise((resolve) => {
+    rl.question('enter the desired dot voting period length, in days > ', (answer) => {
```
ottodevs: Why in days? we allow even seconds to be set, I would set it to seconds instead of limit the user, it takes literally more work to limit it to be days (because of the calculation), unless there is a good reason for this.

templates/open-enterprise/scripts/recover-dao.js
```diff
@@ -0,0 +1,95 @@
...
+      resolve(answer * 1e16)
+    })
+  })
+}
+
+const getQuorum = (minSupport) => {
+  return new Promise((resolve) => {
+    rl.question('enter the dot voting quorum percentage > ', async (answer) => {
+      if ((answer * 1e16) > minSupport) {
+        rl.write('Error: Quorum must be less than or equal to minimum candidate support percentage\n')
```
ottodevs: There are more conditions that should be added to prevent the tx to fail:
```
        require(_minQuorum > 0, ERROR_MIN_QUORUM);
        require(_minQuorum <= PCT_BASE, ERROR_MIN_QUORUM);
        require(_minQuorum >= _candidateSupportPct, ERROR_MIN_QUORUM); /// <- You just added this one
```

templates/open-enterprise/scripts/recover-dao.js
```diff
@@ -0,0 +1,95 @@
...
+})
+const OpenEnterpriseTemplate = fs.readFileSync('../build/contracts/OpenEnterpriseTemplate.json')
+
+const zeroAddress = '0x00000000000000000000000000000000000000000'
+
+let deployedInfo = null
+
+// This function queries APM for the contract address of the Open Enterprise template
+// The script will not exit until this address is resolve or the request fails or times out
+exec('aragon apm info open-enterprise-template --env production', (err, stdout, stderr) => {
```
ottodevs: Really smart solution Kevin!
👏

templates/open-enterprise/scripts/recover-dao.js
topocount: I thought the average production user would be thinking in the timescale of days, but if you think seconds makes more sense we can do that.

package.json
topocount: I saw this as something we would use to help support users who run into this issue. I don't think prompting users to install this repo is that the best way to go. If we want to give users a self-service option, building a a microsite on top of this logic that creates the tx for them in metamask would be a huge improvement for relatively little work, but we're not sure it's necessary yet.

topocount:
@ottodevs this is intended as an internal tool. I've written some documentation for the team here: https://www.notion.so/autark/DAO-Recovery-9e64e29c56884a748cbfeb1b5eccba61.

templates/open-enterprise/scripts/recover-dao.js
ottodevs: I don't think the average production user will ever use this without guided support from anyone of the Autark team, because of the complexity using this tool involves (node packages, cli commands, truffle, etc), in the way it is currently designed, I would think of us (Autark team) as the target user of it (while providing support to other people).

If you seriously think this must be a tool for production users, I would think in differently architecting some assumptions (as I briefly indicated in other comments) , like providing encapsulated binaries or even a GUI tool, but for now I really think, for the support scenario I mention, it is ideal in its state, and a perfect starting point to iterate from, in case we want it to evolve that way 👍 

That being said, it is obviously not strictly needed to change it from days to seconds, but I think even a production user could come up with a DAO idea where votes period is less than one day, I don't think it would be justified to limit something that is allowed in the contract, just because of the target user, because that would be providing unfair advantage to other people. My point is, limit everywhere or not limit at all... IMO all our tools should mimic 100% the contract logic, but please, again, consider this a very personal opinion! I understand your motivation for simplification!

package.json
ottodevs: It seems a good idea Kevin, but right now, the way it is implemented there is no chance to run this without installing all the dependencies, like truffle, @aragon/cli, node, etc, so I in its current state I think it is a better idea we think of it as a tool for us (Autark) to provide support to others, but it is still a bit complex to use for a "regular" user in those terms...

PR #1743, opened 2019-12-09
Open submission flow added

rkzel:
Requirement: to be able to submit work for bounties without
being an approved (via Application Review) owner.

On UI side it is done by Switch added to Fund Issue form.
The Switch sets openSubmission bool, which ends up in IPFS
data file, and it is later retrieved from there. Hence changes
to the store.

openSubmission variable is to be used in changes made in other
tickets, it doesn't do anything special in this one.

Screen:
Screenshot from 2019-12-09 12-19-55


apps/projects/app/components/Content/IssueDetail/BountyCard.js
```diff
@@ -83,8 +92,9 @@ Status.propTypes = issueShape
 
 const Submissions = ({ issue }) => {
   const { reviewApplication, reviewWork } = usePanelManagement()
+  const workStatus = issue.openSubmission ? 'openSubmission' : issue.workStatus
```
chadoh: Same thing here

apps/projects/app/components/Content/IssueDetail/BountyCard.js
```diff
@@ -147,13 +158,16 @@ BountyDot.propTypes = PropTypes.string.isRequired
 const Action = ({ issue }) => {
   const { requestAssignment, submitWork } = usePanelManagement()
   const { connectedAccount } = useAragonApi()
+  const workStatus = issue.openSubmission ? 'openSubmission' : issue.workStatus
```
chadoh: Same thing here

apps/projects/app/components/Panel/FundIssues/FundIssues.js
```diff
@@ -1,5 +1,3 @@
-/* eslint-disable react/prop-types */
-// issues are validated using correct shape - eslint problem?
```
chadoh: I don't like that we're disabling this for the entire file, especially such a large file as this. What problem was this working around?

apps/projects/app/components/Panel/FundIssues/FundIssues.js
```diff
@@ -63,6 +63,7 @@ const bountiesFor = ({ bountySettings, issues, tokens }) => issues.reduce(
       slotsIndex: 0,
       payout: issue.payout || 0,
       token: tokens.find(t => t.symbol === issue.symbol) || tokens[0],
+      openSubmission: issue.openSubmission,
```
chadoh: We are setting this on a per-issue basis, but in the form we are setting it globally, for all issues-being-funded. Is that correct?

Do we use this version of `openSubmission`? Can we remove it?

apps/projects/app/components/Panel/FundIssues/FundIssues.js
chadoh: Oh, nevermind, you removed this. Good! 😄

apps/projects/app/components/Panel/FundIssues/FundIssues.js
rkzel: Removed.

apps/projects/app/components/Content/IssueDetail/BountyCard.js
rkzel: Fixed.

apps/projects/app/components/Content/IssueDetail/BountyCard.js
rkzel: Fixed.

apps/projects/app/components/Content/IssueDetail/BountyCard.js
```diff
@@ -50,8 +50,17 @@ const pluralize = (word, number) => `${number} ${word}${number > 1 ? 's' : ''}`
 
 const Status = ({ issue }) => {
   const theme = useTheme()
+  const workStatus = issue.openSubmission ? 'openSubmission' : issue.workStatus
```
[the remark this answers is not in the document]
rkzel: Fixed.

PR #1745, opened 2019-12-09
BountyContextMenu modernized up to Figma design

rkzel:
- "open work submission" bounties got added,
- giant {condition && displayStuff)} was replaced by more readable giant
switch statement
- menu items are now using a common component,
- pluralization is better ('application' vs 'applications (2)')

Relevant screens:

Issue funded without application required

Screenshot 1

Issue funded without application required + submissions submitted

Screenshot 2

Issue funded with application required + application submitted (more than one)

Screenshot 3

Issue funded with application required + application assigned, assignee view:

Screenshot 45

...and public view:

Screenshot 4

Issue funded with application required + work submitted

Screenshot 7

Fulfilled:

Screenshot 8


apps/projects/app/components/Shared/BountyContextMenu.js
```diff
@@ -17,110 +41,84 @@ const BountyContextMenu = ({ issue }) => {
...
-              </ActionLabel>
-            </Item>
-          )}
-          <Item onClick={() => reviewApplication(issue)}>
-            <IconView color={`${theme.surfaceContent}`} />
-            <ActionLabel>
-              View applications ({issue.requestsData.length})
-            </ActionLabel>
-          </Item>
+  const pluralize = (word, number) => `${word}${number > 1 ? 's (' + number + ')' : ''}`
```
chadoh: Please move this outside the component

apps/projects/app/components/Shared/BountyContextMenu.js
```diff
@@ -1,14 +1,38 @@
...
+      </ActionLabel>
+    </Item>
+  )
+}
+MenuItem.propTypes = {
+  panel: PropTypes.func.isRequired,
+  caption: PropTypes.string.isRequired,
+  panelParams: PropTypes.oneOfType([
+    PropTypes.object,
+    PropTypes.arrayOf(PropTypes.object)
```
chadoh: Looks like we are only ever setting `panelParams={issue}`, so we don't need this to allow arrays. Let's not generalize our interfaces/type checking more than we need; that's a form of premature optimization. 

Also, since we only ever pass `issue` into this prop, why not just call it `issue`? We're only using `MenuItem` within this file; it doesn't need to be so general in its language.

apps/projects/app/components/Shared/BountyContextMenu.js
```diff
@@ -17,110 +41,84 @@ const BountyContextMenu = ({ issue }) => {
...
-          {/*  </ActionLabel>*/}
-          {/*</Item>*/}
-        </React.Fragment>
+      <MenuItem
+        panel={reviewWork}
+        panelParams={issue}
+        caption={pluralize('View work submission', issue.workSubmissions.length)}
+        Icon={IconView}
+      />
+      {openSubmission || (
```
chadoh: Does this `||` work as expected? Do we end up with `true` in the UI sometimes?

I think I'd prefer we rephrased this as `!openSubmission &&`

apps/projects/app/components/Shared/BountyContextMenu.js
```diff
@@ -17,110 +41,84 @@ const BountyContextMenu = ({ issue }) => {
...
+  )
+  case 'fulfilled': return (
+    <>
+      <MenuItem
+        panel={reviewWork}
+        panelParams={issue}
+        caption={pluralize('View work submission', issue.workSubmissions.length)}
+        Icon={IconView}
+      />
+      {openSubmission || (
```
chadoh: Same thing here

apps/projects/app/components/Shared/BountyContextMenu.js
rkzel: Fixed.

apps/projects/app/components/Shared/BountyContextMenu.js
rkzel: We've seen something similar elsewhere... But in this particular case there is one place we have:

`<MenuItem panel={allocateBounty} panelParams={[issue]} caption="Fund issue" .../>`

BountyContextMenu is used in singular issue (IssueDetails) and in Actions menu of Issues Tab, and there an array of issues is (for now) perfectly valid as parameter. Funding Panel is an exception, all other panels (apart from Curation actually) deal with single issues. So, for now this needs to stay as it is - and I called it `panelParams`, precisely because it's not always just one issue...

apps/projects/app/components/Shared/BountyContextMenu.js
rkzel: It works as expected, but sure, I changed it to more standard `!openSubmission &&`.

apps/projects/app/components/Shared/BountyContextMenu.js
rkzel: Changed.

PR #1746, opened 2019-12-09
Removed bounty info from IssueDetails

rkzel:
Because all the info now lives in Bounty widget (former Status widget).
Also: cleaned up IssueDetails component - a lot of code was no longer
required. PanelComponents got touched, because Issue Title is a very
reusable component - in all panels, and issue details as well.
What needed to be done was restricting Link to only issue's ID
(it used to be whole div - including GH icon), and adding extra
children prop to the component, so I could do this:
```
<IssueTitleLink issue={issue}>
   {dot}
   <IconInfo
...
```
as per design. One note: we only download open issues right now.
What happens when an issue a bounty is attached to becomes closed...
That requires a separate ticket.

### DIFF
--- a/apps/projects/app/components/Content/IssueDetail/DetailsCard.js
+++ b/apps/projects/app/components/Content/IssueDetail/DetailsCard.js
@@ -19,7 +19,7 @@ import { formatDistance } from 'date-fns'
 const OpenedAgo = ({ date }) =>
   formatDistance(new Date(date), new Date(), { addSuffix: true })
 
-const dot = <span css="margin: 0px 10px">&middot;</span>
+const Dot = () => <span css="margin: 0px 10px">&middot;</span>
 
 const Title = ({ issue }) => {
   const theme = useTheme()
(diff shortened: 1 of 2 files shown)


apps/projects/app/components/Content/IssueDetail/DetailsCard.js
```diff
@@ -15,147 +16,109 @@ import { issueShape } from '../../../utils/shapes.js'
...
+      flex-direction: column;
+      flex-basis: 100%;
+      flex: 2;
+      margin-right: 20px;
     `}>
-      {children}
+      <Text.Block size="xxlarge" style={{ marginBottom: 1.5 * GU + 'px' }}>
+        {issue.title}
+      </Text.Block>
+      <IssueTitleLink issue={issue}>
```
chadoh: Maybe this instead?

```js
<IssueTitleLink
  issue={issue}
  append={
    <>
      <Dot />
      <IconInfo ... />
      Opened ...
    </>
  }

Comment (chadoh) [reply apps/projects/app/components/Content/IssueDetail/DetailsCard.js]: Or maaaybe we don't need to pass any new props to `IssueTitleLink`?

```js
<div css="display: flex">
  <IssueTitleLink issue={issue} />
  <Dot />
  <IconInfo ... />
  <span>Opened ...</span>
</div>

File: apps/projects/app/components/Content/IssueDetail/DetailsCard.js [hunk 2]
```diff
@@ -15,147 +16,109 @@ import { issueShape } from '../../../utils/shapes.js'
...
 
-const SummaryCell = ({ label, children, grid }) => (
-  <div css={`
-    display: flex;
-    flex-direction: column;
-    grid-area: ${grid};
-    overflow: hidden;
-  `}>
-    <Label text={label} />
+const dot = <span css="margin: 0px 10px">&middot;</span>
```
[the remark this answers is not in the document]
rkzel: Changed.

apps/projects/app/components/Content/IssueDetail/DetailsCard.js
rkzel: With careful use of display: flex (actually checking all css directives) it was possible to achieve the same effect without passing any additional props/children to IssueTitleLink - like in the second snippet :)

PR #1750, opened 2019-12-09
Minimum Viable AragonRouter

chadoh:
The magic of ReactRouter is mostly powered by the `history` module. To make our own AragonRouter, I implemented a minimal version of `AragonHistory`. Then, in the AragonRouter component, I listen for updates to `path` and pass them along to `AragonHistory`.

I've demonstrated how this can be used by implementing some basic routing in Allocations. Note that the allocations app itself never needs to bother with `usePath`.

### DIFF
--- a/apps/address-book/package.json
+++ b/apps/address-book/package.json
@@ -31,7 +31,7 @@
   },
   "dependencies": {
     "@aragon/api": "2.0.0-beta.8",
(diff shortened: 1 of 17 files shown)


chadoh:
Closing for now; this was a good proof-of-concept but I don't want to distract us with this conversation right now. I've implemented the Budget Detail view with raw `usePath`; see it at #1758

PR #1754, opened 2019-12-09
Rewards/MyRewards: Double Check Icon

### LINKED ISSUE #1639: Rewards:/MyRewards: change claimed icon
Change the icon for the claimed status from a circled check to a double check, according to the designs: https://www.figma.com/file/23c7bPgSZpY5lgQR6PmHsA/Rewards?node-id=311%3A0

e18r:
Fixes #1639

Screenshot from 2019-12-09 17-44-05


stellarmagnet:
The "description" and "disbursement date" columns are a bit too close...
The description column seems really short in width. What does this end up looking like if the description was longer?

topocount:
I agree, but the claimed date wasn't added in this PR. I think the claimed
date should be in a tooltip or something else where it's not monopolizing
table width

PR #1756, opened 2019-12-10
Both Review panels can be now read-only

rkzel:
A bool variable has been introduced, so links in Activity log
can modify default behavior - which is to show review form
in those panels (unless review had already taken place).

No cleanup was intended here - there is another PR dealing
with Status and Avatar components. This PR is to strictly
add RO mode to the panels.

Screen (work submission, application submission looks pretty much the same):
Screenshot from 2019-12-10 16-25-04

### DIFF
--- a/apps/projects/app/components/Content/IssueDetail/EventsCard.js
+++ b/apps/projects/app/components/Content/IssueDetail/EventsCard.js
@@ -53,13 +53,13 @@ IssueEvent.propTypes = {
 
 const applicationLink = (user, onReviewApplication, issue, index) => (
   <React.Fragment>
-    {user} submitted <Link onClick={() => onReviewApplication(issue, index)}>an application for review</Link>
+    {user} submitted <Link onClick={() => onReviewApplication(issue, index, true)}>an application for review</Link>
   </React.Fragment>
 )
 
 const workLink = (user, onReviewWork, issue, index) => (
   <React.Fragment>
-    {user} submitted <Link onClick={() => onReviewWork(issue, index)}>work for review</Link>
+    {user} submitted <Link onClick={() => onReviewWork(issue, index, true)}>work for review</Link>
   </React.Fragment>
 )
 
--- a/apps/projects/app/components/Content/IssueDetail/EventsCard.js
+++ b/apps/projects/app/components/Content/IssueDetail/EventsCard.js
@@ -53,13 +53,17 @@ IssueEvent.propTypes = {
 
 const applicationLink = (user, onReviewApplication, issue, index) => (
   <React.Fragment>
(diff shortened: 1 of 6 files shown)


apps/projects/app/components/Panel/ReviewApplication/ReviewApplication.js
```diff
@@ -158,7 +158,8 @@ const ReviewApplication = ({ issue, requestIndex }) => {
             {request.review.feedback.length ? request.review.feedback : 'No feedback was provided'}
           </Text.Block>
         </React.Fragment>
-      ) : (
+      )}
+      {!(readOnly || 'review' in request) && (
```
chadoh: I would find this a lot easier to read if we got rid of the parentheses and did something like:

```js
!readOnly && !request.review
```

apps/projects/app/components/Content/IssueDetail/EventsCard.js
```diff
@@ -53,13 +53,13 @@ IssueEvent.propTypes = {
 
 const applicationLink = (user, onReviewApplication, issue, index) => (
   <React.Fragment>
-    {user} submitted <Link onClick={() => onReviewApplication(issue, index)}>an application for review</Link>
+    {user} submitted <Link onClick={() => onReviewApplication(issue, index, true)}>an application for review</Link>
```
chadoh: I find it really hard to understand what a function will do with a raw boolean like this, and would prefer that we passed it with a name. We can do that like this:

```js
onReviewApplication({ issue, index, readOnly: true })
```

apps/projects/app/components/Panel/ReviewWork/ReviewWork.js
```diff
@@ -155,7 +156,8 @@ const ReviewWork = ({ issue }) => {
             </Text.Block>
           </FieldText>
         </React.Fragment>
-      ) : (
+      )}
+      {!(readOnly || 'review' in work) && (
```
chadoh: Same comment here about making boolean logic readable

apps/projects/app/components/Panel/ReviewApplication/ReviewApplication.js
rkzel: Fixed!

apps/projects/app/components/Panel/ReviewWork/ReviewWork.js
rkzel: Fixed as well. Much more readable now.

apps/projects/app/components/Content/IssueDetail/EventsCard.js
rkzel: Changed in all places for both panels. Actually easier than it seemed.

PR #1757, opened 2019-12-10
Issue Card icons changed

rkzel:
First few changes for the Issue Card (displayed in Issues tab). Icons changed, color added to IconInfo...
Screen before:
Screenshot from 2019-12-13 15-49-48
Screen after:
Screenshot from 2019-12-13 16-02-07


apps/projects/app/components/Card/Issue.js
```diff
@@ -45,6 +56,52 @@ FlexibleDiv.propTypes = {
   children: PropTypes.node.isRequired,
 }
 
+const Bounty = ({ issue }) => {
+  const theme = useTheme()
+  const { layoutName } = useLayout()
+  const { workStatus, deadline, expLevel, openSubmission } = issue
+
+  let status = BOUNTY_STATUS[workStatus]
+  if (openSubmission && workStatus === 'funded') status = 'Accepting work submissions'
```
chadoh: I like that all the other status phrases are co-located in the `BOUNTY_STATUS` object. Can we move this string there, as well, and do something here like:

```js
if (openSubmission && workStatus === 'funded') status = BOUNTY_STATUS['open-submission']

File: apps/projects/app/components/Card/Issue.js [hunk 2]
```diff
@@ -45,6 +56,52 @@ FlexibleDiv.propTypes = {
...
+          {status}
+        </span>
+        {dot}
+        <span css="white-space: nowrap">
+          <div css={`
+              display: inline-block;
+              vertical-align: bottom;
+              margin-right: 6px;
+              margin-bottom: -8px;
+            `}>
```
chadoh: Why does `IconClock` require a special `div` wrapper, but the other icons do not?

rkzel: Apparently it doesn't. It was there before, but the relevant bit is only that -8px margin. I made it more standard (that component's local standard).

apps/projects/app/components/Card/Issue.js
rkzel: Sure. `'open-submission-funded'`.

Issue #1731, opened 2019-12-04
Improve New Allocation panel


chadoh:
See [Figma frame D 1.3](https://www.figma.com/file/9c4GAhAa1LhH1TmJC8vCHA/Allocations?node-id=878%3A26926) for current designs. Screenshot:

![design mockup](https://user-images.githubusercontent.com/221614/70544097-84018780-1b39-11ea-96f4-2d30cffe048a.png)

* Update styling of the labels (Look at the NewBudget panel, those are the correct styles)
* Update how we articulate "Available Funds" – instead of current two-liner, conditionally display Available Funds as either 1. the funds remaining in the budget or 2. as the funds available in the vault depending on the following:
  * If Available in Vault < Available in Budget, Available Funds = Available in Vault
    “Available funds: x” (where x is the amount available in vault)
  * If Available in Vault > Available in Budget, Available Funds = Available in Budget
    “Available funds: x” (where x is the amount available in budget)

Text for help component next to ‘Available funds’ displays the following text:

> **Available funds** displays the amount of funds remaining within this allocation’s budget. If this amount exceeds the organization’s vault funds, an accordingly adjusted value will be shown here.

javieralaves:
@chadoh You accidentally created this issue on the product repo :) I transferred it to open-enterprise, but you have to reassign it to the epic. I’m not able to do this on my phone right now

javieralaves:
Will be able to test once fix to #1642 is deployed.

javieralaves:
@chadoh @PeterMPhillips Not sure this is 100% good to go. Have a look at what the help component displays to me today:
![Screenshot 2019-12-20 at 07.57.58.png](https://images.zenhubusercontent.com/5cf4fe30c7859b3c51c81634/049e1389-9e4c-4ab1-b661-d771a52a5b0e)
December 17th was three days ago, and as far as I can tell this DAO has a manually defined period duration of a day, so if anything it should probably say 'until 20 December 2019'.

stellarmagnet:
@javieralaves why do you think it has a period duration of a day?

javieralaves:
@stellarmagnet Have a look [at the link](https://rinkeby.aragon.org/#/neudune/0x66cad2ce53968d08461cfb0b71568d66f2985d4f/) and have a look at the metrics panel, or click on the `New budget` button. You'll notice that the duration is 1 day.

javieralaves:
[...]

Issue #1642, opened 2019-11-14
Integrate filters bar and new buttons set up


javieralaves:
As shown [in frame D 1.0 – Main here](https://www.figma.com/file/9c4GAhAa1LhH1TmJC8vCHA/Allocations?node-id=878%3A20278).

![screenshot](https://images.zenhubusercontent.com/5699129a638e718c99601ac2/c48af4ab-8981-4da3-ad5e-7d32ea97c77b)

javieralaves:
Testing this on the continuous deploy from December 17th 2019.

✅ Filters work as expected
✅ Buttons are placed as expected
❌ Screen goes blank when I click on 'New allocation'

Waiting for a fix on the 'New allocation' button in order to mark this one off. @chadoh @e18r please let me know when changes have been merged to re-review this issue.

javieralaves:
@chadoh Were you able to merge a fix for this?

chadoh:
@javieralaves yes! Fixed by #1794, and it looks like Travis successfully deployed since then.

javieralaves:
Reviewing this in the continuous deploy [Neudune DAO](https://rinkeby.aragon.org/#/neudune/0x66cad2ce53968d08461cfb0b71568d66f2985d4f/) on December 20th, 2019. Looks good!

Issue #1573, opened 2019-11-07
Improve the style of the budget card


stellarmagnet:
* Fix the location of the bar so they are always at the same location on all budget cards, regardless of the length of the budget name
* Change the location of the "per period"
* Increase the width of the utilization bar
* Add margin around budget cards on narrow screens (see #1702 for screenshots)

Design:

[See Figma frame D 1.0 for latest](https://www.figma.com/file/9c4GAhAa1LhH1TmJC8vCHA/Allocations?node-id=964%3A27)

![screenshot](https://images.zenhubusercontent.com/5699129a638e718c99601ac2/cbd242d8-bcc2-4123-9902-bc43b914d7c8)

jayalaves:
Just updated the design, you can find the latest version [available here](https://www.figma.com/file/9c4GAhAa1LhH1TmJC8vCHA/Allocations?node-id=192%3A2032).

jayalaves:
Design available in the Figma master branch [here](https://www.figma.com/file/9c4GAhAa1LhH1TmJC8vCHA/Allocations?node-id=878%3A20278).

javieralaves:
Looking good!

javieralaves:
Tested this in the continuous deploy from December 17th, 2019.

✅ Looks as expected.

PR #1758, opened 2019-12-10
Allocations: add Budget Detail page

### LINKED ISSUE #1575: Implement budget detail page
Whenever you click on a budget card, you should be able to view the details of the budget and it's allocation history.

WIP design:
https://www.figma.com/file/9c4GAhAa1LhH1TmJC8vCHA/Allocations?node-id=798%3A6377

chadoh:
Fixes #1575 

A lot changed to support this. If you have questions about specific changes, you can check individual commits & commit messages, which are quite detailed. 

There's a bit in `Card/Budget.js` where indentation changed. To hide this noise, view the diff [with indentation changes hidden](https://github.com/AutarkLabs/open-enterprise/pull/1758/files?w=1).

budget detail

### DIFF
--- a/apps/allocations/app/components/App/App.js
+++ b/apps/allocations/app/components/App/App.js
@@ -1,10 +1,18 @@
 import React from 'react'
-import { useAragonApi } from '../../api-react'
+import { useAragonApi, usePath } from '../../api-react'
 import { Main, SidePanel, SyncIndicator } from '@aragon/ui'
 
 import { IdentityProvider } from '../LocalIdentityBadge/IdentityManager'
-import { Overview } from '.'
+import { BudgetDetail, Overview } from '.'
 import { usePanel } from '../../context/Panel'
+
+const BUDGETS_REGEX = new RegExp('^/budgets/')
+
+function Routes() {
+  const [path] = usePath()
+  if (path.match(BUDGETS_REGEX)) return <BudgetDetail />
+  return <Overview />
+}
 
 const App = () => {
   const { api, appState } = useAragonApi()
@@ -32,7 +40,7 @@ const App = () => {
         onResolve={handleResolveLocalIdentity}
         onShowLocalIdentityModal={handleShowLocalIdentityModal}
       >
-        <Overview />
+        <Routes />
         <SyncIndicator visible={isSyncing} />
         <SidePanel
           title={(panel && panel.data.heading) || ''}
--- a/apps/allocations/app/components/App/BudgetDetail.js
+++ b/apps/allocations/app/components/App/BudgetDetail.js
@@ -23,6 +23,12 @@ const ID_REGEX = new RegExp('^/budgets/(?<id>[0-9]+)')
 const percentOf = (smaller, bigger) =>
   `${BigNumber(100 * smaller / bigger).dp(1).toString()}%`
 
+const formatDate = date =>
+  new Intl.DateTimeFormat(undefined, {
+    dateStyle: 'long',
+    timeStyle: 'long',
+  }).format(date)
+
 function CurrencyValue({ amount, context, token }) {
   const theme = useTheme()
 
@@ -117,7 +123,7 @@ export default function BudgetDetail() {
   }
 
   const allocations = (appState.allocations || []).filter(a => a.accountId === id)
-
+  const period = appState.period
   const utilized = budget.amount - budget.remaining
 
   return (
@@ -144,7 +150,7 @@ export default function BudgetDetail() {
           <CurrencyValue
             amount={budget.amount}
             token={budget.token}
-            context="per 4 weeks"
+            context="per 30 days"
           />
         </InfoBlock>
         <InfoBlock title="Utilized">
@@ -165,6 +171,14 @@ export default function BudgetDetail() {
       <Box heading="Budget info">
         <InfoBlock title="Budget ID">
           #{budget.id}
+        </InfoBlock>
+      </Box>
+      <Box heading="Period info">
+        <InfoBlock title="Start Date">
+          {formatDate(period.startDate)}
+        </InfoBlock>
+        <InfoBlock title="End Date">
+          {formatDate(period.endDate)}
         </InfoBlock>
       </Box>
       { !!allocations.length && <AllocationsHistory allocations={allocations} /> }
--- a/apps/allocations/app/components/App/BudgetDetail.js
+++ b/apps/allocations/app/components/App/BudgetDetail.js
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useState } from 'react'
 import PropTypes from 'prop-types'
 import BigNumber from 'bignumber.js'
 import { useAragonApi, usePath } from '../../api-react'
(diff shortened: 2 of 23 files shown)


chadoh:
@javieralaves @stellarmagnet I've requested reviews from both of you because I'd love if you could check out the attached gif and see if you can spot any style issues before this gets merged.

chadoh:
Thanks, @javieralaves! 

- Would it be possible to make the whole card clickable instead of just the title? _**This will be done as part of #1573**_
- Budget title text should be set to content style (hex: 212B36) _**You mean the title in the card? This will also be part of #1573**_
- [x] Context menu item icons should should be set to surfaceIcon style (hex: 8FA4B5) _**Done**_
- [x] Remove the 'Budget' column in the detail view table _**Done**_
- Change the single check icon for a double check icon as shown in the designs (Emilio must have already uploaded the svg file since it's also being used in Rewards) and set the text color to positive (hex: 2CC68F) _**This strikes me as a separate task from any current issues that we have, so I created #1763 for it)**_
- The styling of the 'New allocation' panel input container labels has to be updated. Look at how you've designed the 'Reactivate budget' panel, those are the correct styles :) _**I made this part of #1731**_

I've updated the GIF in the description to show the two improvements. Do you think it looks good and that it makes sense to work these other items in other PRs?

chadoh:
@javieralaves I was mistaken about this one: 

* Would it be possible to make the whole card clickable instead of just the title? 

I've updated it to work like this:

clickable cards

This is a big change from the design, because I removed the context menu. The context menu does not work, when the entire card is clickable. It also provides little value on this Overview page, and less value once #1642 is implemented.

I've freestyled the hover styles here. Let me know what you think of all of this.

apps/allocations/app/components/BudgetContextMenu.js
```diff
@@ -0,0 +1,58 @@
...
+  }, [])
+
+  return (
+    <ContextMenu>
+      {!path.match('^/budgets') && budget.active && (
+        <ContextMenuItem onClick={() => newAllocation(budget.id)}>
+          <IconPlus color={theme.surfaceIcon} />
+          <ActionLabel>New allocation</ActionLabel>
+        </ContextMenuItem>
+      )}
```
PeterMPhillips: I think we can just remove this since we no longer have the context menu in the Overview page

apps/allocations/app/components/App/App.js
```diff
@@ -32,7 +35,8 @@ const App = () => {
         onResolve={handleResolveLocalIdentity}
         onShowLocalIdentityModal={handleShowLocalIdentityModal}
       >
-        <Overview />
+        {path === '/' && <Overview />}
```
PeterMPhillips:
```suggestion
        {!path.match(BUDGETS_REGEX) && <Overview />}
```
Until we implement 404 pages, we should at least fail gracefully into the Overview page when a user submits an incorrect url. Right now if I type '/overview' into the url (or anything else that does not match the regex), I get a blank page.

apps/allocations/app/components/App/BudgetDetail.js
```diff
@@ -0,0 +1,242 @@
...
+  const period = usePeriod()
+  const theme = useTheme()
+
+  const { id } = parsePath('/budgets/:id')
+
+  const budget = appState.budgets.find(b => b.id === id)
+  if (!budget) {
+    patientlyRequestPath('/')
+    return null
+  }
```
PeterMPhillips: This should be handled before loading the BudgetDetails page. Say a user is loading the link for the first time, they enter '/budgets/1' into the url, but `appState.budgets` hasn't synced yet. In this code, `if (!budget)` will return true, and the url will get reset to '/', so the user will be brought to the overview page, even if '/budgets/1' exists. If you don't call `requestPath('/')` then the url won't be reset and the page will load properly once `appState.budgets` has synced.

apps/allocations/app/hooks/usePathHelpers.js
```diff
@@ -0,0 +1,47 @@
...
+  // if this page is the first page loaded for the user,
+  // we may still have incomplete state from aragonAPI;
+  // let's give it a second before redirecting
+  const patientlyRequestPath = React.useCallback(path => {
+    if (firstTry) {
+      setTimeout(() => setFirstTry(false), 1000)
+    } else {
+      requestPath(path)
+    }
+  }, [ path, requestPath ])
```
PeterMPhillips: I cannot get this to work properly. If I put the wrong ID into the url, I'm left with a blank page. If I log `firstTry` it always returns `true`. I think `usePathHelpers` must be getting reloaded and the state keeps getting reset.

apps/allocations/app/components/App/App.js
PeterMPhillips: In a lower comment I offered up another solution to this issue, so you can ignore this.

stellarmagnet:
I don't necessarily agree with removing the context menu from the budget card. The court has cards + context menu coming up. We should ask Pierre if they intend to enable this functionality (cards being clickable + context menu).

image

ottodevs:
> The context menu does not work, when the entire card is clickable. It also provides little value on this Overview page, and less value once #1642 is implemented.

This seems not quite accurate, a simple `stopPropagation` in the click handler for the context menu should be enough to fix that situation, [here is some reference in case you ask](https://stackoverflow.com/questions/38619981/react-prevent-event-bubbling-in-nested-components-on-click)

As a user I really think that saving one click is a good UX design, in this case removing the context menu from the overview is adding an extra click - at least - so I am not fully satisfied about your decision Chad, anyway the designers should have the last word about this, just adding my technical take on this, and my opinion as potential user, in case that makes sense.

javieralaves:
> This seems not quite accurate, a simple `stopPropagation` in the click handler for the context menu should be enough to fix that situation, here is some reference in case you ask

@ottodevs Does this mean that it's possible to have both a clickable card and a context menu?

That said – As @chadoh points out, once #1642 is implemented, most day to day interactions will be readily available with one or two clicks; regardless of whether there is a context menu or not. I guess this would no longer hold true if we were to add some form of additional functionality that a context menu could provide room for, so sticking to the pattern won't hurt for now.

What I wanted to originally avoid (and still want to avoid) is having the card title be styled with the `link` color.

apps/allocations/app/hooks/usePathHelpers.js
chadoh: Ah, thanks! This broke when I started using `useCallback` with the definition of this function. We need to redefine the function when the `firstTry` variable changes. Fixed locally; pushing soon.

apps/allocations/app/components/App/BudgetDetail.js
chadoh: The `patiently` part means that it waits one second for appState to get set up properly before redirecting. In my experience (and it sounds like in Kevin's), this one second delay is sufficient for getting the app state loaded correctly, even when loading this link for the first time. See definition of `patientlyRequestPath` for details on how it works.

apps/allocations/app/components/App/App.js
chadoh: I'm going to go with something closer to this solution, I think 😄 

Your second suggestion seems very specific to the current routing structure of Allocations, and I want to try to come up with patterns that will be useful in more apps and as we add new routes to Allocations.

apps/allocations/app/components/BudgetContextMenu.js
chadoh: Great catch!

PR #1759, opened 2019-12-10
Allocations: Deactivate budget

PeterMPhillips:
This PR allows a user to deactivate a budget and reactivate it again. Additionally, there are changes to the radspec to account for this new format. Finally, to keep everything aligned with the figma, I changed the appearance of the Budget cards to display 'Active' or 'Inactive' depending on the state of the budget.
AllocationsDeactivate


apps/allocations/app/components/App/Budgets.js
```diff
@@ -70,24 +62,22 @@ const Budgets = () => {
     setPanel({
       content: NewBudget,
       data: {
-        heading: 'Edit budget',
+        heading: editingBudget.amount > 0 ? 'Edit budget' : 'Reactivate budget',
```
chadoh: Rather than our entire frontend needing to know that "`amount === 0` means `inactive`", I would rather we added a line to our `script.js` logic to set `budget.active` to `true` or `false`. Then this line can become:

```suggestion
        heading: editingBudget.active ? 'Edit budget' : 'Reactivate budget',
```

apps/allocations/app/components/Card/Budget.js
```diff
@@ -151,6 +173,20 @@ const StyledCard = styled(Card)`
   width: auto;
 `
 
+const CardBottom = styled.div`
+  padding: 4px 12px;
+  width: 100%;
+  display: flex;
+  justify-content: space-between;
+  position: absolute;
```
chadoh: Why do we need this? `position: absolute` can cause weird visual bugs, and I'd like to avoid it if we can!

apps/allocations/app/components/Card/Budget.js
```diff
@@ -37,37 +37,20 @@ const Budget = ({
...
-  }
 
   return (
     <Wrapper
       name={name}
       theme={theme}
+      amount={amount}
+      symbol={token.symbol}
+      inactive={inactive}
       menu={
```
chadoh: I think we only pass this `menu` prop in once, now. Maybe it makes sense to get rid of the prop and define it inline. Agree?

PeterMPhillips: I'm hesitant to do this as we'd have to pass 3 functions as props instead of 1 node. Is that what you want?

apps/allocations/app/components/Card/Budget.js
PeterMPhillips: This positions the period info and active/inactive label at the bottom of the card. There are other ways of doing this, but I believe they'll take more lines, and rearrange other divs. Maybe a better place to adjust this would be on this PR: https://github.com/AutarkLabs/open-enterprise/pull/1761, since I will have to update that one to account for any visual changes on this PR?

apps/allocations/app/components/App/Budgets.js
PeterMPhillips: This has been updated in the latest commit

apps/allocations/app/components/Card/Budget.js
chadoh: Ah, nevermind, no need to bother with this right now. Thanks!

apps/allocations/app/components/Card/Budget.js
```diff
@@ -76,32 +59,34 @@ const Budget = ({
...
           <ContextMenuItem onClick={edit}>
             <IconEdit />
-            <ActionLabel>Edit</ActionLabel>
+            <ActionLabel>{active ? 'Edit' : 'Reactivate'}</ActionLabel>
           </ContextMenuItem>
-          {/* <ContextMenuItem onClick={deactivate}> */}
-          {/*   <IconProhibited /> */}
-          {/*   <ActionLabel>Deactivate</ActionLabel> */}
-          {/* </ContextMenuItem> */}
+          {amount > 0 && (
```
chadoh: We can change this to `active &&` now

chadoh: I amended your most recent commit with this change ✅

PR #1765, opened 2019-12-13
Allocations: New Budget panel

PeterMPhillips:
~When the `budget-detail` branch (PR #1758) gets merged I'll redirect this to `dev`~

Small visual changes to the panel:
Screenshot from 2019-12-12 20-44-17

And here it shows weeks if the number of days is a multiple of 7:
Screenshot from 2019-12-12 20-44-57

### DIFF
--- a/apps/allocations/app/components/Panel/NewBudget.js
+++ b/apps/allocations/app/components/Panel/NewBudget.js
@@ -4,7 +4,9 @@ import { useAragonApi } from '../../api-react'
 import {
   DropDown,
   Field,
+  GU,
   Info,
+  Text,
   TextInput,
   useTheme
 } from '@aragon/ui'
@@ -16,6 +18,9 @@ import { BigNumber } from 'bignumber.js'
 import { ETH_DECIMALS, MIN_AMOUNT } from '../../utils/constants'
 import CurrencyBox from '../Form/Field/CurrencyBox'
 import useSaveBudget from '../../hooks/useSaveBudget'
+
+const MS_PER_DAY = 24 * 60 * 60 * 1000
+const DEFAULT_DURATION = 30 * MS_PER_DAY
 
 // TODO:: This should be votingTokens from account?
 const INITIAL_STATE = {
@@ -95,11 +100,15 @@ class NewBudget extends React.Component {
         disabled={name === '' || amount === '' || nameError || amountError}
         errors={
           <ErrorContainer>
-            { this.props.editingBudget.id && (
+            { this.props.editingBudget.id ? (
               <Info>
                 Please keep in mind that any changes to the budget amount may only be effectuated upon the starting date of the next accounting period.
               </Info>
-            ) }
+            ) : (
+              <Info>
+                Budgets represent a spending limit that is imposed upon an allocation category for the accounting period. Budgets do not hold any actual funds.
+              </Info>
+            )}
           </ErrorContainer>
         }
       >
@@ -139,7 +148,7 @@ class NewBudget extends React.Component {
                 onChange={this.handleSelectToken}
               />
             )}
-
+            <Duration />
           </InputGroup>
         </Field>
       </Form>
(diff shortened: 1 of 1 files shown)


apps/allocations/app/components/Panel/NewBudget.js
```diff
@@ -139,14 +146,33 @@ class NewBudget extends React.Component {
...
+  const days = Math.round(ms/86400000)
+  if(days%7 === 0){
+    const weeks = days/7
+    return `${weeks === 1 ? 'week' : `${weeks} weeks`}`
+  }
+  return `${days === 1 ? 'day' : `${days} days`}`
+}
+
 const NewBudgetWrap = props => {
   const { appState: { tokens = [] } } = useAragonApi()
```
chadoh: Let's move the `duration` calculation here, please!

apps/allocations/app/components/App/Overview.js
```diff
@@ -11,14 +11,21 @@ import useSaveBudget from '../../hooks/useSaveBudget'
 
 const Overview = () => {
   const { appState } = useAragonApi()
-  const { allocations = [], budgets = [], isSyncing = true } = appState
+  const { allocations = [], budgets = [], period, isSyncing = true } = appState
   const { setPanel } = usePanel()
   const saveBudget = useSaveBudget()
 
+  //Set duration to default, then calculate if 'period' is set
+  let duration = 2592000000
```
chadoh: We can move this outside of the component. And this number is incomprehensible to me; I'd love to see it shown as a calculation. 

```js
const DEFAULT_DURATION = 30 * 24 * 60 * 60 * 1000

File: apps/allocations/app/components/Panel/NewBudget.js [hunk 3]
```diff
@@ -139,14 +146,33 @@ class NewBudget extends React.Component {
                 onChange={this.handleSelectToken}
               />
             )}
-
+            <Text
+              css={`
+                white-space: nowrap;
+                line-height: 40px;
+                padding-left: 16px;
```
chadoh: We should probably use `padding-left: ${2 * GU}px` here

apps/allocations/app/components/Panel/NewBudget.js
```diff
@@ -139,14 +146,33 @@ class NewBudget extends React.Component {
...
+            </Text>
           </InputGroup>
         </Field>
       </Form>
     )
   }
 }
 
+const timeString = ms => {
+  const days = Math.round(ms/86400000)
```
chadoh: I'd prefer we expand this large number into its constituent pieces, too – we could set up a constant, if you want. `ms / MS_PER_DAY`, where `const MS_PER_DAY = 24 * 3600 * 1000` is set at the top of the file

apps/allocations/app/components/Panel/NewBudget.js
```diff
@@ -139,14 +146,33 @@ class NewBudget extends React.Component {
...
+  if(days%7 === 0){
+    const weeks = days/7
+    return `${weeks === 1 ? 'week' : `${weeks} weeks`}`
+  }
+  return `${days === 1 ? 'day' : `${days} days`}`
```
chadoh:
```js
return days === 1 ? 'day' : `${days} days`
```

Seems cleaner to get rid of the wrapping backticks/interpolation

PeterMPhillips:
@chadoh I turned the timeString() function into a component called Duration. It calls the useAragonApi and useTheme hooks so they don't have to passed as props. Let me know if that works for you, happy to amend it if its not to your liking

apps/allocations/app/components/Panel/NewBudget.js
```diff
@@ -139,29 +148,53 @@ class NewBudget extends React.Component {
...
+      `}
+    >
+      every {string}
+    </Text>
   )
 }
 
+const NewBudgetWrap = props => {
+  const { appState: { tokens = [] } } = useAragonApi()
+  return <NewBudget tokens={tokens} {...props} />
```
chadoh: Why stop passing in `saveBudget` and `theme`? This seems like a bug.

PeterMPhillips: I think that `saveBudget` must have been removed when I rebased. But I believe it gets passed in `{...props}', no?

`theme` was only being used by a `Text` component that has been moved to `Duration` so `NewBudget` doesn't need it anymore

chadoh: `saveBudget` was being defined in `NewBudgetWrap` and passed into `NewBudget`. I think you'll find that saving budgets no longer works.

PR #1767, opened 2019-12-13
Allocations period details

PeterMPhillips:
Covers this issue: https://github.com/autarklabs/open-enterprise/issues/1576

There are some changes that affect other components. In particular, I moved the `usePeriod` function from `BudgetDetail` to the hooks folder and it's now being used by `NewBudget` and `PeriodDetails`

Currently, only verified tokens are used to calculate the budget. As you can see in this screen `AUTD` doesn't contribute to the 'Budgeted' field or the 'Budget utilized' field.

Screenshot from 2019-12-18 17-11-12


ottodevs:
I retargeted base branch to `dev`, because `allocations-newbudgetpanel` PR was already merged

apps/allocations/app/components/App/PeriodDetails.js
```diff
@@ -0,0 +1,138 @@
...
+  const days = Math.round(ms/MS_PER_DAY)
+  if(days%7 === 0){
+    const weeks = days/7
+    return [ String(weeks), `week${weeks === 1 ? '' : 's'}` ]
+  }
+  return [ String(days), `day${days === 1 ? '' : 's'}` ]
+}
+
+const getBudgetTotals = (budgets, rates) => {
+  if(budgets.length && rates){
```
chadoh: Let's change this to an early return so we don't have to indent so much

```js
if (!budgets.length || !rates) {
  return { amount: BigNumber(0), remaining: BigNumber(0) }
}

const convertedBudgets = budgets.map(...)

return convertedBudgets.reduce(...)
```

apps/allocations/app/components/App/PeriodDetails.js
```diff
@@ -0,0 +1,138 @@
...
+    const convertedBudgets = budgets.map(budget => {
+      const rate = rates[budget.token.symbol]
+      return {
+        amount: rate ? (new BigNumber(budget.amount)).div(rate).div(10**budget.token.decimals) : new BigNumber(0),
+        remaining: rate ? (new BigNumber(budget.remaining)).div(rate).div(10**budget.token.decimals) : new BigNumber(0)
```
chadoh: Hmm, if we have rates but not rates for _this_ token, then we'll show some pretty disappointing statistics

apps/allocations/app/components/App/PeriodDetails.js
```diff
@@ -0,0 +1,138 @@
...
+        .map(({ token }) => token.symbol)
+      const res = await fetch(convertApiUrl(symbols))
+      const convertRates = await res.json()
+      if (JSON.stringify(rates) !== JSON.stringify(convertRates)) {
+        setRates(convertRates)
+      }
+    }
+  }, CONVERT_THROTTLE_TIME)
+
+  updateConvertedRates()
```
chadoh: On every render of this component, `updateConvertedRates` will get re-defined and re-called. I don't think we want that. Instead, we can make use of the `useEffect` hook like this:

```js
const [ ratesFetchedAt, setRatesFetchedAt ] = useState(new Date())

useEffect(() => {
  const interval = setInterval(() => setRatesFetchedAt(new Date()), CONVERT_INTERVAL)
  return () => clearInterval(interval)
}, [])

useEffect(() => {
  async function updateConvertedRates() {
    ...
  }
  updateConvertedRates()
}, [ ratesFetchedAt ])
```

The second argument to the first `useEffect` here is `[]` – this means that the effect will be run _once_, when the component first renders, and not again. The return value of the same effect, `() => clearInterval(interval)`, will be run when the component unmounts.

The second effect will be run when the component first mounts and then every time the `ratesFetchedAt` variable changes.

apps/allocations/app/components/App/PeriodDetails.js
```diff
@@ -0,0 +1,138 @@
...
+import { Box, GU, Text, textStyle, useTheme } from '@aragon/ui'
+import throttle from 'lodash.throttle'
+import BigNumber from 'bignumber.js'
+import usePeriod from '../../hooks/usePeriod'
+import { ETHER_TOKEN_FAKE_ADDRESS, isTokenVerified } from '../../../../../shared/utils/token-utils'
+
+const MS_PER_DAY = 24 * 60 * 60 * 1000
+const MONTHS = [ 'Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec' ]
+const CONVERT_API_BASE = 'https://min-api.cryptocompare.com/data'
+const CONVERT_THROTTLE_TIME = 5000
```
chadoh: Every five seconds seems really frequent. Could we set this to once every, I don't know, minute?

shared/utils/token-utils.js
```diff
@@ -1,9 +1,9 @@
 import { ETHER_TOKEN_VERIFIED_ADDRESSES } from './verified-tokens'
 import { toUtf8 } from './web3-utils'
-import tokenSymbolAbi from '../json-abis/token-symbol.json'
-import tokenSymbolBytesAbi from '../json-abis/token-symbol-bytes.json'
-import tokenNameAbi from '../json-abis/token-name.json'
-import tokenNameBytesAbi from '../json-abis/token-name-bytes.json'
+import tokenSymbolAbi from '../abi/token-symbol.json'
+import tokenSymbolBytesAbi from '../abi/token-symbol-bytes.json'
+import tokenNameAbi from '../abi/token-name.json'
+import tokenNameBytesAbi from '../abi/token-name-bytes.json'
```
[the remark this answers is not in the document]
PeterMPhillips: ¯\_(ツ)_/¯ I guess none of the other apps are using this? Rewards has it's own version.

But it's needed for the `isTokenVerified` function

apps/allocations/app/components/App/PeriodDetails.js
PeterMPhillips: I'm not sure what to do here. Either we show a bunch of values denominated in their native tokens or we convert the values to USD. But if those tokens have no market value, then there is nothing to show.

chadoh: Yeah, I think this is the best we can do

apps/allocations/app/components/App/PeriodDetails.js
```diff
@@ -0,0 +1,138 @@
...
+          title="Time remaining"
+          large={remainingArray[0]}
+          small={remainingArray[1]}
+        />
+      </div>
+    </Box>
+  )
+}
+
+const InfoBlock = ({ title, large, small }) => {
```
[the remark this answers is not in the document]
PeterMPhillips: I was unable to get the `css` prop to work, it would just come out as `undefined` when values were passed to it. I used the `style` prop instead

chadoh:
See my implementation. Gotta pass `className`
  

  
  

  
  
>

apps/allocations/app/components/App/PeriodDetails.js
```diff
@@ -1,16 +1,16 @@
...
-import { Box, GU, Text, textStyle, useTheme } from '@aragon/ui'
-import throttle from 'lodash.throttle'
+import { Box, GU } from '@aragon/ui'
 import BigNumber from 'bignumber.js'
+import InfoBlock from './InfoBlock'
 import usePeriod from '../../hooks/usePeriod'
 import { ETHER_TOKEN_FAKE_ADDRESS, isTokenVerified } from '../../../../../shared/utils/token-utils'
 
 const MS_PER_DAY = 24 * 60 * 60 * 1000
 const MONTHS = [ 'Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec' ]
```
chadoh: We're no longer using `MONTHS`, right?

PeterMPhillips: good catch

PR #1782, opened 2019-12-16
Allocations/routing: wait, don't redirect

chadoh:
Since it can take quite some time from initial load of an app until the time that it may have enough data loaded to show a certain URL, we do not want to change the route if no matching item is found. Instead, we want to ignore the unknown URL.

This sets us up to do that by making the BudgetDetail component ignore routing, instead passing `budget` in as a prop. This allows all routing logic to be located in the Routes component in App.js.

I also standardized on `usePathHelpers`, rather than sometimes using raw `usePath`.


chadoh:
Ah, great point, @PeterMPhillips! I removed it

PR #1788, opened 2019-12-17
bug fix: Status component used in Review panels updated

rkzel:
So it can be used in both panels without significant changes.
Status (rejected/approved) is stored in different variables -
WorkSubmissions have variable called 'accepted', the same
variable in Applications is called 'approved'.

### DIFF
--- a/apps/projects/app/components/Panel/PanelComponents.js
+++ b/apps/projects/app/components/Panel/PanelComponents.js
@@ -71,9 +71,9 @@ const statuses = [
   { color: 'positive', text: 'Accepted' },
 ]
 
-export const Status = ({ review }) => {
+export const Status = ({ reviewDate, approved }) => {
   const theme = useTheme()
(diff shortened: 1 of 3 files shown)


apps/projects/app/components/Panel/PanelComponents.js
```diff
@@ -71,26 +71,24 @@ const statuses = [
...
     </div>
   )
 }
 Status.propTypes = {
-  review: PropTypes.shape({
-    approved: PropTypes.bool.isRequired,
-    reviewDate: PropTypes.string.isRequired,
-  }).isRequired,
+  reviewDate: PropTypes.string.isRequired,
+  approved: PropTypes.bool.isRequired,
```
chadoh: Why switch the order of these to be non-alphabetical?

rkzel: Fixed.

PR #1800, opened 2019-12-18
1789 multibounties funding

rkzel:
This adds "Number of bounties" field:

Screenshot from 2019-12-18 19-12-14

It is then saved on IPFS and makes an u-turn ending up in store, available in issue object.


ottodevs:
Just to clarify here @rkzel : What is the status of this PR? is still draft? Will it be part of the decoupled projects efforts?

PR #1801, opened 2019-12-18
feat: informative dot votes

PeterMPhillips:
The PR adds informative dot voting to the Dot Voting app. It covers everything in this epic: https://github.com/AutarkLabs/product/issues/48

Just a note, in the gif it shows the error message: 'Options may not be identical' but it has been updated to 'The same option cannot be specified twice.'

InformativeDotVote

### DIFF
--- a/apps/dot-voting/app/App.js
+++ b/apps/dot-voting/app/App.js
@@ -1,9 +1,10 @@
 import React, { useCallback, useEffect, useState } from 'react'
 import PropTypes from 'prop-types'
-import { EmptyStateCard, GU, Header, LoadingRing, Main, SyncIndicator } from '@aragon/ui'
+import { Button, EmptyStateCard, GU, Header, IconPlus, LoadingRing, Main, SidePanel, SyncIndicator } from '@aragon/ui'
 import { useAragonApi } from './api-react'
 import { IdentityProvider } from './components/LocalIdentityBadge/IdentityManager'
 import { AppLogicProvider, useAppLogic } from './app-logic'
+import { NewVote } from './components/Panels'
 import Decisions from './Decisions'
 import emptyStatePng from './assets/voting-empty-state.png'
 
@@ -37,7 +38,7 @@ const useVoteCloseWatcher = () => {
   }, [ votes, voteTime ])
 }
 
-const Empty = ({ isSyncing }) => (
+const Empty = ({ isSyncing, onClick }) => (
   <div
     css={`
       position: fixed;
@@ -63,8 +64,9 @@ const Empty = ({ isSyncing }) => (
             <span>Syncing…</span>
           </div>
         ) : (
-          'After you create an allocation or issue curation, you can vote here.'
+          'No dot votes here'
         )}
+      action={<Button label="New dot vote" onClick={onClick} />}
       illustration={illustration}
     />
   </div>
@@ -72,12 +74,16 @@ const Empty = ({ isSyncing }) => (
 
 Empty.propTypes = {
   isSyncing: PropTypes.bool.isRequired,
+  onClick: PropTypes.func.isRequired,
 }
 
 const App = () => {
   useVoteCloseWatcher()
 
   const { api } = useAragonApi()
+  const [ panelOpen, setPanelOpen ] = useState(false)
+  const newVote = () => setPanelOpen(true)
+  const closePanel = () => setPanelOpen(false)
 
   const handleResolveLocalIdentity = useCallback(address => {
     return api.resolveAddressIdentity(address).toPromise()
@@ -89,18 +95,33 @@ const App = () => {
       .toPromise()
   }, [api])
 
-  const { isSyncing, votes, voteTime, pctBase } = useAppLogic()
-
-  if (!votes.length) return <Empty isSyncing={isSyncing}/>
+  const { isSyncing, votes } = useAppLogic()
 
   return (
     <IdentityProvider
       onResolve={handleResolveLocalIdentity}
-      onShowLocalIdentityModal={handleShowLocalIdentityModal}
-    >
-      <Header primary="Dot Voting" />
-      <Decisions />
-      <SyncIndicator visible={isSyncing} />
+      onShowLocalIdentityModal={handleShowLocalIdentityModal}>
+      {!votes.length ? (
+        <Empty isSyncing={isSyncing} onClick={newVote} />
+      ) : (
+        <>
+          <Header
+            primary="Dot Voting"
+            secondary={
+              <Button
+                mode="strong"
+                icon={<IconPlus />}
+                onClick={newVote}
+                label="New dot vote"
+              />
+            } />
+          <Decisions/>
+          <SyncIndicator visible={isSyncing} />
+        </>
+      )}
+      <SidePanel title='New dot vote' opened={panelOpen} onClose={closePanel}>
+        <NewVote onClose={closePanel} />
+      </SidePanel>
     </IdentityProvider>
   )
 }
--- a/apps/dot-voting/app/Decisions.js
+++ b/apps/dot-voting/app/Decisions.js
@@ -31,7 +31,8 @@ const OUTCOME_FILTER_REJECTED = 2
 const OUTCOME_FILTER_ENACTED = 3
 const OUTCOME_FILTER_PENDING = 4
 const APP_FILTER_ALLOCATIONS = 1
-const APP_FILTER_PROJECTS = 2
+const APP_FILTER_DOT = 2
+const APP_FILTER_PROJECTS = 3
 
 const useFilterVotes = (votes, voteTime) => {
   const { appState: { globalMinQuorum = 0 } } = useAragonApi()
@@ -67,6 +68,9 @@ const useFilterVotes = (votes, voteTime) => {
       if (appFilter !== NULL_FILTER_STATE) {
         if (appFilter === APP_FILTER_ALLOCATIONS &&
           type !== 'allocation'
+        ) return false
+        if (appFilter === APP_FILTER_DOT &&
+          type !== 'information'
         ) return false
         if (appFilter === APP_FILTER_PROJECTS &&
           type !== 'curation'
@@ -217,7 +221,6 @@ const Decisions = () => {
                 'Open',
                 'Closed',
               ]}
-              width="128px"
             />
             {voteStatusFilter !== STATUS_FILTER_OPEN && (
               <DropDown
@@ -226,7 +229,6 @@ const Decisions = () => {
                 selected={voteOutcomeFilter}
                 onChange={handleVoteOutcomeFilterChange}
                 items={[ 'All', 'Passed', 'Rejected', 'Enacted', 'Pending' ]}
-                width="128px"
               />
             )}
             <DropDown
@@ -234,8 +236,7 @@ const Decisions = () => {
               placeholder="App"
               selected={voteAppFilter}
               onChange={handleVoteAppFilterChange}
-              items={[ 'All', 'Allocations', 'Projects' ]}
-              width="128px"
+              items={[ 'All', 'Allocations', 'Dot Voting', 'Projects' ]}
             />
           </div>
         </Bar>
--- a/apps/dot-voting/app/components/VoteStatus.js
+++ b/apps/dot-voting/app/components/VoteStatus.js
@@ -22,7 +22,7 @@ const VoteStatus = ({ vote }) => {
       color: theme.textSecondary,
     },
     [VOTE_STATUS_SUCCESSFUL]: {
-      label: 'Passed (pending)',
+      label: `Passed ${vote.data.executable ? '(pending)' : ''}`,
       Icon: IconCheck,
       color: theme.positive,
     },
--- a/apps/dot-voting/app/store/votes.js
+++ b/apps/dot-voting/app/store/votes.js
@@ -2,7 +2,7 @@ import { combineLatest, from } from 'rxjs'
 import { first, map, mergeMap } from 'rxjs/operators'
 
 import { app } from './'
-import { EMPTY_CALLSCRIPT } from '../utils/vote-utils'
+import { EMPTY_CALLSCRIPT, FUNC_TYPES } from '../utils/vote-utils'
 import { ETHER_TOKEN_FAKE_ADDRESS, tokenAbi } from '../../../../shared/lib/token-utils'
 import { loadTokenSymbol } from '../../../../shared/store-utils/token'
 import { addressesEqual } from '../../../../shared/lib/web3-utils'
@@ -62,12 +62,13 @@ const loadVoteData = async (voteId, settings) => {
       .pipe(first())
       .subscribe(async voteData => {
         let funcSig = voteData.executionScript.slice(58, 66)
-        if (funcSig == 'b3670f9e') {
-          resolve(loadVoteDataProjects(voteData, voteId))
+        if (FUNC_TYPES[funcSig]) {
+          resolve(loadVoteDataInformative(voteData, voteId, FUNC_TYPES[funcSig]))
         } else {
           resolve(loadVoteDataAllocation(voteData, voteId, settings))
         }
       })
+
   })
 }
 
@@ -122,12 +123,13 @@ const loadVoteDataAllocation = async (vote, voteId, settings) => {
           tokenSymbol: symbol,
           metadata: vote.voteDescription,
           type: 'allocation',
+          executable: true,
         })
       })
   })
 }
 
-const loadVoteDataProjects = async (vote, voteId) => {
+const loadVoteDataInformative = async (vote, voteId, type) => {
   return new Promise(resolve => {
     combineLatest(
       app.call('getCandidateLength', voteId),
@@ -145,7 +147,8 @@ const loadVoteDataProjects = async (vote, voteId) => {
         resolve({
           ...marshallVote(decoratedVote),
           metadata: vote.voteDescription,
-          type: 'curation',
+          type,
+          executable: false,
           canExecute,
           options,
         })
--- a/shared/ui/components/form/Form.js
+++ b/shared/ui/components/form/Form.js
@@ -1,0 +1,62 @@
+import PropTypes from 'prop-types'
+import React from 'react'
+import styled from 'styled-components'
+import { Button } from '@aragon/ui'
+
+const Form = ({
+  children,
+  className,
+  onSubmit,
+  submitText,
+  disabled,
+  errors,
+}) => {
+  return (
+    <form
+      className={className}
+      onSubmit={(e) => {
+        e.preventDefault()
+        onSubmit()
+      }}
+    >
+      <VerticalSpace />
+      {children}
+      <Button
+        type="submit"
+        style={{ marginTop: '24px' }}
+        mode="strong"
+        wide
+        disabled={disabled}
+      >
+        {submitText}
+      </Button>
+      <ErrorBlock>
+        {errors}
+      </ErrorBlock>
+    </form>
+  )
+}
+
+const ErrorBlock = styled.div`
+  margin-top: 24px;
+`
+
+const VerticalSpace = styled.div`
+  height: 24px;
+`
+
+Form.propTypes = {
+  children: PropTypes.node.isRequired,
+  className: PropTypes.string,
+  onSubmit: PropTypes.func.isRequired,
+  submitText: PropTypes.string.isRequired,
+  disabled: PropTypes.bool,
+  errors: PropTypes.node,
+}
+
+Form.defaultProps = {
+  submitText: 'Submit',
+  disabled: false,
+}
+
+export default Form
--- a/shared/ui/components/form/inputs/OptionsInput.js
+++ b/shared/ui/components/form/inputs/OptionsInput.js
@@ -1,98 +1,84 @@
 import PropTypes from 'prop-types'
 import React from 'react'
 import styled from 'styled-components'
-import { Button, IconRemove, TextInput, theme, unselectable } from '@aragon/ui'
+import { Button, IconPlus, IconRemove, TextInput, useTheme, unselectable } from '@aragon/ui'
 
 const OptionsInput = ({
-  input,
-  name,
+  options,
   onChange,
-  placeholder = '',
-  validator,
-  values,
 }) => {
-  const validated = !input || validator(values, input.addr)
+  const theme = useTheme()
+
+  const changeOption = (value, id) => {
+    onChange({
+      target: { name: 'optionsChange', value, id },
+    })
+  }
 
   const addOption = () => {
     onChange({
-      target: validated
-        ? { name, value: [ ...values, input ] }
-        : { name: 'addressError', value: true }, // enable error msg if needed
+      target: {
+        name: 'optionsAdd',
+        value: true,
+      }
     })
-    cleanInputBox()
   }
 
-  const removeOption = option => {
-    // perform the change on the parent by using onChange prop without modifying value prop
-    option &&
-      onChange({ target: { name, value: values.filter(v => v !== option) } })
-  }
-
-  const cleanInputBox = () => {
-    onChange({ target: { name: 'userInput', value: { addr: '' } } })
-  }
-
-  const onChangeInput = ({ target: { value } }) => {
+  const removeOption = id => {
     onChange({
-      target: { name: 'userInput', value: { addr: value } },
+      target: {
+        name: 'optionsRemove',
+        id,
+      }
     })
   }
-
-  const loadOptions = values.map((option, i) => (
-    <StyledOption key={i}>
-      <StyledInput readOnly value={option.addr} />
-      <IconContainer
-        style={{ transform: 'scale(.8)' }}
-        onClick={() => removeOption(option)}
-        title="Click to remove the option"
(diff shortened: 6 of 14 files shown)


apps/dot-voting/app/App.js
```diff
@@ -63,21 +64,26 @@ const Empty = ({ isSyncing }) => (
...
+  onOpen: PropTypes.func.isRequired,
 }
 
 const App = () => {
   useVoteCloseWatcher()
 
   const { api } = useAragonApi()
+  const [ opened, setOpened ] = useState(false)
+  const onOpen = () => setOpened(true)
+  const onClose = () => setOpened(false)
```
chadoh: I find this name awkward. What is opened? Part of me thinks we should immediately reach for the `context/Panel` pattern [from Allocations](https://github.com/AutarkLabs/open-enterprise/blob/e3e861a6acb621e99e9a91a1fccf8ff18047d8aa/apps/allocations/app/context/Panel.js), but that's probably overkill for now. How about:

```js
const [ addingVote, setAddingVote ] = useState(false)
const newVote = () => setAddingVote(true)
const closePanel = () => setAddingVote(false)
```

apps/dot-voting/app/App.js
```diff
@@ -89,18 +95,34 @@ const App = () => {
...
 
   return (
-    <IdentityProvider
-      onResolve={handleResolveLocalIdentity}
-      onShowLocalIdentityModal={handleShowLocalIdentityModal}>
-      <Header primary="Dot Voting" />
-      <Decisions />
-      <SyncIndicator visible={isSyncing} />
-    </IdentityProvider>
+    <Main assetsUrl={ASSETS_URL}>
```
chadoh: `Main` is already wrapping everything down in the `DotVoting` component. We don't need it twice!

I think we could put this all inside the `IdentityProvider` component instead, though:

```jsx
return (
  <IdentityProvider
  >
    {!votes.length ... }
    <SidePanel ... />
  </IdentityProvider>
)


)

File: apps/dot-voting/app/Decisions.js [hunk 3]
```diff
@@ -226,16 +230,16 @@ const Decisions = () => {
...
             )}
             <DropDown
               header="App"
               placeholder="App"
               selected={voteAppFilter}
               onChange={handleVoteAppFilterChange}
-              items={[ 'All', 'Allocations', 'Projects' ]}
-              width="128px"
+              items={[ 'All', 'Allocations', 'Dot Voting', 'Projects' ]}
+              width="140px"
```
chadoh: I vote we get rid of these hard-coded widths entirely while we're here. They give us very little other than maintenance cost. Here's what it looks like without them:

Screen Shot 2019-12-18 at 16 27 15

@javieralaves do you think that looks ok? For comparison, here they all are when hard-coded to `140px` width, which is as Peter has it now:

Screen Shot 2019-12-18 at 16 28 46

apps/dot-voting/app/components/Panels/NewVotePanelContent.js
```diff
@@ -0,0 +1,123 @@
...
+      {children}
+    </div>
+  )
+}
+
+ErrorMessage.propTypes = {
+  children: PropTypes.node,
+}
+
+const NewVotePanelContent = ({ onClose }) => {
```
chadoh: I see this name is following the naming convention set by the first panel, but I find it a needlessly wordy convention. I say we change the file+component name to just `NewVote`. Do you agree?

apps/dot-voting/app/components/Panels/NewVotePanelContent.js
```diff
@@ -0,0 +1,123 @@
...
+  const { api } = useAragonApi()
+  const [ description, setDescription ] = useState()
+  const [ options, setOptions ] = useState({ '0':'' })
+  const [ optionsId, setOptionsId ] = useState(0)
+  const [ submitDisabled, setSubmitDisabled ] = useState(true)
+  const [ errors, setErrors ] = useState(null)
+
+  useEffect(() => {
+    if(!description) return setSubmitDisabled(true)
+    if(Object.values(options).filter(o => o !== '').length === 0) return setSubmitDisabled(true)
```
chadoh:
```js
if (Object.values(options).some(v => !v)) return setSubmitDisabled(true)
```

I'd also take `.some(v => v === '')`

apps/dot-voting/app/components/Panels/NewVotePanelContent.js
```diff
@@ -0,0 +1,123 @@
...
+    if(Object.values(options).length !== new Set(Object.values(options)).size){
+      setErrors(<ErrorMessage>The same option cannot be specified twice.</ErrorMessage>)
+      return setSubmitDisabled(true)
+    } else {
```
chadoh: Since you have the `return` in each of the `if` blocks, you can get rid of the `else` here. eslint has a rule for this called [`no-else-return`](https://eslint.org/docs/rules/no-else-return), where they explain the concept well.

apps/dot-voting/app/components/Panels/NewVotePanelContent.js
```diff
@@ -0,0 +1,123 @@
...
+      setErrors(null)
+      return setSubmitDisabled(false)
+    }
+  }, [ description, options ])
```
chadoh: It _might_ read more clearly to use two separate `useEffect` calls, one for `description` and the other for `options`. What do you think?

apps/dot-voting/app/components/Panels/NewVotePanelContent.js
```diff
@@ -0,0 +1,123 @@
...
+
+  const updateDescription = e => setDescription(e.target.value)
```
chadoh: Given that this function will be redefined on every render anyhow, I don't think it gets us much. I'd rather we defined this as an inline anonymous function:

```js
onChange={e => setDescription(e.target.value)}
```

apps/dot-voting/app/components/Panels/NewVotePanelContent.js
```diff
@@ -0,0 +1,123 @@
...
+
+  const updateOptions = e => {
+    if(e.target.name === 'optionsAdd'){
+      const newOptions = { ...options }
+      const newId = optionsId + 1
+      newOptions[newId] = ''
+      setOptionsId(newId)
```
chadoh: We can do this without `optionsId`:

```js
const newOptions = { ...options }
newOptions[options.length] = ''

File: apps/dot-voting/app/components/Panels/NewVotePanelContent.js [hunk 10]
```diff
@@ -0,0 +1,123 @@
...
+    if(e.target.name === 'optionsRemove'){
+      const newOptions = { ...options }
+      delete newOptions[e.target.id]
+      setOptions(newOptions)
+    }
+  }
+
+  const createVote = async () => {
+    const currentApp = await api.currentApp().toPromise()
+    const callscript = encodeInformationVote(currentApp.appAddress, description, Object.values(options).filter(o => o !== ''))
```
chadoh: Let's break up this long line

apps/dot-voting/app/components/VoteStatus.js
```diff
@@ -22,7 +22,7 @@ const VoteStatus = ({ vote }) => {
       color: theme.textSecondary,
     },
     [VOTE_STATUS_SUCCESSFUL]: {
-      label: 'Passed (pending)',
+      label: `Passed ${vote.data.executable ? '(pending)' : ''}`,
```
chadoh, line 25: Do we have design for this? I find the difference between "passed (pending)" and "passed (enacted)" pretty confusing already – when I first used the app, I was quite surprised that manual action was needed to "enact" a vote; I thought it would be enacted automatically once it passed. Getting rid of the "(pending)" until a vote is _executable_ seems like it would add even more to potential confusion. How long will it take for a vote to be executable?

apps/dot-voting/app/store/votes.js
```diff
@@ -62,12 +62,13 @@ const loadVoteData = async (voteId, settings) => {
       .pipe(first())
       .subscribe(async voteData => {
         let funcSig = voteData.executionScript.slice(58, 66)
-        if (funcSig == 'b3670f9e') {
-          resolve(loadVoteDataProjects(voteData, voteId))
+        if (funcSig == '00000000' || funcSig == 'b3670f9e') {
```
chadoh: Please always use `===`

But also, don't we want to use `FUNC_TYPES` here? 

```js
if (FUNC_TYPES[funcSig]) {

File: shared/ui/components/form/Form.js [hunk 13]
```diff
@@ -0,0 +1,54 @@
...
+}) => {
+  return (
+    <React.Fragment>
+      <VerticalSpace />
+      {children}
+      <Button
+        style={{ userSelect: 'none', marginTop: '24px' }}
+        mode="strong"
+        wide
+        onClick={onSubmit}
```
chadoh: Why not use an actual `<form>` element with an `onSubmit` handler? That also lets us pass `className` through so the `Form` can be wrapped with `styled` (or equivalently, so we can use the `css=` prop with it).

```js
const Form = ({
   children,
   onSubmit,
   submitText,
   disabled,
   errors,
 }) => {
   return (
    <form onSubmit={onSubmit}>
      ...
      <Button type="submit" ...>

File: apps/dot-voting/app/components/VoteStatus.js (hunk 11 above)
Comment (PeterMPhillips) [reply apps/dot-voting/app/components/VoteStatus.js:25]: Some vote are not executable such as informative votes from Dot Voting and curation votes from Projects. Once they're passed, that's it. No need to display '(pending)' or '(enacted)'

File: apps/dot-voting/app/store/votes.js (hunk 12 above)
Comment (PeterMPhillips) [reply apps/dot-voting/app/store/votes.js]: oh that's nice

File: apps/dot-voting/app/App.js (hunk 2 above)
Comment (PeterMPhillips) [reply apps/dot-voting/app/App.js]: ah, this is because of the rebase I did from dev. I didn't catch the change

File: apps/dot-voting/app/components/Panels/NewVotePanelContent.js (hunk 7 above)
Comment (PeterMPhillips) [reply apps/dot-voting/app/components/Panels/NewVotePanelContent.js]: I'd rather keep them together to ensure we return `setSubmitDisabled` to `false` only when all conditions are met. If we separate them, we would have an issue where fixing `description` could enable the submit button even when the `options` don't validate

File: shared/ui/components/form/Form.js (hunk 13 above)
Comment (PeterMPhillips) [reply shared/ui/components/form/Form.js]: Then the submit function has to have e.preventDefault(). However, I can use a `div` instead of `React.Fragment` and then pass the className to it.

File: apps/dot-voting/app/components/Panels/NewVotePanelContent.js (hunk 4 above)
Comment (PeterMPhillips) [reply apps/dot-voting/app/components/Panels/NewVotePanelContent.js]: completely. I had originally written it as NewVote and then changed it to match the convention

File: apps/dot-voting/app/components/Panels/NewVotePanelContent.js (hunk 9 above)
Comment (PeterMPhillips) [reply apps/dot-voting/app/components/Panels/NewVotePanelContent.js]: Unfortunately we can't. Doing it this way would produce unexpected behaviour if a user removes any option except for the last one, creating a new option will overwrite the last option in the list.

File: apps/dot-voting/app/App.js (hunk 1 above)
Comment (PeterMPhillips) [reply apps/dot-voting/app/App.js]: I agree that the naming can be improved but I feel like `addingVote` doesn't accurately represent what the bool is doing. I changed it to `const [ panelOpen, setPanelOpen ] = useState(false)`, does that look alright to you?

Comment (chadoh) [reply apps/dot-voting/app/App.js]: Yes, looks good

File: apps/dot-voting/app/components/Panels/NewVotePanelContent.js (hunk 7 above)
Comment (chadoh) [reply apps/dot-voting/app/components/Panels/NewVotePanelContent.js]: Makes sense!

File: shared/ui/components/form/Form.js (hunk 13 above)
Comment (chadoh) [reply shared/ui/components/form/Form.js]: I think having an `e.preventDefault()` is worth it. This allows normal form behavior, such as pressing "enter" to submit while editing a form field, instead of clicking the button.

File: apps/dot-voting/app/Decisions.js (hunk 3 above)
Comment (javieralaves) [reply apps/dot-voting/app/Decisions.js]: Agree with your decision @chadoh. Makes things easier for me too. What then becomes the rule for how wide dropdowns should be?

File: shared/ui/components/form/Form.js (hunk 13 above)
Comment (PeterMPhillips) [reply shared/ui/components/form/Form.js]: I wasn't thinking that I could have the `preventDefault` inside the form's `onSubmit` function like so:
```
    <form
      className={className}
      onSubmit={(e) => {
        e.preventDefault()
        onSubmit()
      }}
   >
```
This saves having to include it in the functions that are passed to this form

File: apps/dot-voting/app/App.js [hunk 14]
```diff
@@ -89,18 +95,34 @@ const App = () => {
...
 
   return (
-    <IdentityProvider
-      onResolve={handleResolveLocalIdentity}
-      onShowLocalIdentityModal={handleShowLocalIdentityModal}>
-      <Header primary="Dot Voting" />
-      <Decisions />
-      <SyncIndicator visible={isSyncing} />
-    </IdentityProvider>
+    <>
```
chadoh: I think putting `Empty` inside the `IdentityProvider` is probably a good thing?

shared/ui/components/form/Form.js
```diff
@@ -0,0 +1,63 @@
...
+      }}
+    >
+      <VerticalSpace />
+      {children}
+      <Button
+        type="submit"
+        style={{ userSelect: 'none', marginTop: '24px' }}
+        mode="strong"
+        wide
+        onClick={onSubmit}
```
chadoh: Now that we have the `onSubmit` on the form, we do not need the `onClick` on this button

shared/ui/components/form/inputs/OptionsInput.js
```diff
@@ -1,98 +1,83 @@
...
-
-  const removeOption = option => {
-    // perform the change on the parent by using onChange prop without modifying value prop
-    option &&
-      onChange({ target: { name, value: values.filter(v => v !== option) } })
   }
 
-  const cleanInputBox = () => {
-    onChange({ target: { name: 'userInput', value: { addr: '' } } })
+  const addOption = () => {
```
chadoh, line 18: I think it would be great if we could focus the most recently added field, rather than requiring the user to click or tab backwards twice to focus it themselves. That might be out of scope for now, though. I'm having trouble thinking of an easy and error-free way of doing it.

PeterMPhillips: I did this by setting `autofocus` to true for every input except the first:
```
<StyledOption key={id}>
              <TextInput
                value={options[id]}
                onChange={v => changeOption(v.target.value, id)}
                wide
                autofocus={index !== 0}
              />
```
However, the only issue is when a user removes one of the options, the focus resets to the first option in the list. I'm not sure why that is the case. But it seems like such a small issue, I feel like its not worth putting the time in to solving it right now

shared/ui/components/form/inputs/OptionsInput.js
```diff
@@ -1,98 +1,86 @@
...
       <StyledButton
-        disabled={!validated}
         compact
-        mode="secondary"
+        mode="normal"
         onClick={addOption}
-        children={'+ Add option'}
-        title={validated ? 'Click to add' : ''}
-      />
+        title={'Click to add'}
```
chadoh: No need for the `{}` here. `title="Click to add"`

shared/ui/components/form/Form.js
```diff
@@ -0,0 +1,62 @@
...
+      onSubmit={(e) => {
+        e.preventDefault()
+        onSubmit()
+      }}
+    >
+      <VerticalSpace />
+      {children}
+      <Button
+        type="submit"
+        style={{ userSelect: 'none', marginTop: '24px' }}
```
chadoh: ugh, I don't know why we add `user-select: none` all over the place; I often get frustrated by UIs that don't let me select text

ottodevs: Well @chadoh why would you want to select a Button label? Even if I agree with you about webs abusing `user-select: none`, for these cases of UI elements, I don't think is that bad, users normally want to select the content but not the interface elements text.

I also don't normally see we use `user-select: none` a lot (I actually checked it and we just use in 4 places, - actually 3 because of repeated component, and another is unnecessary, because is the same case as here, a `Button` already having the style - )

We should difference between content and UI elements, the latter rarely provide any interesting information for the user more than the needed to interact with the element, as opposed to the content type, that it is what normally the user would want to select/copy 

Said this, there is no need to add this style as the [`@aragon/ui` `Button`](https://ui.aragon.org/button/) already [has it ](https://github.com/aragon/aragon-ui/blob/38bda15486985ce9fa0959ddb09ca7b076982e50/src/components/Button/Button.js#L250) applied, furthermore, if we would need to use the style for whatever reason, the correct way is to use the `unselectable` helper from `@aragon/ui` @PeterMPhillips

chadoh: To clarify, I don't think aragonUI should set `user-select: none` almost ever, maybe not at all

I don't think we should ask "why would a user ever want to select such-and-such text?" Instead, I think we should ask "why should we prevent them from selecting it?" Who knows what users will do or why they would want to do it! Trying to guess at all possible motives for any given action they might take seems futile. Maybe they want to select it so they can _ask us a question about it_, and that is valid and we are preventing it. But even if I couldn't come up with such a reason, that doesn't prove that no valid reason exists.

Clicking buttons already works properly. I don't see any value in limiting the possible uses of the text in a button so that the occassional user who tries to select the text will get frustrated.

chadoh: haha, I was trying to select the "comment" button text and sent my comment before I meant to! The "comment" button here does not have `user-select: none`, but I still can't select its text.

Anyhow, y'all are seeing my annoyance at having run into problems with various apps, Aragon apps and others, which have prevented me from selecting text. I just can't figure out why this is a trend right now. I guess maybe `user-select` is still new enough that people are still over-using it.

Anyhow, yeah, aragonUI already adds this annoying rule, so there's no need for us to add it unless we want to override it back to `user-select: auto`.

ottodevs:
@PeterMPhillips please remember to move to "Review" pipeline, it is quite weird to see an "Approved" PR that is "In progress" pipeline.

shared/ui/components/form/Form.js
ottodevs: > To clarify, I don't think aragonUI should set `user-select: none` almost ever, maybe not at all
> 
> I don't think we should ask "why would a user ever want to select such-and-such text?" Instead, I think we should ask "why should we prevent them from selecting it?" Who knows what users will do or why they would want to do it! Trying to guess at all possible motives for any given action they might take seems futile. 

The same thing could apply to the question: "why don't we just publish a code terminal and avoid creating UIs at all? that way users would be able to freely access everything in the app... Yes, it sounds ridiculous, is not about letting the user select one thing or another, a enough experimented user (a developer, for example) will be able to select the text if he wants, it is more a cosmetic, conventional thing, try to select the text from buttons in github? oh, you cannot do that, also, can you do that on your desktop apps, or phone? Definitely not. Is also not needed, there is not problem with that, because nobody needs to select the text in a button, because there should not be content on it, is just part of the UI, you can have whatever opinion about it, but is something conventional since the invention of GUIs. The rule exists for a reason.

>Maybe they want to select it so they can _ask us a question about it_, and that is valid and we are preventing it. But even if I couldn't come up with such a reason, that doesn't prove that no valid reason exists.

All your comment seems highly opinionated, you base all your reasons only into your personal experience and preferences... well you can easily realize that the user can send an screenshot or just write "Cancel button" or "New Allocation button" is not that much effort, and nobody would probably complain about that

> Clicking buttons already works properly. I don't see any value in limiting the possible uses of the text in a button so that the occassional user who tries to select the text will get frustrated.

I don't understand why a user would be frustrated by using the app the way it was designed to be used, would you be frustrated by using a DAO app if you are limited to the tokens you have instead of having the full token supply?

It looks ugly when you try to click a button and because your cursor moves you end selecting half of the UI, and basically it is a way to differentiate context from interface elements, I still don't see the value on allowing selection in the buttons, for me it is a sign of a non-professional or lack of care in an app / website development

As a final note, I want to tell you that, I didn't even state my own opinion about the topic, I just based my answer in the conventions I observed in the development industry.

I totally respect your opinion about that, but I think as developers, we should follow the convention, that, I don't think is "a trend" as you pointed (it was probably there before you even knew what internet was).

Personally, it bothers me when I cannot select the text from a website, fortunately I know how to easily skip the restriction, but in more than 30 years using computers I cannot remember a single occasion when I felt frustration because not being able to select text from a button, really...

ottodevs: > haha, I was trying to select the "comment" button text and sent my comment before I meant to! The "comment" button here does not have `user-select: none`, but I still can't select its text.

Of course it has the rule, try to go to the "Computed" tab in your stylesheet inspector 😉

stellarmagnet: I agree with Otto that I don't think it's important for button text to be copied

stellarmagnet:
Couple minor nit-picks:

- The text should be center-aligned in the Add another and Submit buttons 
- There seems to be excessive space below the Add another button and above the Submit button

I'm not sure if these were previously resolved; I'm going off the animation in the original PR.

PeterMPhillips:
> Couple minor nit-picks:
> 
> * The text should be center-aligned in the Add another and Submit buttons
> * There seems to be excessive space below the Add another button and above the Submit button
> 
> I'm not sure if these were previously resolved; I'm going off the animation in the original PR.

@stellarmagnet As far as I can tell the spacing is the same as the New Allocations panel. I think the '+' makes the 'Add another' button look off-centre. This may just be on my machine though. I can update the spacing for the 'Submit' button but will need to know what space you'd like. Right now its 24px.

Here is an image of New Allocation:
Screenshot from 2020-01-02 09-56-59

And here is an image of New Dot Vote:
Screenshot from 2020-01-02 09-48-50

ottodevs:
> > Couple minor nit-picks:
> > 
> > * The text should be center-aligned in the Add another and Submit buttons
> > * There seems to be excessive space below the Add another button and above the Submit button
> > 
> > I'm not sure if these were previously resolved; I'm going off the animation in the original PR.
> 
> @stellarmagnet As far as I can tell the spacing is the same as the New Allocations panel. I think the '+' makes the 'Add another' button look off-centre. This may just be on my machine though. I can update the spacing for the 'Submit' button but will need to know what space you'd like. Right now its 24px.
> 
> Here is an image of New Allocation:
> Screenshot from 2020-01-02 09-56-59
> 
> And here is an image of New Dot Vote:
> Screenshot from 2020-01-02 09-48-50

For consistency, can we just use the same design that the client has? for example, in the onboarding panel, it shouldn't be too much work to adapt, and personally it looks much better with the icon than the raw + sign:

image

Is just this easy:

```js
          <Button
            icon={
              <IconPlus
                css={`
                  color: ${theme.accent};
                `}
              />
            }
            label="Add more"
            onClick={addMember}
          />
```

source: https://github.com/aragon/aragon/blob/c5382349bc201dbcb217c6ac00611035ad02b257/src/templates/kit/screens/TokensScreen.js#L314

PeterMPhillips:
> > > Couple minor nit-picks:
> > > 
> > > * The text should be center-aligned in the Add another and Submit buttons
> > > * There seems to be excessive space below the Add another button and above the Submit button
> > > 
> > > I'm not sure if these were previously resolved; I'm going off the animation in the original PR.
> > 
> > 
> > @stellarmagnet As far as I can tell the spacing is the same as the New Allocations panel. I think the '+' makes the 'Add another' button look off-centre. This may just be on my machine though. I can update the spacing for the 'Submit' button but will need to know what space you'd like. Right now its 24px.
> > Here is an image of New Allocation:
> > Screenshot from 2020-01-02 09-56-59
> > And here is an image of New Dot Vote:
> > Screenshot from 2020-01-02 09-48-50
> 
> For consistency, can we just use the same design that the client has? for example, in the onboarding panel, it shouldn't be too much work to adapt, and personally it looks much better with the icon than the raw + sign:
> 
> image
> 
> Is just this easy:
> 
> ```js
>           <Button
>             icon={
>               <IconPlus
>                 css={`
>                   color: ${theme.accent};
>                 `}
>               />
>             }
>             label="Add more"
>             onClick={addMember}
>           />
> ```
> 
> source: https://github.com/aragon/aragon/blob/c5382349bc201dbcb217c6ac00611035ad02b257/src/templates/kit/screens/TokensScreen.js#L314

I've changed both the OptionsInput (used by Dot Voting) and RecipientsInput (used by Allocations) to the button style described by @ottodevs:
Screenshot from 2020-01-06 12-25-19

@stellarmagnet @javieralaves how does this look to you?

stellarmagnet:
While I appreciate Otto’s suggestion, I think that is out of scope for this
ticket. If we are to make this change, we should make it on all of our UIs
(new allocation, new dot vote, new issue curation)

PeterMPhillips:
@stellarmagnet Aside from reverting back the changes. Is there anything else you want me to adjust on the `Add another` button?

stellarmagnet:
@PeterMPhillips - there are no other changes I see, I just think it's about maintaining consistency! But if it's easier, I'm not opposed to shipping as is, if it expedites the process, and we can create backlog tickets to update the other panels to your currently implemented design in C1.

PR #1812, opened 2019-12-20
prepare repo for publishing to npm

topocount:
This repo contains a bunch of changes to prepare us to publish to npm. 
- The first an biggest change is renaming `@tps` to `@autarklabs`, which is an org name we actually own in the npmjs registry.
- Second, I updated all the prepublishOnly scripts in the apps to compile the contracts and extract the abi into its own folder, like the A1 apps do. It doesn't appear that apps are used as a dependency for their frontends, so this is not included in the npm release process. Let me know if you have any questions about this.


apps/address-book/package.json
```diff
@@ -14,7 +15,7 @@
     "ganache-cli:test": "sh ../../node_modules/@aragon/test-helpers/ganache-cli.sh",
     "lint": "solium --dir ./contracts",
     "precommit": "lint-staged",
-    "prepublishOnly": "truffle compile",
+    "prepublishOnly": "truffle compile --all && npm run abi:extract -- --no-compile",
```
ottodevs, line 18: `--all` seems redundant here, we have been testing without the option in the about app and it just works fine.

ottodevs: for consistency we may want to call the compile script, as in the rest of apps, anyway not a big deal 
```suggestion
    "prepublishOnly": "npm run compile && npm run abi:extract -- --no-compile",
```

apps/discussions/package.json
```diff
@@ -34,9 +34,11 @@
     "eslint-plugin-standard": "^4.0.0",
     "prettier": "^1.17.1",
     "react-hot-loader": "^4.12.7",
-    "solidity-coverage": "0.5.11"
+    "solidity-coverage": "0.5.11",
+    "truffle-extract": "^1.2.1"
   },
   "scripts": {
+    "abi:extract": "truffle-extract --output abi/ --keys abi",
```
ottodevs, line 41: `prePublishOnly` script is missing here

shared/deployments/staging.md
```diff
@@ -11,7 +11,7 @@
 ℹ Transaction hash: 0x4231262285244a475668e8dda953917a64cb9367885c822ca02eea3cae69bac4
 ```
 
-- command: `./node_modules/.bin/lerna exec --scope="@tps/apps-address-book" --stream aragon apm publish patch -- --files dist/ --environment staging`
+- command: `./node_modules/.bin/lerna exec --scope="@autarklabs/apps-address-book" --stream aragon apm publish patch -- --files dist/ --environment staging`
```
ottodevs, line 14: This document is for historic purposes, so that is the command that was used for that specific deployment, but I guess is ok to change it here anyway

shared/deployments/staging.md
```diff
@@ -22,7 +22,7 @@
 ℹ Transaction hash: 0x0b4bef766abb49543810a242902fe9cea3d534184f8001ade47f341559982c2b
 ```
 
-- command: `./node_modules/.bin/lerna exec --scope="@tps/apps-address-book" --stream aragon apm publish major -- --files dist/ --environment staging`
+- command: `./node_modules/.bin/lerna exec --scope="@autarklabs/apps-address-book" --stream aragon apm publish major -- --files dist/ --environment staging`
```
ottodevs, line 25: same here and for the rest of replacements in this file, it is documenting deployments that already happened

PR #1813, opened 2019-12-20
fix: corrected Estimated Date in View Application Panel

rkzel:
Issue: if not set, Estimated Date shows up as "Invalid Date". it should be either '-' if not set of correct Date in "month day" format. Screen before:

Screen after:

Screenshot from 2019-12-20 15-36-07

### DIFF
--- a/apps/projects/app/components/Form/Input/DatePicker.js
+++ b/apps/projects/app/components/Form/Input/DatePicker.js
@@ -24,14 +24,14 @@ const Container = styled.div`
   min-width: 15em;
   margin: 0 auto;
   padding-top: 0.5em;
-  background: ${theme.contentBackground};
+  background: white;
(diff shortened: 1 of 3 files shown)


rkzel:
> No date is ever shown, even when I set one in my application for the bounty. Also, the DatePicker calendar is translucent. Is this part of the design?

No. This is because that component uses `theme` from older AragonUI and it can't be easily converted to `useTheme` hook (Date being old style class component). It seems the short-term "fix" for now can be adding white background. There is already a ticket for fixing shared components, I added a note about modernizing such code... 

Date not appearing - when I fund an issue and make it requier applications (without changing date), fill in application (setting `Estimated completion` date to a day or a week in the future), I get something like this:
Screenshot from 2019-12-20 23-57-47

PR #1820, opened 2019-12-20
fix: Work Panel was unable to show multiple submissions

rkzel:
As per issue 1818. It was added (transplanted) from very similar Application Review Panel:

Screenshot from 2019-12-20 19-10-30

And the second submission, still unreviewed:

Screenshot from 2019-12-20 21-12-19

### DIFF
--- a/apps/projects/app/components/Panel/ReviewWork/ReviewWork.js
+++ b/apps/projects/app/components/Panel/ReviewWork/ReviewWork.js
@@ -30,7 +30,7 @@ import {
 import workRatings from '../../../utils/work-ratings.js'
 import { DetailHyperText } from '../../../../../../shared/ui'
 
-const ReviewWork = ({ issue, readOnly }) => {
+const ReviewWork = ({ issue, submissionIndex, readOnly }) => {
   const githubCurrentUser = useGithubAuth()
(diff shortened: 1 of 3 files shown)


apps/projects/app/components/Panel/ReviewWork/ReviewWork.js
```diff
@@ -30,7 +30,7 @@ import {
 import workRatings from '../../../utils/work-ratings.js'
 import { DetailHyperText } from '../../../../../../shared/ui'
 
-const ReviewWork = ({ issue, readOnly }) => {
+const ReviewWork = ({ issue, submissionIndex, readOnly }) => {
```
topocount: `submissionIndex` doesn't match the inputs on  this line: https://github.com/AutarkLabs/open-enterprise/blob/fd1f8f01af801452a8b1b726fd0c3ac5438ba7a8/apps/projects/app/components/Content/IssueDetail/EventsCard.js#L64

Please fix so that the correct work submission is opened in the panel when a link is clicked

rkzel: Fixed!

PR #1829, opened 2019-12-21
fix: bounty status after deadline didn't revert to 'not-funded'

rkzel:
Additionally IssueDetails (Bounty widget) and Issue Card needed
to notice bounty's expiration and show appropriate information,
as per Figma design.


apps/projects/app/components/Content/IssueDetail/BountyCard.js
```diff
@@ -212,38 +228,56 @@ const BountyCard = ({ issue }) => {
...
     >
-      <div css={`display: flex; margin-bottom: ${2 * GU}px`}>
-        <BountyDot workStatus={issue.workStatus} />
-        <div css="display: flex; align-items: baseline">
-          <Text size="xxlarge">{issue.balance}</Text>
-          <Text color={`${theme.surfaceContentSecondary}`} css="margin-left: 2px">{issue.symbol}</Text>
-        </div>
-      </div>
+      <div css={`
+         > :not(:last-child) {
```
chadoh, line 232: These sorts of styles that we add to all children of a parent can be tough to work with – can we add this bottom-margin to `Row` instead? It seems like everything in here on which we want bottom-margin is a `Row`

Also, keep in mind the [adjacent sibling combinator](https://css-tricks.com/almanac/selectors/a/adjacent-sibling/) for things like this:

```jsx
Row = styled.div`
  display: flex;
  & + & {
    margin-top: ${GU}px;
  }
```

But be careful with this. Like using `> :not(:last-child)` on the parent, these sorts of styles can overly complicate style maintenance. 

I think the best option in this case is _probably_ to add the margin to `Row` and then override one specific Row, if we need to.

apps/projects/app/components/Content/IssueDetail/BountyCard.js
```diff
@@ -212,38 +228,56 @@ const BountyCard = ({ issue }) => {
...
+          margin-bottom: ${GU}px;
+        }
+      `}>
+
+        {!(pastDueDate(issue.deadline) && issue.workStatus !== 'fulfilled') && (
+          <div css={`display: flex; margin-bottom: ${2 * GU}px`}>
```
chadoh, line 238: Why not use `Row` here?

apps/projects/app/components/Content/IssueDetail/BountyCard.js
```diff
@@ -212,38 +228,56 @@ const BountyCard = ({ issue }) => {
...
-          <Text color={`${theme.surfaceContentSecondary}`} css="margin-left: 2px">{issue.symbol}</Text>
-        </div>
-      </div>
+      <div css={`
+         > :not(:last-child) {
+          margin-bottom: ${GU}px;
+        }
+      `}>
+
+        {!(pastDueDate(issue.deadline) && issue.workStatus !== 'fulfilled') && (
```
chadoh, line 237: * as written: "Not (past due and work status isn't 'fulfilled')"
* without parentheses: "Not past due and not work status not fulfilled"

So... not past due and work status _is_ fulfilled? Can we get rid of the parentheses and write it that way?

apps/projects/app/components/Content/IssueDetail/BountyCard.js
```diff
@@ -212,38 +228,56 @@ const BountyCard = ({ issue }) => {
...
+          </Row>
+        )}
 
-      <Row>
-        <IconFile color={`${theme.surfaceIcon}`} />
-        <BountyText>
-          <Submissions issue={issue} />
-        </BountyText>
-      </Row>
+        {!(pastDueDate(issue.deadline) && issue.workStatus !== 'fulfilled') && (
```
chadoh, line 265: Same comment about hard-to-understand logical statements here

apps/projects/app/components/Content/IssueDetail/BountyCard.js
rkzel: Multiple conditions in `return` seem to be bad anyway, so I've made it a variable:

`const isNotExpired = !pastDueDate(issue.deadline) || issue.workStatus === 'fulfilled'`

to make this possible:

`{isNotExpired && (
<Row>...
`

apps/projects/app/components/Content/IssueDetail/BountyCard.js
rkzel: Yes. And this second case disappeared now (because design).

apps/projects/app/components/Content/IssueDetail/BountyCard.js
rkzel: Indeed. Fixed!

apps/projects/app/components/Content/IssueDetail/BountyCard.js
rkzel: Fixed. Previous approach was caused by uncertainity which `Row` is going to be the last one. Current approach doesn't care, using `margin-top` instead.

rkzel:
There is more to this. Namely: what exactly should and can happen after a bounty expires.
What should happen: the bounty can be edited and deadline extended, possibly with changed parameters like funding and exp level. I have spent too much time getting old [updateBounty](https://github.com/AutarkLabs/open-enterprise/blob/839518763cc09cff7290ec28c95d2381e57396d8/apps/projects/contracts/Projects.sol#L533) function back into life, but that function does not emit appropriate event (so can't update frontend automatically) and it can only change limited set of details of a bounty, so it is of not much use to us as it is right now. That makes quarter of Funding Panel obsolete by the way - I'm happy to clean up.. The other possibility is to remove that bounty and create new funding. Two transactions, but it looks like a more viable way.

PR #1832, opened 2020-01-02
implement commit linting for conventional-commit rules

topocount:
These changes will enforce conventional commit rules in both github and client-side via husky.

Note: I encountered this bug when trying to use an exclamation point to denote a breaking change: https://github.com/conventional-changelog/commitlint/issues/884


chadoh (approved):
Let's try it!

Issue #542, opened 2019-03-13
Improve scrolling style


stellarmagnet:
**Steps to reproduce**
(This can only be tested if you have a lot of Github repos)

Use case A
1) Navigate to the Projects app
2) Click "New Project"

Observe the dual scroll bars (see photo)

Expected behavior: There should only be a single scroll bar and it should be in the outermost edge. It should only be visible while scrolling is actively taking place.

Use case B
1) Navigate to the Projects app, and connect a few repos via "New Project
2) Click on the Issues tab
Observe the scroll bars

Expected behavior: There should only be a single scroll bar and it should be in the outermost edge. It should only be visible while scrolling is actively taking place.

Use case A
![image](https://user-images.githubusercontent.com/2584493/54282478-613b4600-459c-11e9-81eb-26153671ff09.png)

Use case B

javieralaves:
Reviewed in the continuous deploy DAO. Looks good!

PR #1837, opened 2020-01-02
fix: single scrollbar on New Project panel

PeterMPhillips:
This PR covers issue #542. The New Project panel only shows one scrollbar, additionally I fixed the issue where an x-axis scrollbar would appear if a project label was too long. Now we do an ellipses if text overflows.

Screenshot from 2020-01-02 11-31-51


e18r (approved):
Code looks good
Tested the app
No console errors
No linting errors
LGTM :+1:

PR #1847, opened 2020-01-06
fix(tests): fix tests failing because running in parallel

ottodevs:
While previously configuring lerna to run in parallel, it was introduced
a regression that makes test script fail because sharing devchain data,
it was fixed by limiting concurrency to single thread


chadoh (approved):
Good catch, thanks!

PR #1849, opened 2020-01-07
Handle GitHub token revocation

### LINKED ISSUE #654: Clear Stale GitHub Token from cache

e18r:
Fixes #654 

When a GitHub request is made with a revoked token, the end user is informed and asked to sign in again.

Peek 2020-01-07 10-12


apps/projects/app/utils/apollo-client.js
```diff
@@ -1,16 +1,34 @@
...
+            scope: null,
+            token: null,
+          })
+        }
       }
-    },
-  })
+    })
+  }
+  return initApolloClient
```
chadoh: I think we can save all consumers of `useApolloClient` a line or two by doing something like this:

```js
export const useApolloClient = () => {
  const { api, appState } = useAragonApi()
  if (appState.github.token === null) return null
  return new ApolloClient({ ... })
}
```

Note that there's also no reason to pass `token` in as an argument anymore, now that this is a hook!

Do you think this would work? Is there a reason to make this return an `initApolloClient` function, rather than `null` or an actual client?

apps/projects/app/utils/apollo-client.js
```diff
@@ -1,16 +1,34 @@
...
+        }
+      },
+      onError: error => {
+        if (error.networkError && error.networkError.statusCode === 401) {
+          api.emitTrigger(GITHUB_TOKEN_REVOKED, {
+            status: STATUS.REVOKED,
+            scope: null,
+            token: null,
+          })
+        }
```
chadoh: Let's add an `else` with `console.error(error)`, to keep the functionality we had before and make sure these errors are still visible in the console

chadoh (changes requested):

e18r:
Hey @chadoh, this PR is ready for another round of review. Thanks!

e18r:
Rebased with `dev` and solved several conflicts.

apps/projects/app/components/Card/Revoked.js
```diff
@@ -0,0 +1,40 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { Button, EmptyStateCard, GU, Text, useTheme } from '@aragon/ui'
+import { EmptyWrapper } from '../Shared'
+import revoked from '../../assets/revoked.svg'
```
chadoh, line 5: Can we export a component, instead of raw SVG? Like we did for [`IconDoubleCheck` in Allocations](https://github.com/AutarkLabs/open-enterprise/pull/1853/files#diff-373c0e728dced2965861263f8f033827)

apps/projects/app/components/Card/Revoked.js
```diff
@@ -0,0 +1,40 @@
...
+
+const Illustration = () => <img src={revoked} alt="" />
+
+
```
chadoh, line 9: our linter doesn't complain about two line breaks in a row so I guess I have to 🙃

apps/projects/app/utils/apollo-client.js
```diff
@@ -1,16 +1,35 @@
...
-export const initApolloClient = token =>
-  new ApolloClient({
+export const useApolloClient = () => {
+  const {
+    api,
+    appState: {
+      github: { token } = { token: null }
+    }
+  } = useAragonApi()
+  if (token === null) return null
```
chadoh, line 13: This is all a bit weird to read. How about:

```js
const { api, appState: { github } } = useAragonApi()
if (!github || !github.token) return null
```

Then we'll have to use `github.token` elsewhere, but I think that's a good balance of readability here vs terseness there.

chadoh (changes requested):
This looks really nice! I see a couple more small improvements we could make, which would not require the "request changes" status, but I also see something else...

The IssueDetail page doesn't work! I haven't looked into why, but it works on `dev` and is broken on this branch.

PR #1851, opened 2020-01-08
Rewards: Removed token name wrap

### LINKED ISSUE #1823: New reward: Reference token name wraps
There is plenty of space for the token / token name to be on a single line. 
You see in the screenshot here how it wraps to two lines

image

e18r:
Fixes #1823

Screenshot from 2020-01-08 11-34-01

### DIFF
--- a/apps/rewards/app/components/Panel/NewReward/TokenSelectorInstance.js
+++ b/apps/rewards/app/components/Panel/NewReward/TokenSelectorInstance.js
@@ -40,7 +40,6 @@ const IconSpacer = styled.div`
 `
 
 const TokenName = styled.span`
-  max-width: 110px;
   margin-right: 10px;
   overflow: hidden;
   text-overflow: ellipsis;


apps/rewards/app/components/Panel/NewReward/TokenSelectorInstance.js
```diff
@@ -40,7 +40,6 @@ const IconSpacer = styled.div`
 `
 
 const TokenName = styled.span`
-  max-width: 110px;
```
ottodevs, line 43: I guess if `max-width` was used, it was a mechanism to control the size of the element, we need to check that wider elements don't break the container box

ottodevs (changes requested):
The change looks good, feel free to merge it once you confirm the `max-width` removal is not breaking the layout for wider elements.

apps/rewards/app/components/Panel/NewReward/TokenSelectorInstance.js
e18r: I tested that by using a very long token name, and it wraps at the size of the container (minus margin), which is what we want. I still have no idea what was the intent of using that here, but as far as I'm concerned, it's simply not needed.

Issue #1852, opened 2020-01-08
Allocations history: display pending and rejected allocations


e18r:
Currently, when having allocations in different states in Dot Voting,
![Screenshot from 2020-01-08 12-15-58](https://user-images.githubusercontent.com/16065447/72000684-d0f87c80-3211-11ea-9313-1ebd16e92607.png)

Only the ones that are enacted show up in Allocations History:
![Screenshot from 2020-01-08 12-16-10](https://user-images.githubusercontent.com/16065447/72000751-f7b6b300-3211-11ea-9f8e-51b09ec8133f.png)

Make every kind of allocation show up in Allocations History according to the design:
![Screenshot from 2020-01-08 12-26-46](https://user-images.githubusercontent.com/16065447/72000853-2e8cc900-3212-11ea-846b-9b09c5d43678.png)

This also implies changing the current possible Allocations History entry statuses from
![Screenshot from 2020-01-08 12-27-59](https://user-images.githubusercontent.com/16065447/72000916-567c2c80-3212-11ea-9f10-af6f436a3bed.png)

To Pending, Rejected and Enacted.

Also, you need to include a new icon for the Pending status.

javieralaves:
@e18r What specifications are missing design wise?

e18r:
None, @javieralaves. I just assigned you so you can determine in which sprint to include this.

stellarmagnet:
I believe we can’t display rejected allocations based on current contract limitations.

We can’t display pending ones until forwarder API gets merged.

Cc @Quazia

PR #1853, opened 2020-01-08
Allocations: use double check icon and add status colors

### LINKED ISSUE #1763: Allocations table: use double-check icon & colors
## Design:

(See [Figma frame D 1.0](https://www.figma.com/file/9c4GAhAa1LhH1TmJC8vCHA/Allocations?node-id=878%3A27441) for latest)

"Enacted" text color: ~`theme.accent`~ `theme.positive` (`#2CC68F`)

e18r:
Fixes #1763 

* Moved the double check icon from Rewards to a shared location
* Updated the import from Rewards and tested it
* Replaced the check icon with the double check icon in Allocations
* Added status colors
* Discovered a bug whereby only allocations with enacted status show up in Allocations History; created ticket #1852 to track it

Screenshot from 2020-01-08 12-16-10

Screenshot from 2020-01-08 12-31-55
(this last one was updated by manually changing the status)


apps/allocations/app/components/App/AllocationsHistory.js
```diff
@@ -18,6 +19,8 @@ import { STATUSES } from '../../utils/constants'
 import { displayCurrency } from '../../../../../shared/ui/helpers'
 import { addressesEqual } from '../../../../../shared/lib/web3-utils'
 
+const IconDoubleCheck = () => <img src={doubleCheck} alt="" />
```
chadoh: Let's export a usable component from the shared location, so the interface stays similar to the one used by aragonUI

Also, I just want to think aloud about whether `alt=""` is the correct option. I think it probably is, here, since it doesn't add anything to the experience if a screen reader reads something aloud like "double checkmark enacted" instead of just "enacted." Do you agree?

Note that I think in most places we are using inline `svg` tags, which seems better to me. That gives us more control over accessibility, rather than just the `alt` text ([read more](https://www.sitepoint.com/tips-accessible-svg/)), and it allows us to override styles such as the icon color easily using CSS.

apps/allocations/app/components/App/AllocationsHistory.js
```diff
@@ -160,8 +163,11 @@ const RecipientAmount = styled.div`
  `
 
 const StatusContent = styled.div`
-  color: ${({ code, theme }) => code === 0 ?
-    theme.contentSecondary : theme.content};
+  color: ${({ code, theme }) => code === 0
+    ? theme.contentSecondary
+    : code === 1
```
chadoh: Yikes, this `code` argument is really hard to understand – it's _probably_ out-of-scope to improve this now, but I would love to see this variable or its values get more-understandable names.

apps/rewards/app/utils/helpers.js
```diff
@@ -4,7 +4,8 @@ import {
   ETH_DECIMALS_NUMBER,
   RECURRING_DIVIDEND,
 } from './constants'
-import doubleCheck from '../assets/double-check.svg'
+import doubleCheck
+  from '../../../../shared/ui/components/assets/svg/double-check.svg'
```
chadoh: Hmm, I see why we stuck with exporting the raw `svg` rather than a component – it's what we were already doing in Rewards. It looks like this is the only file that uses it, though, so it would be quick to switch to a component instead of just a raw svg file.

Also, it will give us a chance to switch to exporting an inline `<svg>` rather than an ``, as [accessibility experts recommend](https://www.sitepoint.com/tips-accessible-svg/), and fix the absent `alt=` problem on the `img` in this file by adding a customizable `title` attribute to the svg.

chadoh (changes requested):
I like how small and focused this change is, and how thorough your description is! Thanks!

I'd like to see some small improvements to accessibility & code simplicity, but they shouldn't take long. See below.

e18r:
Hi @chadoh, thanks for your review. I addressed both issues you mentioned, so now I consider this PR to be ready for another round of review. Thank you!

chadoh (approved):
👏

PR #1856, opened 2020-01-09
DV a11y fixes

rkzel:
No screenshots, because nothing changed visibly. One thing that changed is less console noise, and couple of Axe warnings disappeared.

### DIFF
--- a/apps/dot-voting/app/components/EmptyFilteredVotes.js
+++ b/apps/dot-voting/app/components/EmptyFilteredVotes.js
@@ -1,6 +1,6 @@
 import React from 'react'
 import PropTypes from 'prop-types'
-import { Box, Button, GU, Text, useTheme } from '@aragon/ui'
+import { Box, Button, GU, Text, textStyle, useTheme } from '@aragon/ui'
 import noResultsSvg from '../assets/no-results.svg'
 
 function EmptyFilteredVotes({ onClear }) {
@@ -24,7 +24,7 @@ function EmptyFilteredVotes({ onClear }) {
           src={noResultsSvg}
           alt="No results"
         />
-        <Text size="xxlarge" color={`${theme.content}`}>
+        <Text css={`${textStyle('title2')}`} color={`${theme.content}`}>
           No results found.
         </Text>
         <div
@@ -33,7 +33,7 @@ function EmptyFilteredVotes({ onClear }) {
             text-align: center;
             margin-bottom: ${4 * GU}px;
             color: ${theme.surfaceContentSecondary};
-            font-size: 16px;
+            ${textStyle('body2')}
           `}
         >
           We can’t find any item matching your search query.{' '}
@@ -46,7 +46,7 @@ function EmptyFilteredVotes({ onClear }) {
               box-shadow: none;
               padding: 0;
               min-width: unset;
-
+              height: auto;
               &:hover {
                 border: none;
                 box-shadow: none;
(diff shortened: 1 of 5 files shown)


apps/dot-voting/app/components/EmptyFilteredVotes.js
```diff
@@ -24,14 +24,14 @@ function EmptyFilteredVotes({ onClear }) {
...
-        <h3
+        <h1
           css={`
             font-size: 28px;
             color: ${theme.content};
           `}
         >
           No results found.
-        </h3>
+        </h1>
```
chadoh: I don't think having this be a heading seems useful. If you think of the headings as being the outline of a document, I don't think it's useful to have "No results found" show up in the outline.

Let's maybe switch to a `<Text size="xxlarge">` while we're here

apps/dot-voting/app/components/VoteDetails/CastVote/EditVoteOption.js
```diff
@@ -68,7 +69,7 @@ const EditVoteOption = ({
           onUpdate={x => onUpdate(Math.round(x * 100))}
           value={value / 100}
         />
-        <HiddenLabel htmlFor="percentage">Percentage</HiddenLabel>
+        <HiddenLabel htmlFor={'percentage' + index}>Percentage {index}</HiddenLabel>
```
chadoh: I think it seems confusing, not helpful, to have the index show up in the user-facing text. The page will read: 

> 0x123 Percentage 1: 67%
> 0x456 Percentage 2: 33%

But it's not the first or second percentage for the label before it – it's the only percentage.

We only need the index here to tie the label to the input, so that clicking on the label focuses the input, and so screen readers understand how the two are related. In that case, we don't need something that's user-facing, and we can use `label`, which we already know is unique.

```jsx
<HiddenLabel htmlFor={`percentage-${label}`}>...
<TextInput id={`percentage-${label}`} ...

chadoh (changes requested):

apps/dot-voting/app/components/EmptyFilteredVotes.js
rkzel: I'll ask design team for opinion :)

apps/dot-voting/app/components/VoteDetails/CastVote/EditVoteOption.js
rkzel: Fixed!

rkzel:
@javieralaves for the "no filtering results" message/card we can use Aragon's named font size instead of what's currently coded: 28px. The choice we have, for the font size name, is either "great" (37px) or "xxlarge" (22px). Screens:

great:

Screenshot from 2020-01-09 15-33-46

xxlarge:

Screenshot from 2020-01-09 15-37-18

Or we can stay at 28px?

chadoh (dismissed):
Code looks good – please wait for input on @javieralaves on your design question before merging!

javieralaves:
@rkzel Can we do textStyle("title2") as described on [aragonUI](https://ui.aragon.org/text-styles/)? And textStyle("body2") for the description text. By the way, the line-height in the body2 text seems too large compared to the design, is there a quick fix to that?

rkzel:
> @rkzel Can we do textStyle("title2") as described on [aragonUI](https://ui.aragon.org/text-styles/)? And textStyle("body2") for the description text.

Sure:

Screenshot from 2020-01-10 17-18-41

OK, @javieralaves ? :)

> By the way, the line-height in the body2 text seems too large compared to the design, is there a quick fix to that?

It was affected by height of the "clear" text (which is a button, not simple text). Fixed, as per the screen.

javieralaves:
Looks good to me, awesome @rkzel 👍

chadoh (approved):

PR #1860, opened 2020-01-10
Improved a11y in Projects

rkzel:
This fixes all problems reported by a11y react-axe caused by our own code (as opposed to problems highlighted in imported UI components). Included in this PR:
 * aria-labels and labels added to input components,
 * used DateInput from shared/ui location instead of Project's own (which could be then finally removed),
 * cosmetic changes required for labels to be correctly passed to final input (HTML input elements) via chain of react components,
 * added landmarks.


shared/ui/components/form/inputs/DateInput.js
```diff
@@ -110,14 +124,15 @@ DateInput.propTypes = {
...
+  wide: PropTypes.bool,
+  label: PropTypes.string.isRequired,
 }
 
 DateInput.defaultProps = {
   value: new Date(),
   format: 'LL/dd/yyyy',
   onChange: () => {},
-  width: '180px',
+  width: '181px',
```
chadoh, line 135: What's this?

chadoh (approved):
Looks great!

Have we made issues in the aragonUI repo for all of the React Axe complaints it's responsible for?

PR #1869, opened 2020-01-10
fix(template): renamed deps with identical names

topocount:
The purpose of these changes is to make the flattened
OpenEnterpriseTemplate.sol compile. Previously, shared names between an
interface and a library are making the flattened template impossible to
compile and therefore impossible to verify
- renamed ScriptHelpers to DynamicScriptHelpers
- use @"aragon" IForwarder interface instead of @"tps"

For proof of these changes working, I've deployed and verified a template with these changes here: https://rinkeby.etherscan.io/address/0x07089831754d1a8f20c2688499567c18633ece5a#code


Quazia (approved):
LGTM

PR #1870, opened 2020-01-11
Rewards a11y improvements

rkzel:
Added few labels to input fields, <main> tag where most appropriate, and one alt text to "no rewards" image.

### DIFF
--- a/apps/rewards/app/components/Card/Empty.js
+++ b/apps/rewards/app/components/Card/Empty.js
@@ -4,7 +4,7 @@ import styled from 'styled-components'
 import { Button, EmptyStateCard, GU, LoadingRing, unselectable } from '@aragon/ui'
 import icon from '../../assets/empty-rewards.svg'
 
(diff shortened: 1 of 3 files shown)


apps/rewards/app/components/Card/Empty.js
```diff
@@ -4,7 +4,7 @@ import styled from 'styled-components'
 import { Button, EmptyStateCard, GU, LoadingRing, unselectable } from '@aragon/ui'
 import icon from '../../assets/empty-rewards.svg'
 
-const Icon = () => <img src={icon} height={20 * GU} />
+const Icon = () => <img src={icon} height={20 * GU} alt="No rewards" />
```
chadoh, line 7: I doubt we want this – looking below, we already say "No rewards here!" Having the alt text repeat that message will result in screen readers saying "No rewards no rewards here!"

Instead, let's set it to empty. `alt=""`

chadoh (approved):
One quick thing for cleanup, but otherwise looks good! 👏

PR #1878, opened 2020-01-13
chore(linting): bump commitlint to v8.3

topocount:
This allows for the breaking change `!` token in the commit headers


Quazia (approved):
BLESS

PR #1894, opened 2020-01-16
refactor(projects): mv Tabs & Header out of App.js

chadoh:
Soon, we will get rid of Issues from the Tabs bar, and when you tap on a project from Overview, you will see a ProjectDetail view. This ProjectDetail view will be very similar to the current Issues component, but will show a back button & filter bar instead of the tabs, and may have different header text.

Rather than adding more edge-case logic to App.js, I simplified it, moving this logic to the Content components. Here's what changed:

* rm Header and Tabs from App.js, add to Content components
* extract Tabs logic to own component, easy to use anywhere
* relocate which-Content-to-show logic to Routes component
* clean up props passed to Content components – Issues saw the biggest cleanup. Some of the props declared in its propTypes were vestigial, and many that were passed from App.js could be looked up in the component itself.

While testing this, I discovered a bug, triggered by these steps:

1. add a project
2. visit `/issues`
3. refresh the page
4. cached data loads initially
5. aragonAPI replays info from the blockchain, since current data is not more than 100 blocks old and nothing is cached in its `store`
6. data gets reset, so `repos` becomes empty
7. we construct a nonsense GraphQL query for an empty set of repos

The fix for this required one line, so I included it here.

I recommend reading the changes [with indentation changes hidden](https://github.com/AutarkLabs/open-enterprise/pull/1894/files?w=1)


rkzel (approved):
It's working. Love it!

PR #1898, opened 2020-01-17
Added filtering to Review Panels

rkzel:
The idea is: Review Panels (work application, work submission) could have simple filters of reviewed/available for review type like this:

Screenshot from 2020-01-15 15-09-25

and this:

Screenshot from 2020-01-15 15-09-55

In case of read-only panels (panels present read-only information (no forms, no review buttons)) when they are opened from activity log in issue details. RO panels looks like this:

Screenshot from 2020-01-16 12-57-34

Screenshot from 2020-01-16 13-51-45

There are 3 possible states for a submission to be in, which is reflected in what appears in the first DropDown:
- "Available for review" - such submissions also display accept/reject buttons,
- "Reviewed" - those are shown with Status,
- "Unreviewed" - and this one is shown when the submisison was not reviewed, but it can't be reviewed, because the issue got fulfilled in the meantime. Or there are more work submissions that bounties (currently there is only one bounty per issue, it will be different for multibounties).

There is also a difference between work applications and work submissions: there is no limit to applications. Until work has been submitted unlimited number of applications can be created - and they can all be accepted. Accepting second application has the following effect: second applicant's address replaces the first applicants as `assignee`, but the applications itself remains accepted (because acceptance info is stored on IPFS). I'm not sure if it is a bug really, feature or side effect - it can be used by someone with enough privilege to accept applications only, and possibility of accepting multiple applications will be useful in the near future anyway, for multibounties.
Work submissions are limited (to one) in non-open-submission mode, and for open-submission mode there is limit of accepted submissions - currently one (but there was some initial work done in preparation for handling multiple submissions for multibounties).


chadoh:
> In case of read-only panels when they are opened from activity log in issue details

I discussed this with @javieralaves, and this is a bug. It makes no sense that you see a read-only panel based on where you open it from!

If you are allowed to review the thing you're looking at, you should be shown the editable version. If you're not allowed, you should see the read-only version.

Do you think it's a small enough change to fix this while we're here, or would you prefer to put it in a separate PR?

chadoh:
> second applicant's address replaces the first applicants as assignee, but the applications itself remains accepted (because acceptance info is stored on IPFS). I'm not sure if it is a bug really, feature or side effect 

Does this essentially have the effect of rejecting the previously-accepted application? This would mean that the initially-accepted applicant can no longer submit their work – right?

If that's the case, it seems like a bug to me. However, maybe waiting to fix it until we work on multibounties is the best path forward.

chadoh:
Thoughts on the commit message. Here's what it is currently:

> feat: review panels, simple filtering added

Since it's marked `feat`, it will show up in our changelog. And it's a feature! So people might actually care to read about it, as opposed to bugfixes, which are probably less interesting for users to learn about.

But I don't think I would understand what this feature is, just reading this description. Now that our commits are meant to be user-facing, let's actually think about how users will understand them!

How about something like this:

> feat(projects): added filtering to review panels
> 
> When reviewing applications or work submissions, you will now be able to filter to see only those applications that need review, or only those that have already been reviewed.

apps/projects/app/components/Panel/ReviewApplication/ReviewApplication.js
```diff
@@ -34,12 +36,32 @@ const ReviewApplication = ({ issue, requestIndex, readOnly }) => {
...
-  const [ index, setIndex ] = useState(requestIndex)
 
-  const updateFeedback = e => setFeedback(e.target.value)
+  const canReview = issue.workStatus !== 'fulfilled' && !readOnly
+
+  const typesToShow = []
+  if (issue.requestsData.findIndex(request => !('review' in request)) !== -1)
+    typesToShow.push(canReview ? 'Available for review' : 'Unreviewed')
+  if (issue.requestsData.findIndex(request => 'review' in request) !== -1)
+    typesToShow.push('Reviewed')
```
chadoh: As in almost all cases with multi-line `if` statements, I would find this more readable with brackets around the bodies of these. I'd like to suggest a couple other changes too; maybe it's easier to show how I'd write it:

```js
if (issue.requestsData.some(request => request.review)) {
  typesToShow.push('Reviewed')
} 

if (issue.requestsData.some(request => !request.review)) {
  if (canReview) typesToShow.push('Available for review')
  else typesToShow.push('Unreviewed')
}
```

What I like about this:

* separating the two `if` statements with a blank line clearly shows that we want to fire both of these, not that we have an `if/else` (I interpreted it as an `if/else` at first)
* using `some` indicates that we're not actually finding a particular item, just checking that there are _some_ requestsData with or without existing reviews
* using brackets means that we can use a nested `if`, instead of mixing `if`s and ternaries
* I don't like the `'thing' in otherThing` syntax – it seems cute at the cost of being too long-winded, and it requires parentheses in situations that the `otherThing.thing` syntax doesn't

Do you find it more readable this way?

But wait a minute, why do we say `'Unreviewed'` if it's available for review, unfulfilled, but it happens to be `readOnly`? That seems like a bug.

apps/projects/app/components/Panel/ReviewApplication/ReviewApplication.js
```diff
@@ -34,12 +36,32 @@ const ReviewApplication = ({ issue, requestIndex, readOnly }) => {
...
+
+  let indexTypeSelected = 0
+  if ('review' in issue.requestsData[requestIndex])
+    indexTypeSelected = (typesToShow[0] === 'Reviewed') ? 0 : 1
```
chadoh: Can you explain to me in English what this logic does?

> If the selected requestsData is reviewed,
> and if 'reviewed' is the first item in `typesToShow`,
> then select the first item,
> otherwise select the second item.

Does that sound right?

Honestly, even if it's right, I'm not sure what it means! Is there a simpler way to write this logic? Would including it in the `if` statements above be possible or helpful?

apps/projects/app/components/Panel/ReviewApplication/ReviewApplication.js
```diff
@@ -34,12 +36,32 @@ const ReviewApplication = ({ issue, requestIndex, readOnly }) => {
...
+
+  const [ indexType, setIndexType ] = useState(indexTypeSelected)
```
chadoh: I find it strange that it's possible that `indexTypeSelected` will change from one render to the next, but the state set up here will not change with it! As in, if on first render `indexTypeSelected` is `1`, then `indexType` will be initialized to `1`. But if on second render `indexTypeSelected` is `0`, `indexType` will still be `1`, because it was already initialized and we're not calling `setState`.

I'm wondering if the calculation of `indexTypeSelected` and `typesToShow` could be moved to an external function or maybe a hook, to be used like:

```js
const [ filters, selectedFilter ] = useReviewFilters(issue.requestsData)
```

Then they could be memoized and only change if `issue.requestsData` changes.

apps/projects/app/components/Panel/ReviewApplication/ReviewApplication.js
```diff
@@ -34,12 +36,32 @@ const ReviewApplication = ({ issue, requestIndex, readOnly }) => {
...
+
+  const requests = issue.requestsData.filter(r => {
+    return (typesToShow[indexType] === 'Reviewed') ? 'review' in r : !('review' in r)
```
chadoh: Yikes. There's gotta be a way to clean this up.

Also, it seems like it could probably be part of the calculations done in that hook I proposed.

apps/projects/app/components/Panel/ReviewApplication/ReviewApplication.js
```diff
@@ -34,12 +36,32 @@ const ReviewApplication = ({ issue, requestIndex, readOnly }) => {
...
+  })
 
+  const newRequestIndex = requests.findIndex(
+    request => request.contributorAddr === issue.requestsData[requestIndex].contributorAddr
+  )
+  const [ index, setIndex ] = useState(newRequestIndex)
```
chadoh: Two things:

1. Let's rename these variables. `index` is generic to the point of meaninglessness.
2. We don't need to calculate `newRequestIndex` on every render, we only need to calculate it on first render, when we set `useState`. We can move the `findIndex` inline and avoid the clumsy name of `newRequestIndex`

Combining these, maybe we end up with:

```js
const [ selectedItem, setSelectedItem ] = useState(requests.findIndex(request =>
  request.contributorAddr === issue.requestsData[requestIndex].contributorAddr
))
```

But actually, something that would be **way** nicer, if it works, would be:

```js
const [ selectedItem, setSelectedItem ] = useState(requests.findIndex(request =>
  request === issue.requestsData[requestIndex]
))
```

I think this might work, given how object identity in JavaScript works. Please try it!

apps/projects/app/components/Panel/ReviewApplication/ReviewApplication.js
```diff
@@ -53,24 +75,27 @@ const ReviewApplication = ({ issue, requestIndex, readOnly }) => {
   const onAccept = () => onReviewApplication(true)
   const onReject = () => onReviewApplication(false)
   const changeRequest = (index) => setIndex(index)
+  const changeRequestType = (index) => {
+    setIndex(0)
+    setIndexType(index)
```
chadoh: I'm really digging my implicit suggestion above to rename `indexType` to `selectedFilter`. I think this block would end up being:

```js
const selectFilter = filter => {
  setItem(0)
  setSelectedFilter(filter)
}

File: apps/projects/app/components/Panel/ReviewWork/ReviewWork.js [hunk 7]
```diff
@@ -40,7 +42,34 @@ const ReviewWork = ({ issue, submissionIndex, readOnly }) => {
 
   const [ feedback, setFeedback ] = useState('')
   const [ rating, setRating ] = useState(-1)
-  const [ index, setIndex ] = useState(submissionIndex)
+
+  const submissionsApproved = issue.workSubmissions.reduce((counter, submission) => {
+    if ('review' in submission && submission.review.accepted) counter++
```
chadoh: I find this mix of `'thing' in otherThing` and `otherThing.thing.whatever` within the same conditional particularly odd.

Also, I think it might be easier to read & work with if we get rid of the `reduce`:

```js
const approvedSubmissions = issue.workSubmissions.filter(s => s.review && s.review.accepted)

// then later:
approvedSubmissions.length

File: apps/projects/app/components/Panel/ReviewWork/ReviewWork.js [hunk 8]
```diff
@@ -40,7 +42,34 @@ const ReviewWork = ({ issue, submissionIndex, readOnly }) => {
...
+    return counter
+  }, 0)
+
+  const canReview = submissionsApproved < (issue.bounties ? issue.bounties : 1) && !readOnly
```
chadoh: So you can only review if there are fewer reviewed submissions than the total number of bounties for this issue?

Two questions:

* do we actually support `issue.bounties` yet? Is this an interface we are planning to add, but have not yet implemented?
* is there a possibility that there would be N bounties and more than N submissions? If so, why do we want to forbid reviewing submission N+1? I think the fact that I'm even asking this question right now _probably_ means that we're being unnecessarily complex with this statement – there's probably a simpler way to write logic about what our system actually does.

apps/projects/app/components/Panel/ReviewWork/ReviewWork.js
```diff
@@ -58,6 +87,10 @@ const ReviewWork = ({ issue, submissionIndex, readOnly }) => {
   const updateRating = (index) => setRating(index)
   const updateFeedback = e => setFeedback(e.target.value)
   const changeSubmission = (index) => setIndex(index)
+  const changeRequestType = (index) => {
+    setIndex(0)
+    setIndexType(index)
+  }
```
chadoh: It looks like all the logic to define `typesToShow`, `indexTypeSelected`, `workSubmissions`, `newSubmissionIndex`, and even `changeRequestType` is _super_ similar to the logic in the `ReviewApplication` panel. If we can pull it into a hook, I'd be surprised if we wouldn't be able to use the same hook in both spots. Maybe it will end up looking something like this:

```js
const {
  filters,
  selectedFilter,
  filteredItems,
  selectedItem,
  selectFilter,
  selectItem,
} = useReviewFilters({
  allItems: issue.requestsData, // or issue.workSubmissions
  initialSelection: issue.requestsData[submissionIndex], // or issue.workSubmissions[submissionIndex]
})
```

(The `initialSelection` thing will only work if my second suggestion for setting `selectedItem` works out. If not, you may have to pass two separate variables. But there should still be a way to do this!)

chadoh (changes requested):

chadoh:
Do you think we could clean this up if we modeled the relationship between the first dropdown and the second with an object?

```js
{ 'Reviewed': [{ addr: '0x123', username: 'rkzel' }, { addr: '0xabc', username: 'chadoh' }],
  'Need review': [{ addr: '0xdef', username: 'quazia' }],
}
```

apps/projects/app/components/Panel/ReviewApplication/ReviewApplication.js
rkzel: Renamed.

apps/projects/app/components/Panel/ReviewApplication/ReviewApplication.js
rkzel: It does seem to work indeed :)

apps/projects/app/components/Panel/ReviewApplication/ReviewApplication.js
rkzel: > Do you find it more readable this way?

Yes, thx!

> why do we say 'Unreviewed' if it's available for review

Because design says so - under some circumstances, if panel is opened in RO mode, applications are not allowed to be reviewed, and marking them as "Available" would be misleading...

apps/projects/app/components/Panel/ReviewApplication/ReviewApplication.js
rkzel: Unchecked evolution of logic leads to code that works, but is hard to explain why... :) I've fixed this.

apps/projects/app/components/Panel/ReviewApplication/ReviewApplication.js
rkzel: Done!

apps/projects/app/components/Panel/ReviewApplication/ReviewApplication.js
rkzel: Hook created and reused in both panels.

apps/projects/app/components/Panel/ReviewWork/ReviewWork.js
rkzel: That's much better, thank you.

apps/projects/app/components/Panel/ReviewWork/ReviewWork.js
rkzel: > So you can only review if there are fewer reviewed submissions than the total number of bounties for this issue?

Yes. Fewer approved submissions.

> do we actually support issue.bounties yet? 

Not just yet, but hopefully soon.

> is there a possibility that there would be N bounties and more than N submissions

Yes. But only N can be approved, hence the logic... slightly improved now though :)

apps/projects/app/components/Panel/ReviewWork/ReviewWork.js
rkzel: Done!

rkzel:
> Do you think we could clean this up if we modeled the relationship between the first dropdown and the second with an object?

Generally yes, but I ended up with 2D array - found that easier to use in DropDowns later. And yes, it simplified a lot :)

PeterMPhillips (approved):
lgtm

PR #1900, opened 2020-01-20
Projects: refactor routing with usePathHelpers

chadoh:
See individual commits for more info

This lays groundwork for replacing the Issues tab with a ProjectDetail and aligns the routing design of Projects with what we have in Allocations


e18r (approved):
LGTM

PR #1901, opened 2020-01-21
feat: bounties tab

PeterMPhillips:
This PR adds a Bounties tab to the Projects app:
Peek 2020-01-20 12-55

It also supports swiping on mobile:
Peek 2020-01-20 12-56

Additionally, this PR refactors some of the code that queries the issues data. All the queries have been moved from Issues and Bounties into an IssuesQueryContext. This reduces the frequency that the user will see the 'Loading...' screen when they navigate to the Issues and Bounties tabs.

### DIFF
--- a/apps/projects/app/App.js
+++ b/apps/projects/app/App.js
@@ -20,6 +20,7 @@ import { LoadingAnimation } from './components/Shared'
 import { EmptyWrapper } from './components/Shared'
 import { Error } from './components/Card'
 import { DecoratedReposProvider } from './context/DecoratedRepos'
+import { BountyIssuesProvider } from './context/BountyIssues'
 import Routes from './Routes'
 
 let popupRef = null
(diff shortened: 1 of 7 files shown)


chadoh:
@javieralaves my understanding of this design was that we decided the white background in the columns should only be as tall as the content within that column, which matches how Trello works. I think Peter may have implemented it using an older design, which has the white background extend to the height of the entire page, even when there's no content in the column.

Which approach would you like us to take here?

javieralaves:
@chadoh Good point, I didn't update the specifications after our discussion to point that out. If it's a quick-fix to update the issue-responsive column height, let's do it. Otherwise, this is a good iteration that we can improve later on.

chadoh:
@PeterMPhillips @javieralaves I've updated this to be compatible with the latest `dev` and made a couple other improvements. Screenshots of dark mode:

Screen Shot 2020-01-23 at 22 10 20

Side scrolling now uses pure CSS instead of a JS plugin:

bounties side scroll 2020-01-23 at 22 11 45

This works well at first page load, but currently breaks when you resize your window, which is a weird browser bug. I doubt this is important enough to justify the complexity & bundle bloat for a JS plugin, though?

One thing that's currently broken in this new version: navigating back from an issue detail page brings you back to the ProjectDetail for that issue, rather than the Bounties tab. I'd like to implement `goBack` functionality in our `usePathHelpers` hook to address that, which I've done once before and then discarded, so it shouldn't take long. I don't know if it's worth doing as part of this PR or not.

topocount (changes requested):
@PeterMPhillips @chadoh This is looking really smooth! Bountties requiring applications worked perfectly, but when I created multiple open-submission bounties at once, only one loaded in the Bounties view (at least until I did something that required another query of bounty state, in this case accepting the fulfillment of another bounty). The missing bounty showed up in the issues view no problem. Probably won't be too hard of a fix since bounties with applications are already solid

**EDIT** Neither I, nor others have been able to replicate this aside from the first time mentioned above. If it pops up during UAT we'll look more deeply into it.

topocount (approved):

PR #1902, opened 2020-01-21
Projects: remake issues tab as project detail

### LINKED ISSUE #1865: Create project detail view
This issue should cover the following acceptance criteria:

- [ ]  The 'Issues' tab is no longer visible in the Tabs area
- [ ]  Clicking on a project takes you to its detail view
image

The basic idea here is to repurpose the current "Issues" component as a new "ProjectDetail" component.

chadoh:
fixes #1865

Rather than having an Issues tab at the top that, when visited, shows all issues for all projects, you can now only see issues for one project at a time.

This has a few benefits:

* A simpler filtering experience: three filters instead of four.
* Less wasted vertical space: the filter bar now contains a back button, and there's no need to show tabs. Your list of issues displays closer to the top of the page.
* Code maintainability: expect faster changes to projects, now that we were able to simplify the complex logic resulting from multi-project filtering.

I recommend viewing this diff [with indentation changes hidden](https://github.com/AutarkLabs/open-enterprise/pull/1902/files?w=1)

## Screenshot

Three quirks to note:

1. When clicking through from a project to an issue, the heading of the page disappears momentarily – this would be awkward to code around, and I don't think it's worth it. The reason this happens is because it is coded in a way that keeps the URL minimal (project detail at `/projects/:projectId`, issue detail at `/issues/:issueId`), and ensures that refreshing the page will still load data correctly and show the correct heading

2. When you apply filters, then select an issue, then click "Back", your filters are lost – this has two causes: 1. We are not currently implementing filtering via the URL, 2. Even if we were, aragonAPI does not support true "go back" behavior; we could hack something together ourselves, but it would work imperfectly

3. The "Show More" behavior has not changed – you'll see how clumsy it is for large projects in the screen recording below

ProjectDetail 2020-01-21 at 17 19 18

### DIFF
--- a/apps/projects/app/components/Content/IssueDetail/index.js
+++ b/apps/projects/app/components/Content/IssueDetail/index.js
@@ -18,21 +18,24 @@ import DetailsCard from './DetailsCard'
 import BountyCard from './BountyCard'
 import { usePanelManagement } from '../../Panel'
 import usePathHelpers from '../../../../../../shared/utils/usePathHelpers'
+import { useDecoratedRepos } from '../../../context/DecoratedRepos'
 
-function Wrap({ children }) {
+function Wrap({ children, repo }) {
   const { requestPath } = usePathHelpers()
   const { setupNewIssue } = usePanelManagement()
 
   return (
     <>
       <Header
-        primary="Projects"
+        primary={repo && repo.metadata.name}
         secondary={
           <Button mode="strong" icon={<IconPlus />} onClick={setupNewIssue} label="New issue" />
         }
       />
       <Bar>
-        <BackButton onClick={() => requestPath('/issues')} />
+        <BackButton onClick={() => {
+          if (repo) requestPath('/projects/' + repo.data._repo)
+        }} />
       </Bar>
       {children}
     </>
@@ -41,6 +44,14 @@ function Wrap({ children }) {
 
 Wrap.propTypes = {
   children: PropTypes.node.isRequired,
+  repo: PropTypes.shape({
+    data: PropTypes.shape({
+      _repo: PropTypes.string.isRequired,
+    }).isRequired,
+    metadata: PropTypes.shape({
+      name: PropTypes.string.isRequired,
+    }).isRequired,
+  }),
 }
 
 const IssueDetail = ({ issueId }) => {
@@ -53,6 +64,13 @@ const IssueDetail = ({ issueId }) => {
     onError: console.error,
     variables: { id: issueId },
   })
+
+  const repos = useDecoratedRepos()
+  const repo = useMemo(() => {
+    if (!data || !data.node) return null
+    return repos.find(repo => repo.data._repo === data.node.repository.id)
+  }, [ data, repos ])
+
   if (loading) return <Wrap>Loading...</Wrap>
   if (error) return <Wrap>{JSON.stringify(error)}</Wrap>
 
@@ -60,7 +78,7 @@ const IssueDetail = ({ issueId }) => {
   const columnView = layoutName === 'small' || layoutName === 'medium'
 
   return (
-    <Wrap>
+    <Wrap repo={repo}>
       {columnView ? (
         <div css="display: flex; flex-direction: column">
           <div css={`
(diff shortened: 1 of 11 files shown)


javieralaves (dismissed):
Looks safe enough to me 👍

apps/projects/app/components/Content/IssueDetail/index.js
```diff
@@ -18,21 +18,24 @@ import DetailsCard from './DetailsCard'
...
-        primary="Projects"
+        primary={repo && repo.metadata.name}
         secondary={
           <Button mode="strong" icon={<IconPlus />} onClick={setupNewIssue} label="New issue" />
         }
       />
       <Bar>
-        <BackButton onClick={() => requestPath('/issues')} />
+        <BackButton onClick={() => {
+          if (repo) requestPath('/projects/' + repo.data._repo)
```
ottodevs, line 37: `_repo` terminology here seems a bit confusing

apps/projects/app/utils/gql-queries.js
```diff
@@ -33,16 +33,15 @@ const issueAttributes = `
...
-    node${i}: node(id: "${repoId}") {
-      id
+export const getIssuesGQL = ({ repoId, count, after }) => gql`
+  query getIssuesForRepo {
+    repository: node(id: "${repoId}") {
       ... on Repository {
         issues(
           states:OPEN,
-          first: ${repos[repoId].fetch},
-          ${repos[repoId].showMore ? `after: "${repos[repoId].endCursor}",` : ''}
```
ottodevs, line 44: should we want to keep tracking cursor for multi-page issues results?

In case we don't need that anymore, it does not make sense to keep using the "show more results" button as we don't have any mechanism to fetch more items

apps/projects/app/components/Shared/FilterBar/FilterBar.js
```diff
@@ -614,30 +591,14 @@ const FilterMenuItem = styled(ContextMenuItem)`
...
-  width: 100%;
-  display: flex;
-  justify-content: space-between;
-`
 const FilterBarMainLeft = styled.div`
   width: 100%;
   display: flex;
+  margin-left: ${GU}px;
   > * {
     margin-right: 8px;
```
ottodevs: Shouldn't be too much to update this to use `GU` in the same run

ottodevs (dismissed):
Thanks @chadoh great work going on here.

I like how this PR cleans out a lot of code and complexity, I like the projects codebase a lot more in this new state, I am still a bit concerned that removing multi-projects filtering feature puts us a bit behind projects like ZenHub where that works seamlessly, but I guess is matter of trade-offs and definitely rather this approach for the time being.

The limitations you commented on definitely seem to have proper solutions, but I agree for now it is reasonable to separate into future tasks.

The left-margin between back button and first filter button seems a bit tight for me.

I added some minor comment that not necessarily need further changes

The filterbar seems completely unusable at this point, the worst iteration I experienced on it... up to 10 seconds just to open a filter, the counts are not accurate and the whole app completely freezes on the Project details view... I don't think it should be a requirement to use a NASA computer to filter some issues so please, let's take a look to that.

Also there are lot of warnings and errors in the browser console, maybe those are out of scope, but we should be fixing them progressively why we implement new features.

I approve the PR because from the code perspective and acceptance criteria it looks good for me, but I am deeply concerned about these performance issues I faced with this new view, that supposedly should work just fine.

apps/projects/app/components/Content/IssueDetail/index.js
chadoh: Yes, I agree, I'd love to rethink this data structure and get rid of `_repo`! However, this is the data structure used all over our app – I think its roots might be in our script.js logic. Changing it seems far outside the scope of this change.

apps/projects/app/utils/gql-queries.js
chadoh: We do!

See the behavior of the "Show More" button:

```js
onClick={() => {
  this.setState({ cachedIssues: allIssues })
  this.props.setQuery({
    after: data.repository.issues.pageInfo.endCursor,
  })
}}
```

This `setQuery` changes the query params in `ProjectDetailWrap`, setting an `after` attribute to the `endCursor` from the current query. This causes `ProjectDetailQuery` to fire a new query, which gets passed down to `ProjectDetail` as `graphqlQuery`. Then `ProjectDetail` has access to this new `data` object, which in turn has a new `data.repository.issues.pageInfo.endCursor`

chadoh:
@ottodevs good points on the performance issues. I've added two `refactor` commits here that address it somewhat, but it seems that the biggest reason the performance has regressed is the inclusion of `react-axe`. If you comment out the block that adds it in `index.js`, the performance now seems to match the performance on Rinkeby & mainnet.

PR #1907, opened 2020-01-22
UI for native projects in NewProject panel

rkzel:
As per design. Default view (right now GH connectivity is required on App level):

Screenshot from 2020-01-22 14-45-33

When user start typing a title:

Screenshot from 2020-01-23 20-00-06

And when GH is disconnected:

Screenshot from 2020-01-23 19-59-40

### DIFF
--- a/apps/projects/app/components/Panel/NewProject/NewProject.js
+++ b/apps/projects/app/components/Panel/NewProject/NewProject.js
@@ -242,76 +242,100 @@ const GitHubRepoList = ({ handleGithubSignIn }) => {
 }
 GitHubRepoList.propTypes = PropTypes.func.isRequired
 
+const ThematicBreak = () => {
+  const theme = useTheme()
+
+  return (
+    <div css={`
+      align-items: center;
+      display: flex;
+      margin: ${2 * GU}px 0;
+      text-align: center;
+      color: ${theme.surfaceContentSecondary};
+      ::before,
+      ::after {
+        flex: 1;
+        content: ' ';
+        display: block;
+        border-bottom: 1px solid ${theme.border};
+        height: 1px;
+      }
+      ::before {
+        margin-right: ${.5 * GU}px;
+      }
+      ::after {
+        margin-left: ${.5 * GU}px;
+      }
+
+      ${textStyle('label2')};
+    `}>
+      or
+    </div>
+  )
+}
+
 const createProject = () => {}
 
 const NewProject = ({ handleGithubSignIn }) => {
   const [ title, setTitle ] = useState('')
   const [ description, setDescription ] = useState('')
-  const theme = useTheme()
   const { appState: { github = { status: STATUS.INITIAL } } } = useAragonApi()
+
+  if (disableDecoupledProjects) return (
+    <PanelContent>
+      <GitHubRepoList handleGithubSignIn={handleGithubSignIn} />
+    </PanelContent>
+  )
 
   return (
     <PanelContent>
-      {
-        disableDecoupledProjects ? (
-          <GitHubRepoList handleGithubSignIn={handleGithubSignIn} />
-        ) : (
-          <React.Fragment>
-            <InfoBox>
-              Create a new project that belongs and operates entirely within this application, or synchronize one from GitHub.
-            </InfoBox>
-
-            <FormField
-              label="Title"
-              required={github.status !== 'authenticated' || title !== ''}
-              input={
-                <TextInput
-                  wide
-                  value={title}
-                  onChange={e => setTitle(e.target.value)}
-                  aria-label="Title"
-                />
-              }
+      <React.Fragment>
+        <InfoBox>
+          Create a new project that belongs and operates entirely within this application, or synchronize one from GitHub.
+        </InfoBox>
+        <FormField
+          label="Title"
+          required={github.status !== 'authenticated' || title !== ''}
+          input={
+            <TextInput
+              wide
+              value={title}
+              onChange={e => setTitle(e.target.value)}
+              aria-label="Title"
             />
-
-            {title !== '' ? (
-              <>
-              <FormField
-                label="Description"
-                input={
-                  <TextInput.Multiline
-                    name="description"
-                    rows="3"
-                    onChange={e => setDescription(e.target.value)}
-                    value={description}
-                    wide
-                    aria-label="Description"
-                  />
-                }
-              />
-              <Button
-                mode="strong"
+          }
+        />
+        {title !== '' ? (
+          <>
+          <FormField
+            label="Description"
+            input={
+              <TextInput.Multiline
+                name="description"
+                rows="3"
+                onChange={e => setDescription(e.target.value)}
+                value={description}
                 wide
-                onClick={createProject}
-                disabled={title === ''}
-              >
-                Submit
-              </Button>
-              </>
-            ) : (
-              <>
-              <ThematicBreak
-                color={`${theme.surfaceContentSecondary}`}
-                linecolor={`${theme.border}`}
-              >
-                  OR
-              </ThematicBreak>
-              <GitHubRepoList handleGithubSignIn={handleGithubSignIn} />
-              </>
-            )}
-          </React.Fragment>
-        )
-      }
+                aria-label="Description"
+              />
+            }
+          />
+          <Button
+            mode="strong"
+            wide
+            onClick={createProject}
+            disabled={title === ''}
+          >
+            Submit
+          </Button>
+          </>
+        ) : (
+          <>
+          <ThematicBreak />
+          <GitHubRepoList handleGithubSignIn={handleGithubSignIn} />
+          </>
+        )}
+      </React.Fragment>
     </PanelContent>
   )
 }
@@ -343,29 +367,6 @@ const RepoInfo = styled.div`
 `
 const PanelContent = styled.div`
   margin-top: ${3 * GU}px;
-`
-const ThematicBreak = styled.div`
-  align-items: center;
-  display: flex;
-  margin: ${2 * GU}px 0;
-  text-align: center;
-  color: ${props => props.color};
-  ::before,
-  ::after {
-    flex: 1;
-    content: ' ';
-    display: block;
-    border-bottom: 1px solid ${props => props.linecolor};
-    height: 1px;
-  }
-  ::before {
-    margin-right: ${.5 * GU}px;
-  }
-  ::after {
-    margin-left: ${.5 * GU}px;
-  }
-
-  ${textStyle('label2')};
 `
 
 // TODO: Use nodes instead of edges (the app should be adapted at some places)
--- a/apps/projects/app/components/Panel/NewProject/NewProject.js
+++ b/apps/projects/app/components/Panel/NewProject/NewProject.js
@@ -1,22 +1,30 @@
 import React, { useEffect, useRef, useState } from 'react'
 import PropTypes from 'prop-types'
 import styled from 'styled-components'
-import { Button, GU, IconSearch, Info, RadioList, Text, TextInput, theme } from '@aragon/ui'
+import { Button, GU, IconSearch, Info, RadioList, Text, TextInput, textStyle, useTheme } from '@aragon/ui'
 import { GET_REPOSITORIES } from '../../../utils/gql-queries.js'
-import { LoadingAnimation } from '../../Shared'
+import { IconGitHub, LoadingAnimation } from '../../Shared'
 import { Query } from 'react-apollo'
 import { useAragonApi } from '../../../api-react'
 import { usePanelManagement } from '../../Panel'
 import { toHex } from 'web3-utils'
 import noResultsSvg from '../../../assets/noResults.svg'
+import { FormField, FieldTitle } from '../../Form'
+import { STATUS } from '../../../utils/github'
 
-
-const NewProject = () => {
-  const { api, appState: { repos } } = useAragonApi()
+const GitHubRepoList = ({ handleGithubSignIn }) => {
+  const {
+    api,
+    appState: {
+      repos,
+      github = { status: STATUS.INITIAL },
+    },
+  } = useAragonApi()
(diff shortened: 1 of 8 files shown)


stellarmagnet (dismissed):
- The padding is not correct between the field labels. Please double check all of the padding between elements.
- For the "And when GH is disconnected" use case --> the blue box is too close to the field label
- "Feedback" should say "Description"
- Select from Github icon & text should be center-aligned with the button

apps/projects/app/components/Panel/NewProject/NewProject.js
```diff
@@ -113,110 +121,187 @@ const NewProject = () => {
...
-                                color: ${theme.surfaceOpened};
-                                margin-right: 8px;
-                              `}
-                          />
-                        )
-                      }
-                      adornmentPosition="end"
-                      ref={searchRef}
-                      aria-label="Search"
+  const RepoQuery = () => (
```
chadoh: Why define this inside `GithubRepoList` instead of as its own standalone component?

apps/projects/app/components/Panel/NewProject/NewProject.js
```diff
@@ -113,110 +121,187 @@ const NewProject = () => {
...
+      >
+        Submit
+      </Button>
+    </>
+  )
+}
+GitHubRepoList.propTypes = PropTypes.func.isRequired
+
+const createProject = () => {
+  console.log('creating native project')
```
chadoh: Ah, it looks like this PR is meant to be left unmerged until we implement the rest of this? 

Can we build this in a "feature flipped" way, so that we don't have to delay merging this PR for an extended period of time, but we also don't have unsupported elements in our UI? Ideally, we could do this with an environment variable, or maybe a URL param (like, add `?features=nativeprojects` to the end of the URL), but for an initial version I think we could even have something like `if (false)` in the code somewhere, to make it easy to toggle on & off in a development environment.

apps/projects/app/components/Panel/NewProject/NewProject.js
```diff
@@ -113,110 +121,187 @@ const NewProject = () => {
...
+            onChange={e => setTitle(e.target.value)}
+            aria-label="Title"
+          />
+        }
+      />
+
+      {title !== '' ? (
+        <>
+        <FormField
+          label="Feedback"
```
chadoh: Wrong field name

chadoh (changes requested):
Code looks pretty clean, but there are some changes I'd like to see, in addition to the UI changes @stellarmagnet requested above

apps/projects/app/components/Panel/NewProject/NewProject.js
rkzel: Amazing how so glaringly obvious can be missed. Fixed.

apps/projects/app/components/Panel/NewProject/NewProject.js
rkzel: It was one cleanup I wasn't convinced needed to be done immediately - but you're right, I've split it into 3 components.

apps/projects/app/components/Panel/NewProject/NewProject.js
rkzel: It was meant to be merged at more or less the same time as the rest of it - but "feature flipped" makes sense as safety measure, to make sure there are no dangling threads accidentaly deployed where they shouldn't.
The easiest way in this particular case was to disable the -OR- part with a small `true?` statement - not very elegant, but it felt justified...

rkzel:
> * The padding is not correct between the field labels. Please double check all of the padding between elements.

Improved that by fixing height line and height of the field label component - which is used everywhere in the app, so I made it as much as I could without checking all the other places.

> * For the "And when GH is disconnected" use case --> the blue box is too close to the field label

Fixed. Styled-components module problem.

> * "Feedback" should say "Description"

Sorry, fixed.

> * Select from Github icon & text should be center-aligned with the button

Improved!

Screenshots added to the first, top comment.

apps/projects/app/components/Panel/NewProject/NewProject.js
```diff
@@ -1,41 +1,87 @@
...
+  repoSelected: PropTypes.number.isRequired,
+  visibleRepos: PropTypes.array.isRequired,
+}
+
+const RepoQuery = ({ onRepoSelected, repoSelected, setRepoSelected }) => {
+  const theme = useTheme()
   const [ filter, setFilter ] = useState('')
-  const [ project, setProject ] = useState()
-  const [ repoSelected, setRepoSelected ] = useState(-1)
+  const searchRef = useRef(null)
```
chadoh: Small note: I don't think you need the `null` here. `useRef()` will initialize just fine without it

apps/projects/app/components/Panel/NewProject/NewProject.js
```diff
@@ -63,160 +204,127 @@ const NewProject = () => {
...
+
+  return (
+    <PanelContent>
+      <InfoBox>
+        Create a new project that belongs and operates entirely within this application, or synchronize one from GitHub.
+      </InfoBox>
+
+      {
+        // eslint-disable-next-line no-constant-condition
+        true ? (
```
chadoh: This isn't the part we need to feature flip!

Until native projects are actually supported, we want to hide everything related to native projects. We want the experience to match what's in mainnet & rinkeby today. 

That means we _only_ want the `GitHubRepoList`. We want to hide the "Title" field. We want to hide the `InfoBox`.

chadoh (changes requested):

chadoh:
@rkzel I explored feature-flipping using URL query params but that [won't work](https://github.com/AutarkLabs/open-enterprise/pull/1918). How about instead we only show the Native Projects stuff if the `connectedAccount` matches an array of hard-coded addresses? We can add our development superuser to this list (`0xb412`) as well as the Rinkeby addresses of @stellarmagnet and @javieralaves.

apps/projects/app/components/Panel/NewProject/NewProject.js
```diff
@@ -63,160 +204,127 @@ const NewProject = () => {
...
+          <GitHubRepoList handleGithubSignIn={handleGithubSignIn} />
+        ) : (
+        <>
+      <FormField
```
chadoh: indentation seems way off

apps/projects/app/components/Panel/NewProject/NewProject.js
chadoh: Ah, I see – only the InfoBox was missed. The indentation of the part below had me thinking this was more broken than it is

apps/projects/app/components/Panel/NewProject/NewProject.js
rkzel: I think so - fixed.

apps/projects/app/components/Panel/NewProject/NewProject.js
rkzel: Added InfoBox to the excluded part and fixed misleading intendation.

apps/projects/app/components/Panel/NewProject/NewProject.js
rkzel: Yes. Fixed it in case it's going to be needed for more than one day.

apps/projects/app/components/Panel/NewProject/NewProject.js
```diff
@@ -63,160 +206,128 @@ const NewProject = () => {
...
+const NewProject = ({ handleGithubSignIn }) => {
+  const [ title, setTitle ] = useState('')
+  const [ description, setDescription ] = useState('')
+  const theme = useTheme()
+  const { appState: { github = { status: STATUS.INITIAL } } } = useAragonApi()
+
+  return (
+    <PanelContent>
+      {
+        // eslint-disable-next-line no-constant-condition
```
chadoh: you don't need this comment anymore

rkzel: True. Removed.

apps/projects/app/components/Panel/NewProject/NewProject.js
```diff
@@ -63,160 +206,127 @@ const NewProject = () => {
...
+        wide
+        onClick={handleNewProject}
+        disabled={repoSelected < 0}
+      >
+        Submit
+      </Button>
+    </>
+  )
+}
+GitHubRepoList.propTypes = PropTypes.func.isRequired
```
chadoh, line 243: what's this?

apps/projects/app/components/Panel/NewProject/NewProject.js
```diff
@@ -63,160 +206,127 @@ const NewProject = () => {
...
+                Submit
+              </Button>
+              </>
+            ) : (
+              <>
+              <ThematicBreak
+                color={`${theme.surfaceContentSecondary}`}
+                linecolor={`${theme.border}`}
+              >
+                  OR
```
chadoh: indentation is off

apps/projects/app/components/Panel/NewProject/NewProject.js
```diff
@@ -63,160 +206,127 @@ const NewProject = () => {
...
+                disabled={title === ''}
+              >
+                Submit
+              </Button>
+              </>
+            ) : (
+              <>
+              <ThematicBreak
+                color={`${theme.surfaceContentSecondary}`}
+                linecolor={`${theme.border}`}
```
chadoh: I think it would clean up this component nicely if these props didn't have to be passed in

```jsx
const ThematicBreak = ({ children }) => {
  const theme = useTheme()
  return (
    <div css={`
      ...
    `}>
      {children}
    </div>
  )
}
```

Or maybe if the "or" is also included in the component itself:

```jsx
const ThematicBreak = () => {
  const theme = useTheme()
  return (
    <div css={`
      ...
    `}>
      or
    </div>
  )
}
```

apps/projects/app/components/Panel/NewProject/NewProject.js
```diff
@@ -63,160 +206,127 @@ const NewProject = () => {
...
+  const [ description, setDescription ] = useState('')
+  const theme = useTheme()
+  const { appState: { github = { status: STATUS.INITIAL } } } = useAragonApi()
+
+  return (
+    <PanelContent>
+      {
+        disableDecoupledProjects ? (
+          <GitHubRepoList handleGithubSignIn={handleGithubSignIn} />
+        ) : (
```
chadoh: an early return would look nicer, keep the indentation the way you had it, and make it easier to remove this feature-flipping later:

```jsx
if (disableDecoupledProjects) {
  return (
    <PanelContent>
      <GitHubRepoList handleGithubSignIn={handleGithubSignIn} />
    </PanelContent>
  )
}

return (
  <PanelContent>
    ...
  </PanelContent>
)
```

chadoh (dismissed):
Please at least fix the invalid propTypes before merging, if not also my other comments

Otherwise, it seems this works as desired

apps/projects/app/components/Panel/NewProject/NewProject.js
rkzel: It's a shortcut. In case there is only one property one does not need to write:

```
GitHubRepoList.propTypes = {
   handleGithubSignIn: PropTypes.func.isRequired
}
```

it is sufficient to say:

`GitHubRepoList.propTypes = PropTypes.func.isRequired`

apps/projects/app/components/Panel/NewProject/NewProject.js
rkzel: Ok. The scecond one makes sense - this component is only used once, it will probably retain colors and text if it needs to be reused, and it can alway be re-customized later.. Changed.

apps/projects/app/components/Panel/NewProject/NewProject.js
rkzel: Good point. Removing feature-flipping later will produce much less commit noise too. Changed.

apps/projects/app/components/Panel/NewProject/NewProject.js
rkzel: And that got fixed as well...

apps/projects/app/components/Panel/NewProject/NewProject.js
```diff
@@ -63,160 +206,151 @@ const NewProject = () => {
...
+}
+
+const createProject = () => {}
+
+const NewProject = ({ handleGithubSignIn }) => {
+  const [ title, setTitle ] = useState('')
+  const [ description, setDescription ] = useState('')
+  const { appState: { github = { status: STATUS.INITIAL } } } = useAragonApi()
+
+  if (disableDecoupledProjects) return (
```
Quazia, line 284: This needs to be removed.

apps/projects/app/components/Panel/NewProject/NewProject.js
```diff
@@ -1,41 +1,89 @@
...
+import { IconGitHub, LoadingAnimation } from '../../Shared'
 import { Query } from 'react-apollo'
 import { useAragonApi } from '../../../api-react'
 import { usePanelManagement } from '../../Panel'
 import { toHex } from 'web3-utils'
 import noResultsSvg from '../../../assets/noResults.svg'
+import { FormField, FieldTitle } from '../../Form'
+import { STATUS } from '../../../utils/github'
 
+const disableDecoupledProjects = true
```
Quazia, line 15: Get rid of this too.

apps/projects/app/components/Panel/NewProject/NewProject.js
```diff
@@ -63,160 +206,151 @@ const NewProject = () => {
...
+      }
+
+      ${textStyle('label2')};
+    `}>
+      or
+    </div>
+  )
+}
+
+const createProject = () => {}
```
Quazia, line 277: Add the logic in here.

Quazia (approved):
Added some comments for once we have the contract changes merged but it's looking good to merge.

PR #1908, opened 2020-01-22
test(bounties): remove unnecessary wallet dependency

topocount:
The truffle hd wallet provider in the Standard Bounties subrepo is only
needed for use on the ropsten testnet. Since we aren't deploying
anything to Ropsten currently, I'm removing it since it's currently
crashing truffle during test deploys


rkzel (approved):
LGTM.

PR #1911, opened 2020-01-22
fix(dot-voting): smooth transition for side panel

PeterMPhillips:
This fixes the issue where the side panel would just appear instead of doing a smooth transition. Turns out this is caused by the `autofocus` parameter on the Description field in the NewVote component. Aragon has accounted for this problem though, we just need to get `readyToFocus` from the `useSidePanel()` hook and pass it to the `autofocus` parameter.


chadoh (approved):
huh!

Issue #1897, opened 2020-01-17
Add and modify decentralized issues


Quazia:
This should cover the following acceptance criteria:
 

- [ ] New issues for decentralized projects can be added, edited and removed

![image](https://user-images.githubusercontent.com/6549550/72574902-c3d63000-3898-11ea-92ce-436e150ab9e3.png)

Issue #1875, opened 2020-01-13
Add contract function for decentralized 'New Project' panel call


Quazia:
This issue in combination with #1874 and #1876 should cover the following acceptance criteria:
- [ ]  'New project' panel allows users to switch between decentralized projects and GitHub
- [ ]  If user is not connected to GitHub, request to connect will appear as designed

This will require contract level changes for the Projects app, a function call needs to be added. It's possible that the `addRepo` call can be recycled, but it's probably best to adjust the language, and ensure all data-points we were previously relying on github to capture are being represented in the contracts data (title, description). Some of this may be possible strictly modifying the IPFS, but this alone would require adding a CID to the `addRepo` call.

PR #1912, opened 2020-01-23
Decoupled projects: Contract changes

topocount:
Addresses #1875 and #1897

This PR adds functions to the projects contract for:
- creating decoupled projects that can include an IPFS hash
- creating and modifiying issues that don't yet contain bounties

This PR also contains
- gas optimizations to allow these issues to fit in the contract
- the removal of the standard bounties bytecode check to make room for these features, since `isContract` should be sufficient

**Note**
- [x] merge #1908 before this PR

### DIFF
--- a/apps/projects/contracts/Projects.sol
+++ b/apps/projects/contracts/Projects.sol
@@ -122,6 +122,7 @@ contract Projects is AragonApp, DepositableStorage {
     BountySettings public settings;
     Vault public vault;
     // Auth roles
+    bytes32 public constant CREATE_ISSUES_ROLE =  keccak256("CREATE_ISSUES_ROLE");
     bytes32 public constant FUND_ISSUES_ROLE =  keccak256("FUND_ISSUES_ROLE");
     bytes32 public constant REMOVE_ISSUES_ROLE = keccak256("REMOVE_ISSUES_ROLE");
     bytes32 public constant ADD_REPO_ROLE = keccak256("ADD_REPO_ROLE");
@@ -145,7 +146,6 @@ contract Projects is AragonApp, DepositableStorage {
     string private constant ERROR_INVALID_AMOUNT = "INVALID_TOKEN_AMOUNT";
     string private constant ERROR_ETH_CONTRACT = "WRONG_ETH_TOKEN";
     string private constant ERROR_REPO_MISSING = "REPO_NOT_ADDED";
-    string private constant ERROR_REPO_EXISTS = "REPO_ALREADY_ADDED";
     string private constant ERROR_USER_APPLIED = "USER_ALREADY_APPLIED";
     string private constant ERROR_NO_APPLICATION = "USER_APPLICATION_MISSING";
     string private constant ERROR_NO_ERC721 = "ERC_721_FORBIDDEN";
@@ -175,8 +175,10 @@ contract Projects is AragonApp, DepositableStorage {
     }
 
     struct Repo {
-        mapping(uint256 => Issue) issues;
         uint index;
+        mapping(uint256 => Issue) issues;
+        bool decoupled;
+        string repoData;
     }
 
     struct AssignmentRequest {
@@ -188,28 +190,25 @@ contract Projects is AragonApp, DepositableStorage {
     struct Issue {
         bytes32 repo;  // This is the internal repo identifier
         uint256 number; // May be redundant tracking this
-        bool hasBounty;
-        bool fulfilled;
         address tokenContract;
         uint256 bountySize;
+        uint256 fulfilledAmount;
         uint256 priority;
-        address bountyWallet; // Not sure if we'll have a way to "retrieve" this value from status open bounties
         uint standardBountyId;
         address assignee;
+        string issueData;
         address[] applicants;
         //uint256 submissionQty;
-        uint256[] submissionIndices;
         mapping(address => AssignmentRequest) assignmentRequests;
     }
 
     // Fired when a repository is added to the registry.
-    event RepoAdded(bytes32 indexed repoId, uint index);
+    event RepoAdded(bytes32 indexed repoId, uint index, bool decoupled);
+    event RepoUpdated(bytes32 indexed repoId, uint index, string repoData);
     // Fired when a repository is removed from the registry.
     event RepoRemoved(bytes32 indexed repoId, uint index);
-    // Fired when a repo is updated in the registry
-    event RepoUpdated(bytes32 indexed repoId, uint newIndex);
     // Fired when a bounty is added to a repo
-    event BountyAdded(bytes32 repoId, uint256 issueNumber, uint256 bountySize, uint256 registryId, string ipfsHash);
+    event IssueUpdated(bytes32 repoId, uint256 issueNumber, uint256 bountySize, uint256 registryId, string ipfsHash);
     // Fired when a bounty is removed
     event BountyRemoved(bytes32 repoId, uint256 issueNumber, uint256 oldBountySize);
     // Fired when an issue is curated
@@ -281,7 +280,7 @@ contract Projects is AragonApp, DepositableStorage {
     ) external auth(CHANGE_SETTINGS_ROLE)
     {
         require(_expMultipliers.length == _expLevels.length, ERROR_LENGTH_MISMATCH);
-        require(_isBountiesContractValid(_bountyAllocator), ERROR_STANDARD_BOUNTIES_NOT_CONTRACT);
+        require(isContract(_bountyAllocator), ERROR_STANDARD_BOUNTIES_NOT_CONTRACT);
         settings.expLevels.length = 0;
         settings.expMultipliers.length = 0;
         for (uint i = 0; i < _expLevels.length; i++) {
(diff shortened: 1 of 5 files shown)


apps/projects/test/Projects.test.js
```diff
@@ -1342,7 +1443,7 @@ contract('Projects App', accounts => {
         })
       })
 
-      it('cannot update bounties contract with contract of invalid size', async () => {
```
Quazia, line 1345: Might as well just remove this text I think?

apps/projects/contracts/Projects.sol
```diff
@@ -321,10 +298,14 @@ contract Projects is AragonApp, DepositableStorage {
      * @param _repoId The id of the repo in the projects registry
      * @return index the repo registry index
```
Quazia: Return is wrong now

Quazia (dismissed):
Looks good, in integration we'll need to check for all instances of `IssueUpdated` and the parsing of return data from `getIssue` I'm not seeing anything else that should need changing.

rkzel:
Specification suggests that there should be a fuction to edit project as well as add/remove. Editing will involve only replacing ipfs hash saved in`_projectData`, so all we need is a simple function like this:

```
    function updateRepo(bytes32 _repoId, string _repoData) external auth(....?)
     {
        require(isRepoAdded(_repoId), ERROR_REPO_MISSING);
        uint rowToUpdate = repos[_repoId].index;
        repoIndex[repoIndexLength] = _repoId;
        repos[rowToUpdate].repoData = _repoData;
        emit Repo....?
    }
```
Probably a new right and event to emit  need to be added as well - what do you think, @topocount ? Alternatively if it's too complicated for current timeline, maybe we leave it for now and not allow editing projects at all?

apps/projects/contracts/Projects.sol
```diff
@@ -319,12 +297,16 @@ contract Projects is AragonApp, DepositableStorage {
     /**
      * @notice Get an entry from the registry.
      * @param _repoId The id of the repo in the projects registry
-     * @return index the repo registry index
+     * @return index the repo registry index, number of open bounties, decoupled bool, repo metadata
```
ottodevs, line 300: It is preferable to document `@return` params into separate lines each one

apps/projects/contracts/Projects.sol
```diff
@@ -357,20 +339,25 @@ contract Projects is AragonApp, DepositableStorage {
 // Repository functions
 ///////////////////////
     /**
-     * @notice Add repository
+     * @notice Add new Project: `_projectData`.
      * @param _repoId The id of the repo in the projects registry
```
ottodevs, line 343: We should document the rest of params as well

apps/projects/contracts/Projects.sol
```diff
@@ -578,6 +567,25 @@ contract Projects is AragonApp, DepositableStorage {
...
+     * @param _data the provided proof of fulfillment
+     */
+    function fulfillBounty(
+        address _sender,
+        uint _bountyId,
+        address[] _fulfillers,
+        string _data
+    ) external
+    {
+        revert();
```
ottodevs, line 586: If we just want to display the radspec, why revert? Also this should be a pure function

ottodevs: Revert will cause non 0 exit status and that maybe non gracefully handled by the frontend

apps/projects/contracts/Projects.sol
```diff
@@ -781,40 +754,68 @@ contract Projects is AragonApp, DepositableStorage {
...
+    function _setRepo(
+        bytes32 _repoId,
+        bool _decoupled,
+        string _repoData
+    ) internal
+    {
+        repos[_repoId].repoData = _repoData;
+        if (!isRepoAdded(_repoId)) {
+            repoIndex[repoIndexLength] = _repoId;
+            repos[_repoId].index = repoIndexLength++;
```
ottodevs, line 766: Are we accounting on a potential overflow here?

apps/projects/contracts/Projects.sol
```diff
@@ -781,40 +754,68 @@ contract Projects is AragonApp, DepositableStorage {
...
+        uint256[] _deadlines,
+        uint256[] _tokenTypes,
+        address[] _tokenContracts,
+        string _ipfsAddresses,
+        string _description,
+        bool _open
+    ) internal
+    {
+        // ensure the transvalue passed equals transaction value
+        //checkTransValueEqualsMessageValue(msg.value, _bountySizes,_tokenBounties);
```
ottodevs, line 787: Is this comment relevant anymore or just dead code? We should remove to avoid confusion

apps/projects/contracts/Projects.sol
```diff
@@ -781,40 +754,68 @@ contract Projects is AragonApp, DepositableStorage {
...
+        string memory ipfsHash;
+        uint standardBountyId;
+        require(bytes(_ipfsAddresses).length == (CID_LENGTH * _bountySizes.length), ERROR_CID_LENGTH);
```
ottodevs, line 790: We should test how it behaves with `_bountySizes.length == 0` as it could bypass the condition in edge cases

apps/projects/contracts/Projects.sol
```diff
@@ -926,31 +927,48 @@ contract Projects is AragonApp, DepositableStorage {
         uint256 _bountySize,
         string _ipfsHash
     ) internal
+    {
+        //Issue storage issue = repos[_repoId].issues[_issueNumber];
+        //require(isRepoAdded(_repoId), ERROR_REPO_MISSING);
```
ottodevs, line 932: Is this comment relevant? we should clean in case not

apps/projects/contracts/test/BountiesEvents.sol
```diff
@@ -3,32 +3,38 @@ pragma solidity ^0.4.24;
...
+     * Events
+     */
+
+    event BountyIssued(uint _bountyId, address _creator, address [] _issuers, address [] _approvers, string _data, uint _deadline, address _token, uint _tokenVersion);
+    event ContributionAdded(uint _bountyId, uint _contributionId, address  _contributor, uint _amount);
+    event ContributionRefunded(uint _bountyId, uint _contributionId);
+    event ContributionsRefunded(uint _bountyId, address _issuer, uint [] _contributionIds);
+    event BountyDrained(uint _bountyId, address _issuer, uint [] _amounts);
+    event ActionPerformed(uint _bountyId, address _fulfiller, string _data);
+    event BountyFulfilled(uint _bountyId, uint _fulfillmentId, address  [] _fulfillers, string _data, address _submitter);
```
ottodevs, line 32: Some spacing between array brackets and types seem wrong with inconsistently 1 or 2 spaces, maybe we should try to be consistent about it, I vote for no spacing between type and brackets
```suggestion
    event BountyFulfilled(uint _bountyId, uint _fulfillmentId, address[] _fulfillers, string _data, address _submitter);
```

Some for the rest of lines on this block

apps/projects/test/Projects.test.js
```diff
@@ -1,4 +1,4 @@
-/* global artifacts, assert, before, beforeEach, contract, context, expect, it, web3 */
+/* global artifacts, assert, before, beforeEach, contract, context, expect, it, web3, xit */
```
ottodevs, line 1: `xit` should actually not need to be defined here, as tests should only be skipped during deployment

apps/projects/test/Projects.test.js
```diff
@@ -872,6 +848,46 @@ contract('Projects App', accounts => {
...
+            [9],
+            { from: bountyManager }
+          )
+
+          await app.reviewSubmission(
+            repoId,
+            1,
+            1,
+            true,
+            'QmbUSy8HCn8J4TMDRRdxCbK2uCCtkQyZtY6XYv3y7kLgDl',
```
ottodevs, line 873: I would move most of these arbitrary magic values to constante to have a more manageable/less error prone test suite, but that's only a personal preference

apps/projects/test/Projects.test.js
```diff
@@ -872,6 +848,46 @@ contract('Projects App', accounts => {
...
+          )
+          
+          const bounty1 = await app.getIssue(repoId, 1)
+          assert.deepEqual(
+            bounty1,
+            [ 
+              new web3.BigNumber(56),
+              new web3.BigNumber(15),
+              new web3.BigNumber(0),
+              '0xffffffffffffffffffffffffffffffffffffffff',
```
ottodevs, line 885: We should use ANY_ADDR (line 16) constant in places like here, otherwise, we define it at all?

apps/projects/test/Projects.test.js
```diff
@@ -1156,6 +1170,122 @@ contract('Projects App', accounts => {
       })
     })
 
+    context('decoupled project', () => {
+      let repoId
+      beforeEach(async () => {
+        await app.setRepo('1', true, 'abc')
```
ottodevs, line 1176: We should use real repo ids as in other test cases to avoid false negatives when repoId length does not find in the contract field

ottodevs (changes requested):
Changed base branch to merge, to properly see the exact changes going on here

Most of the contract and tests look fine for me, and they are passing. I am a bit concerned about opening potential security vectors, as we are not auditing this new code on this review.

Just some minor details to comment about:

It would be preferable to properly document each function that was added or changed
I would just take advantage of this work to keep the contract clean and audit-ready by properly documenting this new code.

I would also bet to improve the tests suite and make it more readable/ less error prone by making use of constants and avoiding magic numbers and text everywhere.

topocount:
> Changed base branch to merge, to properly see the exact changes going on here

We're working out of the https://github.com/AutarkLabs/open-enterprise/tree/decentralized-project-integration banch, and this branch has already been merged in there. I'll make your requested changes in there and redirect https://github.com/AutarkLabs/open-enterprise/tree/issue-querying to that branch as well.

apps/projects/contracts/Projects.sol
topocount: This function should never execute. It's just a workaround for getting the radspec showing up in the frontend. BY forcing it to revert, it effectively doesn't exist and is an invalid call on the contract

apps/projects/contracts/Projects.sol
topocount: if someone passes in _bountySizes with a length of zero, they' just be wasting gas, since the for loop won't be entered

apps/projects/test/Projects.test.js
topocount: Projects will be using integers for decoupled repoIds going forward. They'll still get hashed by the frontend before passing them in, but that's really just a messier magic number in the context of tests

topocount:
This branch has been merged into PR #1941. The review above has been addressed there

PR #1914, opened 2020-01-23
Disabled react-axe

rkzel:
react-axe was enabled to find a11y issues. Even in production mode it highlights many issues we can't fix right now - and it has impact on performance. This PR disables it completely.


chadoh (approved):
thanks! 🙏

PR #1917, opened 2020-01-24
feat(projects): add proper "Back" to IssueDetail

chadoh:
If you get to the IssueDetail page from the Bounties tab, clicking "Back" will now send you back to it, rather than to ProjectDetail

If you visit an IssueDetail page directly, or refresh your browser on an IssueDetail page, clicking "Back" will still send you to ProjectDetail

This has been added in a way that will be easy to use in other places around Autark apps

goBack to Bounties tab 2020-01-24 at 9 05 29


topocount (approved):
LGTM

PR #1918, opened 2020-01-24
refactor: add 'query' logic to usePathHelpers

chadoh:
This reverts commit c981c34c6df185504c7afc855b57275963f41fad

Nothing uses this now, but it's useful logic to have around.  Specifically, it could be nice to use for things like feature-flipping features that are under construction by adding params to the URL.  Example:

    /?features=native-projects

### DIFF
--- a/shared/utils/usePathHelpers.js
+++ b/shared/utils/usePathHelpers.js
@@ -1,6 +1,9 @@
 import React from 'react'
 import { usePath } from '@aragon/api-react'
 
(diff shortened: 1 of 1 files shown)


chadoh:
@rkzel if you like this, you can merge it, rebase `1874-native-projects-panel` onto `dev`, and then use this to add proper (though simple) feature-flipping to the New Project panel

chadoh:
NEVERMIND. This doesn't work because `usePath` doesn't pass in the query parameters as of `@aragon/api-react@2.0.0-beta.9` 😒 😒 😒

PR #1923, opened 2020-01-24
Projects: improve querying behavior

chadoh:
Everything has been implemented according to the idea. Effect: pagination works. Downloading issues from GH is fast enough (not blindingly, but it is better than it used to be, especially with pagination). However... there are challenges discovered along the way:
1. Labels can be used for filtering issues easily: 

`filterBy: { labels: ["bug", "app: projects"]}`

Filtering by milestones doesn't work that way. When we discuss all of it, we may need to decide to take out milestones completely.

2. Text search is not compatible with filtering by labels. Filtering in normal `query` realized with `filterBy` (as above) does not provide any mean for filtering issues by title. What was intended: using `search` query like this:
```
    search(
      query:"is:issue state:open ${text to search for} in:title label:bug label:enhancement",
    ) {
```
the problem is, when there is more than one label, it searches for issues with _all_ the labels - that's different to what we do now, which is searching for issues with any of the labels.
In other words, `label:bug label:enhancement` will select issues with both of those labels, and in projects app selecting "bug" and "enhancement" from Labels dropdown will select issues with any of them.
This required using different queries for filtering by labels and text search. More complication in the code. And then another problem was discovered: text filtering uses whole words. It looks like this:

Aragon-_1_

There is also quite a lot of flickering. Some of the causes were addressed in the old FilterBar related tickets, which got discussed but never merged.

# Goals:

- [x] use standard `gql` syntax
- [x] remove `ProjectDetailQuery` component
- [x] improve page load time (load 25 issues at a time)
- [x] improve "Show More" behavior
  issue paging improvements 2020-01-24 at 15 49 28
- [x] make filtering update query, not filter client-side
- [x] show all labels & milestones for repo, not only those from current issue set

### DIFF
--- a/apps/projects/app/components/Content/ProjectDetail.js
+++ b/apps/projects/app/components/Content/ProjectDetail.js
@@ -5,7 +5,6 @@ import { useQuery } from '@apollo/react-hooks'
 
 import { useAragonApi } from '../../api-react'
 import { Button, GU, Header, IconPlus, Text } from '@aragon/ui'
-import { compareAsc, compareDesc } from 'date-fns'
 
 import useShapedIssue from '../../hooks/useShapedIssue'
 import { getIssuesGQL } from '../../utils/gql-queries.js'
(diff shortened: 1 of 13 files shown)


apps/projects/app/components/Shared/FilterBar/prepareFilters.js
```diff
@@ -26,41 +26,9 @@ const prepareFilters = (issues, bountyIssues) => {
...
-      }
-    } else {
-      if (!('milestoneless' in filters.milestones)) {
-        filters.milestones['milestoneless'] = {
-          title: 'Issues without milestones',
-          id: 'milestoneless',
-        }
-      }
-    }
+  filters.milestones = repo.metadata.milestones
```
Quazia: Why not just put these in the initial `filters` decleration?

apps/projects/app/components/Content/ProjectDetail.js
```diff
@@ -316,24 +337,65 @@ const ProjectDetailWrap = ({ repo, ...props }) => {
     requestPath('/issues/' + id)
   })
 
-  const graphqlQuery = useQuery(SEARCH_ISSUES, {
+  const SEARCH_ISSUES_2 = gql`
```
Quazia, line 321: Why is this named `SEARCH_ISSUES_2`

Quazia (dismissed):
LGTM

PR #1941, opened 2020-01-28
Decentralized project integration

rkzel:
NewProject panel has been connected to the contract via api.addProject function. New field, 'decoupled', is used to discern between GH and native projects.

DecoratedRepos context is where most interesting changes happened: new function has been added to pull metadata from IPFS - previously there was only one function there collecting information about GH projects directly from GH via Apollo client. Instead of adding IPFS to the same function, new function seemed to make sense, because they are only superficially similar - there were subtle differences, enough to warrant keeping the functionality separate. Also, in case there are more sources of repos/projects/issues added, it shall be done in the same manner.

What is working now:
- creating and displaying native and GH projects.
- removing both types.

What is not working:
- editing Projects,
- ProjectsDetail tab (that's waiting for: PR #1923  Projects: improve querying behavior).


apps/projects/app/components/Panel/NewProject/NewProject.js
```diff
@@ -113,110 +121,187 @@ const NewProject = () => {
...
+    <Query
+      fetchPolicy="cache-first"
+      query={GET_REPOSITORIES}
+      onError={console.error}
+    >
+      {({ data, loading, error, refetch }) => {
+        if (data && data.viewer) {
+          const reposDownloaded = filterAlreadyAdded(data.viewer.repositories.edges)
+          const visibleRepos = filter ? filterByName(reposDownloaded) : reposDownloaded
+          // eslint-disable-next-line react/prop-types
```
Quazia: Why are we disabling the prop-types here?

apps/projects/app/components/Panel/NewProject/NewProject.js
```diff
@@ -113,110 +121,187 @@ const NewProject = () => {
...
-                  <Button wide mode="strong" onClick={() => refetch()}>
-                      Try refetching?
-                  </Button>
-                </RepoInfo>
-              )
-            }}
-          </Query>
+  return (
+    <>
+      {github.status !== 'authenticated' ? (
```
Quazia: As it is now, this isn't accessible since the _initial_ splash screen requires auth. We need to go straight into project view.

apps/projects/app/components/Panel/NewProject/NewProject.js
```diff
@@ -3,16 +3,19 @@ import PropTypes from 'prop-types'
...
+import { toHex, sha3 } from 'web3-utils'
 import noResultsSvg from '../../../assets/noResults.svg'
 import { FormField, FieldTitle } from '../../Form'
 import { STATUS } from '../../../utils/github'
+import { useDecoratedRepos } from '../../../context/DecoratedRepos'
+import { ipfsAdd } from '../../../../../../shared/utils/ipfs'
 
-const disableDecoupledProjects = true
+
+const disableDecoupledProjects = false
```
ottodevs: This seems not needed anymore, as it cannot mutate in runtime, we can get rid of it unless we plan to do something with it in the future.
(Also remove condition in line 297)

apps/projects/app/components/Content/IssueDetail/index.js
```diff
@@ -46,37 +55,70 @@ Wrap.propTypes = {
...
   const shapeIssue = useShapedIssue()
-  const { loading, error, data } = useQuery(GET_ISSUE, {
-    client,
-    onError: console.error,
-    variables: { id: issueId },
-  })
 
+  const storedIssue = useMemo(() => {
+    return issues.find(i => issueId === i.data.issueId )
+  }, [issues])
```
ottodevs: We may need `issueId` also here, I suggest adding the react-hooks linting rules to better detect these cases, as we have in about app and in the client, so maybe not needed to change here, (as it does not seem to explicitly break the `storedIssue` value), but better when we add the rule in a separate commit.

apps/projects/app/components/Content/ProjectDetail.js
```diff
@@ -227,43 +225,86 @@ class ProjectDetail extends React.PureComponent {
...
-    const { data, loading, error, refetch } = this.props.graphqlQuery
-
-    if (loading) return this.queryLoading()
+    const { data, loading, error, refetch, fetchMore } = this.props.repo.decoupled ? {} : this.props.graphqlQuery
     if (error) return this.queryError(error, refetch)
 
-    const allIssues = [ ...cachedIssues, ...data.repository.issues.nodes ]
+    let dataSource
+    let pageInfo
+    if (data && !this.props.repo.decoupled) {
```
ottodevs: This condition does not seem necessary as line 228 already limits the `data` value to non-decoupled repos

apps/projects/app/components/Content/ProjectDetail.js
```diff
@@ -277,54 +318,102 @@ class ProjectDetail extends React.PureComponent {
...
+
   const shapeIssue = useShapedIssue()
   const { setupNewIssue } = usePanelManagement()
-  const [ client, setClient ] = useState(null)
-  const [ query, setQueryRaw ] = useState({ repoId, count: ISSUES_PER_CALL })
+  const [ query, setQueryRaw ] = useState({
+    repo: repo.decoupled ? repo.id : `${repo.metadata.owner}/${repo.metadata.name}`,
+    search: '',
+    sort: 'updated-desc',
+    owner: repo.decoupled ? 'AutarkLabs' : repo.metadata.owner,
```
ottodevs: Are we sure we need to hardcode `AutarkLabs` here?

apps/projects/app/components/Content/ProjectDetail.js
```diff
@@ -277,54 +318,102 @@ class ProjectDetail extends React.PureComponent {
...
+    name: repo.decoupled ? 'open-enterprise' : repo.metadata.name,
```
ottodevs: Are we sure we want to hardcode the `open-enteprise` value?

apps/projects/app/components/Panel/FundIssues/FundIssues.js
```diff
@@ -48,7 +48,10 @@ const errorMessages = {
 
 const bountiesFor = ({ bountySettings, issues, tokens }) => issues.reduce(
   (bounties, issue) => {
+    const data = issue.repository.decoupled ? issue : undefined
```
ottodevs: call here just to ensure that `undefined` is the appropriate type for non-decoupled repos data field

apps/projects/app/components/Panel/NewIssue/NewIssue.js
```diff
@@ -103,66 +112,125 @@ class NewIssue extends React.PureComponent {
...
+        title,
+        body: description,
+        author: {
+          login: this.props.account,
+          avatarUrl: '',
+          url: '',
+        },
+        repository: reposManaged[selectedProject - 1],
+        labels: { totalCount: 0, edges: Array(0) },
+        milestone: null,
```
ottodevs: It could make sense to get rid of the `milestone` field if we don't want to use it anymore, it was removed from the filterbar but still present across the code, and apparently always being assigned a "null" value, what is the purpose of having a value that will always be null? This may happen for other fields here.

apps/projects/app/components/Panel/NewIssue/NewIssue.js
```diff
@@ -103,66 +112,125 @@ class NewIssue extends React.PureComponent {
...
+        <Field label="Project">
+          <DropDown
+            items={items}
+            selected={selectedProject}
+            onChange={projectChange}
+            wide
+            required
+          />
+        </Field>
+        {selectedProject > 0 && (reposManaged[selectedProject - 1].decoupled ? (
```
ottodevs, line 159: No need to change right now, but just FYI we can use the dropdown placeholder to avoid these index modification hacks that are error-prone:
https://github.com/aragon/aragon-ui/tree/master/src/components/DropDown#placeholder

apps/projects/app/context/DecoratedRepos.js
```diff
@@ -1,13 +1,25 @@
...
+import { ipfsGet } from '../utils/ipfs-helpers'
 
 const repoQuery = repoId => `{
   node(id: "${repoId}") {
     ... on Repository {
       name
       url
       description
+      owner { login }
+      labels(first: 100) {
```
ottodevs, line 13: We should add pagination as we have in the issues query, as this is omitting lot of repositories.

apps/projects/app/utils/shapes.js
```diff
@@ -30,9 +30,30 @@ const issueShape = PropTypes.shape({
 }).isRequired
 
 const userGitHubShape = PropTypes.shape({
-  avatarUrl: PropTypes.string.isRequired,
+  avatarUrl: PropTypes.string,
```
ottodevs: It seems removing `isRequired` will hide potential warning messages, because at least `PanelComponents.js`->`Avatar` component is using this `prop` without any explicit check for its value, so it is actually a required prop unless we add any kind of check to that component (maybe there are other components involved as well)

ottodevs (dismissed):
Overall I didn't find code that is clearly wrong or causing malfunction of the app, just asking some minor questions, that are not necessarily involving further changes to this PR, but would be good to clarify in case they could hide some potential caveats.

Besides that, I detected some issues while testing the app locally,  the biggest problem I experienced was not being able to create decoupled issues nor create permissions to do so

I could not assign the needed permission to create decoupled issues, as it is not figuring in the permissions panel:

image

But still showing up as unassigned in the general permissions list:
image

ottodevs:
Fixes AUT-19

apps/projects/app/context/DecoratedRepos.js
topocount: I'm curious what this might look like, since the `useDecoratedRepos` hook that implements this query is used only to populate repos that are already added. We;re just not paginating the labels on these repos we know we already want to query. I've created [AUT-92](https://linear.app/issue/AUT-92/add-pagination-to-repo-query) to discuss this further

ottodevs (approved):
Just an overall recommendation, PLEASE AVOID FORCE-PUSHING PR changes once there is an active review, because it becomes impossible to track the changes this way...  Specially with such huge PR. 

Alternatively, there is a squash option when merging where we can squash every commit, so Force-pushing does not seem to offer any kind of advantage over that option.

Besides that personal advice/request, this seems ok now, great changes!

- Decoupled projects and their issues worked fine in my environment 
- Permissions look correct now, the "Create and update projects issues" one is finally showing up

As long as we consider the coverage decrease as acceptable for now, this looks ready to merge for me.

PR #1957, opened 2020-01-28
fix:  Very small dot-voting participation rounded

### LINKED ISSUE #1946: Fix percentage rounding for votes with 0% participation needed
Currently, it displays '1e-16% needed'; instead, it should display 0%.

Screenshot 2020-01-28 at 11 16 58

Quazia:
Fixes #1946 

Since a _very_ small participation is needed (can't be _true_ 0) We must round present the desired display value.

Screenshot from 2020-01-28 12-08-20


topocount (approved):
:+1:

PR #1975, opened 2020-02-07
fix: make start:dev script work again

ottodevs:
This is a temporary measure to unblock development from dev branch, it just removes the custom token-manager and whitelist from the dev-template, since this template was not updated yet to use this feature.

As this is temporary, some of the script files were copied as they are, the only actual modifications were adjusting the import path and replacing token-manager-custom by regular aragon token-manager and removing whitelist-oracle.


topocount (approved):
Thanks for jumping on this Otto! I don't see a ton of value in utilizing the new OE template for dev work since it's just so big, and the apps we added to it aren't (right now) under a ton of development. I think this fix allows the dev-template in it's current form to be a totally fine medum-term environment to work out of.

PR #1976, opened 2020-02-07
fix: reset remaining budget for new period and fix hard coded duration

PeterMPhillips:
This PR fixes two things:
1) It removes the hardcoded 'per 30 days' in BudgetDetails and sets it based on period.duration
2) If the current time is past the endDate time of the contract period. the budget remaining is set to the full budget amount. this way users can allocate funds after the period has ended

### DIFF
--- a/apps/allocations/app/app-state-reducer.js
+++ b/apps/allocations/app/app-state-reducer.js
@@ -19,7 +19,9 @@ const getTokenFromAddress = (tokenAddress, tokenList) => {
 }
 
 function appStateReducer(state) {
-  const { accounts: budgets, balances , allocations } = state || {}
+  const { accounts: budgets, balances , allocations, period } = state || {}
+  const { endDate } = period || {}
+  const currentDate = new Date()
 
   const balancesBn = balances
     ? balances
@@ -43,7 +45,10 @@ function appStateReducer(state) {
       ...budget,
       active: budget.hasBudget && Number(budget.amount) > 0,
       // get some extra info about the token
-      token: getTokenFromAddress(budget.token, balances)
+      token: getTokenFromAddress(budget.token, balances),
+      // if current period end date is less than now, current period is over
+      // but not updated on chain. so we reset the remaining budget
+      remaining: endDate < currentDate ? budget.amount : budget.remaining,
       // amount: new BigNumber(budget.amount),
       // numData: {
       //   amount: parseInt(budget.amount, 10),
--- a/apps/allocations/app/components/App/BudgetDetail.js
+++ b/apps/allocations/app/components/App/BudgetDetail.js
@@ -27,6 +27,8 @@ import BudgetContextMenu from '../BudgetContextMenu'
 import { formatDate } from '../../utils/helpers'
 import * as types from '../../utils/prop-types'
 
+const MS_PER_DAY = 24 * 60 * 60 * 1000
+
 const percentOf = (smaller, bigger) =>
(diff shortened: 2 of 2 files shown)


apps/allocations/app/components/App/BudgetDetail.js
```diff
@@ -38,6 +40,15 @@ const displayCurrency = (amount, decimals) => {
     .toLocaleString()
 }
 
+const processTime = ms => {
+  const days = Math.round(ms/MS_PER_DAY)
+  if(days%7 === 0){
+    const weeks = days/7
+    return `per ${weeks === 1 ? 'week' : `${String(weeks)} weeks`}`
+  }
+  return `per ${days === 1 ? 'day' : `${String(days)} days`}`
```
ottodevs, line 49: I am in favour of having this calculated already in the state or app-reducer instead of each time the component is rendered, so we can eventually use it elsewhere instead of forcing us to repeat these calculations each time we want to use them in the UI, so I think defining this locally is not the optimal solution.

apps/allocations/app/app-state-reducer.js
```diff
@@ -43,7 +45,10 @@ function appStateReducer(state) {
       ...budget,
       active: budget.hasBudget && Number(budget.amount) > 0,
       // get some extra info about the token
-      token: getTokenFromAddress(budget.token, balances)
+      token: getTokenFromAddress(budget.token, balances),
+      // if current period end date is less than now, current period is over
+      // but not updated on chain. so we reset the remaining budget
+      remaining: endDate < currentDate ? budget.amount : budget.remaining,
```
ottodevs, line 51: Just wondering if this will be intuitive for the user, I think the "remaining" field in the UI with abruptly change with the condition transition, but not sure if we should show also some visual indicator indicating the budget or period reset

apps/allocations/app/components/App/BudgetDetail.js
```diff
@@ -100,7 +111,7 @@ export default function BudgetDetail({ budget }) {
                     title="Budget"
                     large={displayCurrency(budget.amount, budget.token.decimals)}
                     small={budget.token.symbol}
-                    context="per 30 days"
+                    context={processTime(period.duration)}
```
ottodevs, line 114: Is this the only place that needs to be changed? I was not sure if the Account period details component also needs some change

ottodevs (commented):
Apparently the code is good, just trying to make sure if this is the only place that needs to be changed

apps/allocations/app/components/App/BudgetDetail.js
ottodevs: Anyway we can have it this way if we are sure this is not needed elsewhere for now 👍

apps/allocations/app/app-state-reducer.js
stellarmagnet: I think that for now we don't have to get fancy with a transition, can log an enhancement for the future though? It's rare that people will be staring at the page right when the transition occurs, and would prefer to avoid churn on getting this feature out

apps/allocations/app/components/App/BudgetDetail.js
stellarmagnet: Right now the ACCOUNT PERIOD DETAILS is correctly displaying the correct period duration.

So yes, this is the only place we need to update this right now.

ottodevs (approved):
Ok, let's merge this, thanks for your review @stellarmagnet

PR #1977, opened 2020-02-10
refactor(test-helpers): remove test-helpers part one: contracts folder

ottodevs:
Even if this commit looks huge, it is actually quite simple:
- It removes `contracts` folder from `@tps/test-helpers` and its content
- Moves ADynamicForwarder and DynamicScriptHelpers locally to DotVoting
- Change DotVoting imports
- Inline IForwarder and rename to solve issues with coverage
- Removing pre and post coverage hooks, unnecessary for validation after
inlining the interface

Note: ADynamicForwarder and DynamicScriptHelpers are now part of DotVoting app, that is why coverage was decreased. I decided to not ignore those files, as I think they should be tested as well.


apps/dot-voting/contracts/lib/ADynamicForwarder.sol
```diff
@@ -8,9 +8,21 @@ import "@aragon/os/contracts/lib/math/SafeMath.sol";
 import "@aragon/os/contracts/lib/math/SafeMath64.sol";
 
 // TODO: Use @aragon/os/contracts/ version when it gets merged
-import "../evmscript/DynamicScriptHelpers.sol";
+import "./DynamicScriptHelpers.sol";
 // TODO: Research why using the @aragon/os version breaks coverage
-import "@aragon/os/contracts/common/IForwarder.sol";
```
topocount, line 13: Let's not re-implement this until we decide how to proceed with renaming this import during coverage

apps/dot-voting/package.json
```diff
@@ -13,9 +13,7 @@
     "frontend": "npm run sync-assets && parcel app/index.html --port 4444",
     "ganache-cli:test": "sh ../../node_modules/@aragon/test-helpers/ganache-cli.sh",
     "lint": "solium --dir ./contracts",
-    "postcoverage": " sed -i 's+./IForwarder.sol+@aragon/os/contracts/common/IForwarder.sol+g' ../shared/test-helpers/contracts/common/ADynamicForwarder.sol",
     "precommit": "lint-staged",
-    "precoverage": "sed -i 's+@aragon/os/contracts/common/IForwarder.sol+./IForwarder.sol+g' ../shared/test-helpers/contracts/common/ADynamicForwarder.sol",
```
topocount, line 18: It doesn't make sense to remove this since it addresses a bug in the coverage tool. We should upgrade our coverage tool before removing this since it doesn't make sense to create our own duplicated import when it already exists in a dependency we're using

topocount (changes requested):
We should discuss making the IForwarder change without addressing the bug in the coverage tool. Curious what @Quazia thinks

apps/dot-voting/contracts/lib/ADynamicForwarder.sol
ottodevs: AFAIK, there is no coverage bug, I properly tested the change and coverage is working.

In my opinion this is the best tradeoff we can opt for now, as there was no other approach suggested for months.

We should not invest more time thinking on this or keep maintaining the ridiculous `tps/test-helpers` module just for this reason, I think we should instead focus into upgrading `@aragon/os` upstream to import the ADynamicForwarder from there, so as temporary solution, it seems fine for me.

What is your specific concern about this solution @topocount did you have the chance to test this new code to check that coverage is working correctly?

Now is the moment to decide, this task was around for months and we are always postposing it, I spent a reasonable amount of time testing that coverage works correctly, so, please, let's decide on that, please, be more specific and share your concerns about the change so we can address them.

apps/dot-voting/package.json
ottodevs: If you take the time to test the code you will notice I fixed the coverage bug so there is no need to use these hooks anymore.

I commented about it in the commit message:
`- Removing pre and post coverage hooks, unnecessary for validation after
inlining the interface`

PR #1980, opened 2020-02-10
fix(allocations): distributions with zero-support candidates

topocount:
when all payout funds are transferred for a given allocation,
the allocation should be marked as "enacted". However, candidates
with zero support cannot execute and therefor cannot meet the execution
condition, so a special case should be made for them to ensure
the payout is marked as "enacted" instead of just "approved"


PeterMPhillips (approved):
Looks good!

PR #1981, opened 2020-02-10
fix: add period advancement to set distribution function

PeterMPhillips:
This PR makes a contract change by adding the `transitionsPeriod` modifier to `setDistribution` which will allow the period to advance when funds are allocated via Dot Voting.


Quazia (approved):
LGTM

PR #1986, opened 2020-02-11
fix(allocations): filter out state missing token data

topocount:
budgets and allocations instances should not be passed into the frontend
if they are missing token data. This always occurs on syncing because
app contract state syncs completely before any external state (such as
    token data) is passed into the app via events. Filtering out state that is
missing token details will prevent any rendering or data access issues.


PeterMPhillips (approved):
Looks good. I didn't encounter any issues when running locally. Hopefully the issue will be fixed in Rinkeby testing!

PR #1988, opened 2020-02-12
fix(allocations): frontend period and budget remaining override

topocount:
Sometimes the app state reducer doesn't kick in to globally override the
remaining budget. These calculations in the frontend components ensure
the remaining budget is set to zero if the onchain period end date has
passed. (because an un-transitioned accounting period indicates that no
    funds have been spent since the last cycle)


topocount:
Closing this in favor of #1989 since it addresses the root issue

PR #1989, opened 2020-02-12
fix(allocations): update budget state on period transition

topocount:
When a period transitioned, The budget state was not updating
in the frontend and actually reverts to reflecting the last active
period's state. This update pulls in the updated budget contract state
after a period's state changes onchain. frontend logic  for reflecting
budget state before a transition is still sound and not encompassed
in these changes


Quazia (approved):
LGTM

PR #1994, opened 2020-02-13
chore(dev-template): fix hoist conflicts

topocount:
pin dependencies so hoist conflicts are mitigated. This should speed up package installation


ottodevs (approved):
OK

PR #1996, opened 2020-03-06
fix: removed fulfilled bounties from bounties screen

PeterMPhillips:
This PR removes fulfilled bounties from the Bounties screen


e18r (approved):
LGTM

PR #2002, opened 2020-05-20
Decoupled Projects fixes

topocount:
fixes DEV-190
fixes DEV-189
fixes DEV-165
fixes DEV-161

### DIFF
--- a/apps/projects/app/store/helpers/issues.js
+++ b/apps/projects/app/store/helpers/issues.js
@@ -83,14 +83,22 @@ export const loadIpfsData = async ({ ipfsHash }) => {
 
 export const loadDecoupledIssueData = async ({ repoId, issueNumber }) => {
   const {
+    standardBountyId,
+    balance,
+    assignee,
+    fulfilled,
     data: ipfsHash
   } = await app.call('getIssue', repoId, issueNumber).toPromise()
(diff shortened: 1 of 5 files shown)


apps/projects/app/store/helpers/issues.js
```diff
@@ -79,13 +81,18 @@ export const loadIpfsData = async ({ ipfsHash }) => {
...
+export const loadDecoupledIssueData = async ({ repoId, issueNumber }) => {
+  const {
+    data: ipfsHash
+  } = await app.call('getIssue', repoId, issueNumber).toPromise()
+  let ipfsData = await ipfsGet(ipfsHash)
+  return {
+    ...ipfsData,
+    repoId: hexToAscii(repoId),
+    repoHexId: repoId,
+    token: ipfsData.token.addr
```
PeterMPhillips: `ipfsData.token` is undefined when a decoupled issue is first created, so this causes the store to fail when creating an issue, I wrote this to get past the error: `token: ipfsData.token ? ipfsData.token.addr : null` but maybe you have a better way.

PeterMPhillips (changes requested):
Right now when I submit an application and it gets accepted, the work flow doesn't progress any further. I'm unable to submit work.

PeterMPhillips (approved):

